PR #1, opened 2022-03-16
ROE-16 Add makefile with sonar configuration

mattch1:
Added a make file to the application it has in it:
A build command that runs ci
and a clean command that removes the build and dist directory and the zip file as well as the log
and a lint and sonarqube command
and it also runs coverage
It then packages and builds the app


mwestacott:
I don't think the .idea directory should be added to source control.

Makefile
```diff
@@ -0,0 +1,49 @@
+artifact_name       := overseas-entities
```
markpit, line 1: i think artifact name should be overseas-entities-web

mouhajer-ch, line 1: Probably should be `overseas-entities-web`?

mattch1: See: https://github.com/companieshouse/confirmation-statement-web/pull/6/files

mattch1: See: https://github.com/companieshouse/confirmation-statement-web/pull/6/files

markpit: ah, looks like it was corrected since then - https://github.com/companieshouse/confirmation-statement-web/blob/main/Makefile

mwestacott:
One more IDEA file to delete @mattch1 .

mwestacott (approved):

PR #3, opened 2022-03-16
Feature/roe 14 create basic web app structure

mouhajer-ch:
Create basic web app structure

- Add configuration file for tsconfig, eslint and jest and modify gitignore
- Add essential application dependencies
- Add views directory with layout and a placeholder for the landing page
- Add source config file and getEnvironment
- Add http server container logic
- Add application content logic
- Add landing page routing and logic
- Add test setup
- Add unit test for config and controller logic

Related to [ROE-14](https://companieshouse.atlassian.net/browse/ROE-14)


src/bin/www.ts
```diff
@@ -0,0 +1,54 @@
...
+function onError(error: any) {
+  if (error.syscall !== "listen") {
+    throw error;
+  }
+
+  // handle specific listen errors with friendly messages
+  switch (error.code) {
+      case "EACCES":
+        // TODO implement logger
+        // logger.error(config.PORT + " requires elevated privileges");
```
mwestacott, line 43: Should we add the logger integration as a new task to https://companieshouse.atlassian.net/browse/ROE-12 maybe?

src/config/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  if (!value) {
+    throw new Error(`Please set the environment variable "${key}"`);
+  }
+
+  return value;
+};
+
+// APP CONFIGs
+export const APPLICATION_NAME = "overseas-entities-web";
```
mwestacott, line 12: I can't see where this global constant is used. Is it required (yet)?

test/config/index.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { getEnvironmentValue, CDN_HOST } from "../../src/config";
+
+describe('Config test suite', () => {
+  it('should check if CDN_HOST env is returned correctly and fetched from user environment', () => {
+    const test_cdn_host = getEnvironmentValue("CDN_HOST");
+    expect(test_cdn_host).toEqual(CDN_HOST);
+  });
+
+  it('should thrown an error when passing anyNonExistingEnv to getEnvironmentValue()', () => {
```
mwestacott, line 9: V Minor - 'should throw an error...'

src/controllers/index.controller.ts
```diff
@@ -0,0 +1,7 @@
+import { Request, Response } from "express";
+
+import * as config from "../config";
+
+export const startController = (req: Request, res: Response) => {
```
markpit, line 5: we have typically used 'get' and 'post' for the function names in the controllers - eg - https://github.com/companieshouse/confirmation-statement-web/blob/main/src/controllers/start.controller.ts
how do you feel about using that approach?  I think it makes it easier to read as we can easily see what function is called for the get and what is called for the post.

views/layout.html
```diff
@@ -0,0 +1,74 @@
...
+  <script src="https://code.jquery.com/jquery-3.4.1.js"
+          integrity="sha256-WpOohJOqMqqyKL9FccASB9O0KwACQJpFTUBLTYOVvVU="
+          crossorigin="anonymous">
+  </script>
+
+{% endblock %}
+
+{% block header %}
+  {{ govukHeader({
+    homepageUrl: "/overseas-entities-web",
```
markpit, line 35: noticed on confirmation-statement-web we have homepage set to www.gov.uk. I think it was raised as a bug previously, here's the PR that changed it - https://github.com/companieshouse/confirmation-statement-web/commit/68b6f76e47a7b42dea4038f0440f75ddf25fb430

src/controllers/index.controller.ts
markpit: here's the original bug - https://companieshouse.atlassian.net/browse/NCS-290

src/app.ts
```diff
@@ -0,0 +1,33 @@
...
+nunjucksEnv.addGlobal("CDN_HOST", config.CDN_HOST);
+
+app.use(express.json());
+app.use(express.urlencoded({ extended: false }));
+
+app.set("views", path.join(__dirname, "views"));
+app.set("view engine", "html");
+
+// apply our default router to /
+app.use("/", router);
```
markpit, line 31: should we add our application root path here (line 31)
something like 
app.use("/register-an-overseas-entity", router);
and have the "/register-an-overseas-entity" as a constant somewhere (maybe in a constants.ts? or urls.ts?)

package.json
```diff
@@ -0,0 +1,41 @@
...
+  "dependencies": {
+    "express": "^4.17.3",
+    "govuk-frontend": "^4.0.1",
+    "nunjucks": "^3.2.3"
+  },
+  "devDependencies": {
+    "@types/express": "^4.17.13",
+    "@types/jest": "^27.4.1",
+    "@types/node": "^17.0.21",
+    "@types/nunjucks": "^3.2.1",
```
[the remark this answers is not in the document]
mouhajer-ch: Hi, we do not need them in in the  `dependencies`  section. `devDependencies` are required to run only in development mode, used for example for unit tests or helping with type definitions by using (`@types/*`).
After compiling it, we need the dependencies section and the compiled typescript code, in the `dist` folder, without types, typescript and others

src/bin/www.ts
mouhajer-ch: Morning, yes indeed

src/config/index.ts
mouhajer-ch: It is not used yet, but it will be soon essential when we integrate the` @companieshouse/structured-logging-node`

test/config/index.spec.ts
mouhajer-ch: Will do, thanks

src/controllers/index.controller.ts
mouhajer-ch: It sounds good to me, will do

src/app.ts
mouhajer-ch: Thanks Mark, agreed

views/layout.html
mouhajer-ch: Thanks Mark, well spotted. Done

src/app.ts
mouhajer-ch: Done a bit different, router file will have the responsibility and the knowledge of our endpoints

src/config/index.ts
mwestacott: Ok, not an issue, but suggest we only add code at the point it's needed and (ideally) each line of code should somehow track back to a requirement in the story or the task the PR is related to. Sorry, I get a bit obsessed about this!

mwestacott: Just seen you've removed this. Ignore my comment and thanks!

src/config/index.ts
```diff
@@ -14,7 +14,8 @@ export const NODE_ENV = getEnvironmentValue("NODE_ENV");
 export const CDN_HOST = getEnvironmentValue("CDN_HOST");
 
 // ROUTING PATH
-export const LANDING_URL = "/";
+export const REGISTER_AN_OVERSEAS_ENTITY_URL = "/register-an-overseas-entity";
+export const LANDING_URL = REGISTER_AN_OVERSEAS_ENTITY_URL;
```
mwestacott, line 18: Just wondering, why have LANDING_URL and why not just use/reference the new REGISTER_AN_OVERSEAS_ENTITY_URL?

mouhajer-ch: Hi, just naming convention, I could have used the REGISTER_AN_OVERSEAS_ENTITY_URL but it is going to be the first section of all our endpoints, so I included it in the landing page as well. It will have been overkilling having an endpoints like this with just  "/" at the end 
`export const LANDING_URL = REGISTER_AN_OVERSEAS_ENTITY_URL + "/";`

mwestacott: So, yes, I like the REGISTER_AN_OVERSEAS_ENTITY_URL name but still not sure why we bother keeping a LANDING_URL constant which is exactly the same as REGISTER_AN_OVERSEAS_ENTITY_URL. Maybe I'm missing something? happy to have a quick call...

mouhajer-ch: Lets have a chat :)

mwestacott (approved):
Looks good. Did you add that JIRA task to do the logger integration though?

mouhajer-ch:
> Looks good. Did you add that JIRA task to do the logger integration though?

I am talking with Mark, we will add the subtask related to `@companieshouse/structured-logging-node` now, thanks

PR #17, opened 2022-03-21
ROE-28 Create Service Avialability middleware

mattch1:
Added new middleware directory. For now left out the whitelisted urls as this can be added later when required.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -4,6 +4,7 @@ import * as path from "path";
 import * as config from "./config";
 import { logger } from "./utils/logger";
 import router from "./routes";
+import { serviceAvailabilityMiddleware } from "./middleware/service.availability.middleware";
 
 const app = express();
 
@@ -27,6 +28,8 @@ app.use(express.urlencoded({ extended: false }));
 
 app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
+
+app.use(serviceAvailabilityMiddleware);
 
 // apply our default router to /
 app.use("/", router);
--- a/src/types/template.paths.ts
+++ b/src/types/template.paths.ts
@@ -1,0 +1,3 @@
+export enum Templates {
+  SERVICE_OFFLINE = "service-offline"
+}
--- a/src/utils/feature.flag.ts
+++ b/src/utils/feature.flag.ts
@@ -1,0 +1,15 @@
+/**
+ * Feature flags will be determined by environment variables and all environment variables in nodejs are
+ * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
+ */
+export const isActiveFeature = (flag: string | undefined): boolean => {
+
+  if (flag === undefined) {
+    return false;
+  }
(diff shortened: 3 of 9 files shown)


src/app.ts
```diff
@@ -4,6 +4,7 @@ import * as path from "path";
 import * as config from "./config";
 import { logger } from "./utils/logger";
 import router from "./routes";
+import { serviceAvailabilityMiddleware } from "./middleware/service.availability.middleware";
```
mouhajer-ch, line 7: These changes should be added in the routes file her https://github.com/companieshouse/overseas-entities-web/blob/main/src/routes/index.ts

src/utils/feature.flag.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Feature flags will be determined by environment variables and all environment variables in nodejs are
+ * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
+ */
+export const isActiveFeature = (flag: string | undefined): boolean => {
+
+  if (flag === undefined) {
+    return false;
+  }
+  const featureFlag = flag.toLowerCase();
```
mouhajer-ch, line 10: We have already set the feature flag to be `false`? I do not think we need all of these condition

src/utils/properties.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Gets an environment variable. If the env var is not set and a default value is not
+ * provided, then it is assumed it is a mandatory requirement and an error will be
+ * thrown.
+ */
+
+const getEnvironmentVariable = (key: string, defaultValue?: any): string => {
```
mouhajer-ch, line 7: These logic (improved) has already been implemented here https://github.com/companieshouse/overseas-entities-web/blob/main/src/config/index.ts

src/types/template.paths.ts
```diff
@@ -0,0 +1,3 @@
+export enum Templates {
```
mouhajer-ch, line 1: The `service-offline` template should be added on config file here https://github.com/companieshouse/overseas-entities-web/blob/eb7535674ad3a1457b3c0275021bdb4bfd3ccd46/src/config/index.ts#L22

src/utils/feature.flag.ts
mattch1: The relates to all feature flags not just this one.

src/app.ts
mattch1: Removed from the app and placed this in routes.

src/utils/properties.ts
mattch1: Removed properties file and put this into config.

src/types/template.paths.ts
mattch1: Moved templates to config and removed the types folder from this PR.

mouhajer-ch (commented):
Hi, you have not yet pushed the latest changes :)

mouhajer-ch (approved):

PR #23, opened 2022-03-22
ROE-32 adding presenter page template and unit test

SVarallo:
PR to add the presenter page html and integrate that into the existing presenter controller and unit test class. Also added a back link includes as this is an html element that will be used throughout the service.


mouhajer-ch (commented):
Can we add a description for the PR, please

PR #31, opened 2022-03-24
Feature/roe 35 create session utils

mouhajer-ch:
### JIRA link

Part of [ROE-35](https://companieshouse.atlassian.net/browse/ROE-35)

### Change description

Create session utils

### Work checklist

- [X ] Tests added where applicable

### DIFF
--- a/src/model/data.types.model.ts
+++ b/src/model/data.types.model.ts
@@ -1,0 +1,12 @@
+export interface Address {
+    addressLine1?: string
+    addressLine2?: string
+    town?: string
+    county?: string
+    postcode?: string
+}
+
+export enum userResponse {
+    No = 0,
+    Yes = 1
+  }
--- a/src/model/data.types.model.ts
+++ b/src/model/data.types.model.ts
@@ -6,7 +6,7 @@ export interface Address {
     postcode?: string
 }
 
-export enum userResponse {
+export enum yesNoResponse {
     No = 0,
     Yes = 1
   }
--- a/src/model/entity.model.ts
+++ b/src/model/entity.model.ts
@@ -1,0 +1,21 @@
+import { Address, userResponse } from "./data.types.model";
+
+export const EntityKey = "entity";
+/*
+  The Entity fields will have to match the name field on the HTML file to
+  be able to do the mapping correctly
+*/
+export const EntityKeys: string[] = ["overseasEntityName", "incorporationCountry", "principalAddress", "isAddressSameToPrincipalAddress",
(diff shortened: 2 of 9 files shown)


src/model/data.types.model.ts
```diff
@@ -0,0 +1,12 @@
+export interface Address {
+    addressLine1?: string
+    addressLine2?: string
+    town?: string
+    county?: string
+    postcode?: string
+}
+
+export enum userResponse {
```
markpit, line 9: seeing as this can only be YES or No, would calling it yesNoResponse be better? Might make it easier to understand when looking at the interfaces that use it

src/model/entity.model.ts
```diff
@@ -0,0 +1,21 @@
...
+  The Entity fields will have to match the name field on the HTML file to
+  be able to do the mapping correctly
+*/
+export const EntityKeys: string[] = ["overseasEntityName", "incorporationCountry", "principalAddress", "isAddressSameToPrincipalAddress",
+  "serviceAddress", "email", "legalForm", "governedLaw", "publicRegister", "registrationNumber" ];
+export interface Entity {
+    overseasEntityName?: string
+    incorporationCountry?: string
+    principalAddress?: Address
+    isAddressSameToPrincipalAddress?: userResponse
```
markpit, line 14: I'd rename isAddressSameToPrincipalAddress to isAddressSameAsPrincipalAddress (just change the To to As)

test/utils/application.data.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    const data = getApplicationData(session);
+    expect(data).toEqual(APPLICATION_DATA_MOCK);
+  });
+  test("setApplicationData should store application data into the session", () => {
+    const session = getSessionRequestWithExtraData();
+    setApplicationData(session, ENTITY_OBJECT_MOCK, EntityKey);
+
+    const data = getApplicationData(session);
+    expect(data).toEqual( { ...APPLICATION_DATA_MOCK, [EntityKey]: { ...ENTITY_OBJECT_MOCK } });
+  });
```
markpit, line 18: should we have a test for the prepareData function? i think we could pass it an array of keys and data and check that it returns the correct object.

src/model/entity.model.ts
mouhajer-ch: Done

src/model/data.types.model.ts
mouhajer-ch: Done

test/utils/application.data.spec.ts
mouhajer-ch: Done

markpit (approved):

PR #49, opened 2022-03-30
Roe 114 create corporate managing officer page

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-114

### Change description
Created managing officer corporate page with routing, controller and unit testing.

### Work checklist

### Merge instructions

```
```


src/controllers/managing.officer.corporate.controller.ts
```diff
@@ -0,0 +1,8 @@
+import { Request, Response } from "express";
+import * as config from "../config";
+import { logger } from "../utils/logger";
+
+export const get = (req: Request, res: Response) => {
+  logger.debug(`GET ${config.MANAGING_OFFICER_CORPORATE_PAGE}`);
+  return res.render(config.MANAGING_OFFICER_CORPORATE_PAGE);
```
SVarallo, line 7: Do you know the page that comes before this right now? Will need to pass in the backLinkUrl if so

views/managing-officer-corporate.html
```diff
@@ -0,0 +1,303 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Corporate managing officer – {{ serviceName }} – GOV.UK
```
markpit, line 4: need to replace service name with the name of our app (see other templates) and the " – " character is flagged as invalid in the vscode editor  - The character U+2013 "–" could be confused with the character U+002d "-", I would change them to be minus signs.

views/managing-officer-corporate.html
```diff
@@ -0,0 +1,303 @@
...
+
+      {{ govukButton({
+        text: "Add another corporate managing officer",
+        href: "",
+        classes: "govuk-button--secondary govuk-!-margin-top-6"
+      }) }}
+
+      {{ govukButton({
+        text: "I do not need to add any more corporate managing officers",
+        href: "/register/check-answers"
```
markpit, line 295: button needs a submit attribute - see https://github.com/companieshouse/overseas-entities-web/blob/main/views/beneficial-owner-gov.html#L311

test/controllers/managing.officer.corporate.controller.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+import { NextFunction, Request, Response } from "express";
+import request from "supertest";
+import app from "../../src/app";
+import { authentication } from "../../src/controllers";
+import { MANAGING_OFFICER_CORPORATE_URL } from "../../src/config";
+
+const mockAuthenticationMiddleware = authentication as jest.Mock;
+mockAuthenticationMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => next() );
+
+const MANAGING_OFFICER_CORPORATE_PAGE_TITLE = "Tell us about the corporate managing officer";
```
markpit, line 14: page heading should go in https://github.com/companieshouse/overseas-entities-web/blob/main/test/__mocks__/text.mock.ts with the others

views/managing-officer-corporate.html
SVarallo: Wont the other button need an id as well so the automated tests know the difference between them?

markpit: yes, although i didn't add one on my template (for the first button) not sure what the the id would be as submit would be used for the 'submit' page (second button)

views/managing-officer-corporate.html
markpit: does the {{ serviceName }} get replaced? i had replaced it with text on mine but perhaps I didn't need to, does it get replaced in your docker env?

views/managing-officer-corporate.html
markpit: i think Sam is right, we should add an id attribute for the first button, I'll do the same on my page. I'm going to add an id of 'add' on mine, if you do the same here then they'll be consistent.

markpit (approved):

PR #58, opened 2022-04-04
ROE-121 Make types of beneficial owner conditional and reusable

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-121

### Change description

Replace existing checkboxes to match prototype version three and pass session value into the conditional. Also make the checkboxes reusable.

### Work checklist

### Merge instructions

```
```


views/beneficial-owner-type.html
```diff
@@ -10,55 +10,17 @@
...
-          {{ govukDetails({
-            summaryText: "What is a beneficial owner?",
-            text: "A beneficial owner is an individual who ultimately owns or controls more than 25% of a company's shares or voting rights, or who otherwise exercise control over the company or its management."
-          }) }}
-          <p class="govuk-hint">Select all that apply.</p>
-        {% endset %}
+        {% if (statement.beneficialOwnerStatement === "allIdentifiedAllSupplied") %}
+          {% include "includes/beneficial-owners-check.html" %}
+        {% else %}
+          {% include "includes/beneficial-owners-check.html" %}
```
markpit, line 21: the include for the 'else' is the same as the 'if' - could just remove the 'if else' and just include

mattch1: A non issue as this is only temporary. I want the if else structure there as this is part of the development to match the AC but also want to ensure that we still get something on the page. This will be addressed on the next ticket.

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -11,10 +11,12 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     logger.debug(`GET ${config.BENEFICIAL_OWNER_TYPE_PAGE}`);
 
     const appData: ApplicationData = getApplicationData(req.session);
+    const statement = appData.beneficialOwnerStatement;
 
     return res.render(config.BENEFICIAL_OWNER_TYPE_PAGE, {
       backLinkUrl: config.BENEFICIAL_OWNER_STATEMENTS_URL,
-      ...appData.beneficialOwnerType
+      ...appData.beneficialOwnerType,
+      statement
```
markpit, line 19: I think it would be better to check the statement types in the controller, to calculate the 2 flags, one to show the 'top half' (beneficial owners?) and one for the 'bottom half' (managers?). That would keep the logic in the template to a minimum, I think the templates should be as 'dumb' as possible. Passing a flag to display or not display a section should keep the template simple.

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -4,26 +4,36 @@ import { ApplicationData, ApplicationDataType, beneficialOwnerTypeType } from ".
...
-      statement
+      isBeneficialOwners
     });
   } catch (error) {
     logger.errorRequest(req, error);
     next(error);
   }
 };
 
+const areBeneficialOwnersIdneitified = (statement: BeneficialOwnerStatement | undefined): boolean => {
```
markpit, line 28: typo in function name

markpit (approved):

PR #78, opened 2022-04-18
Feature/roe save address object on managingofficers and beneficialowner to the same datatype

mouhajer-ch:
### JIRA link

Related to [ROE-158](https://companieshouse.atlassian.net/browse/ROE-158)

### Change description

- Save address object for MOs and BOs to the same data type Address, improve readability and fix a few address issues
- Needed for check your answers page to reuse the [views/includes/display_address.html](https://github.com/companieshouse/overseas-entities-web/blob/main/views/includes/display_address.html)

### Work checklist


src/controllers/check.your.answers.controller.ts
```diff
@@ -32,8 +32,13 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
...
 
     const transaction: Transaction = await postTransaction(req.session as Session);
+    logger.info(`Transaction created, ID: ${transaction.id}`);
+
     const overseaEntity: OverseasEntityCreated = await createOverseasEntity(req.session as Session, transaction.id as string);
-    await closeTransaction(req.session as Session, transaction.id as string, overseaEntity.id);
+    logger.info(`Overseas Entity Created, ID: ${overseaEntity.id}`);
+
+    const responseClosedTransaction = await closeTransaction(req.session as Session, transaction.id as string, overseaEntity.id);
+    logger.info(`Transaction Closed,  ${JSON.stringify(responseClosedTransaction)}`);
```
markpit, line 41: mentioned in PR 77 (which this is based on), this should be changed to log just the ID instead of the whole transaction object

markpit (approved):

src/controllers/check.your.answers.controller.ts
mouhajer-ch: Thanks Mark for looking into it and as I said, I agree but first we need to change the response coming from the SDK, it will be done in a future story

PR #110, opened 2022-05-03
ROE-168 Display v5 fields and headings on entity page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-168

### Change description

Changed the template to match the fields on the v5 template.

Not changed the model in the wen have created subtask: [roe-413](https://companieshouse.atlassian.net/browse/ROE-413)

for this.

### Work checklist

### Merge instructions

```
```


views/entity.html
```diff
@@ -10,254 +10,233 @@
...
+        },
+        hint: {
+          text: "We’ll use this to send the Overseas Entity ID number, the notice of registration, and any other digital correspondence"
+        },
+        id: "email",
+        name: "email",
+        value: email
+      }) }}
+
+      {% set hintText = "For example, limited company" %}
```
SVarallo, line 176: Screenshot shows this hint text as "For example, limited company or partnership"

views/entity.html
```diff
@@ -10,254 +10,233 @@
...
+          text: "Which country was it formed in?",
+          classes: "govuk-label--m",
+          isPageHeading: false
+        },
+        id: "incorporation_country",
+        name: "incorporation_country",
+        value: incorporation_country
+      }) }}
+
+      {% include "includes/inputs/principal-address-input.html" %}
```
SVarallo, line 46: As Mark pointed out there's two address radio options. The includes used above contains where you input the principal address, the radio buttons and the subsequent area to input the service address so you wont need the that bit just below here down to line 160 I believe.

SVarallo: Have noticed that the text on the radio buttons in the includes element I made is incorrect and should be "Is the correspondence address the same as the principal or registered office address" so that needs to be changed, feel free to do it in this or I'll do it in my story

views/entity.html
mattch1: Interesting: Here https://roe-v5.herokuapp.com/register/entity we have "For example, limited company" . I was looking at this it seems to be different from the screenshot.

views/entity.html
mattch1: I added the includes at the last minute and forgot to take the address options on the main page away.

views/entity.html
markpit: I agree, probably best to take the text from the prototype as it is probably more up to date than the screenshot.

markpit:
**What** - Sonar has failed the pull request as it has picked up duplicate blocks of code
**Why** - Duplicate code is considered bad practice, it makes code harder to maintain. For more in depth explanations there are many pages online, here is the wikipedia entry - https://en.wikipedia.org/wiki/Duplicate_code
**How** - One of the blocks picked up by Sonar is the name input field on lines 21-30 of beneficial-owner-gov.html which is duplicated on line 22-31 of beneficial-owner-other.html. This can be made into an 'include' file and included in both templates. I would suggest creating a name.html file under the views/includes/inputs/fields. You will need to merge the latest code from main to have an up to date folder structure. This should reduce the duplication percentage, maybe enough for Sonar to approve.

mattch1:
Hilarious

markpit:
> Hilarious

I really don't understand anymore. You refuse to speak to your team mates when we offer you help. You ignore our Slack messages. You then insist that we add comments to your PR's in a What Why How format which is exactly what I have done. Martin asked you for an example of this comment structure which you ignored. You are not covering yourself in glory here and need to start acting as a team player.

mattch1:
Alright perhaps I went too far here. So apologies for that. My original intentions were good in trying to simplify this  by focusing just on the template changes as we encountered massive PR for the web with https://companieshouse.atlassian.net/browse/ROE-275. It turned out not be as simple as was hoping but that happens and I should have adjusted  to this accodingly. Will revise my approach to this going forward.

markpit:
I apologise too, I overreacted. I know your intentions are well meant, I think we are all getting frustrated by the project and deadlines.

markpit (approved):

mouhajer-ch (approved):
I will approve it, thanks @mattch1 @markpit, @SVarallo

PR #111, opened 2022-05-03
Feature/add new msg error on authentication middleware and update directory path

mouhajer-ch:
### Change description

- Update path for authentication middleware and add new log

### DIFF
--- a/src/middleware/authentication.middleware.ts
+++ b/src/middleware/authentication.middleware.ts
@@ -11,8 +11,7 @@ import {
 export const authentication = (req: Request, res: Response, next: NextFunction): void => {
   try {
     if (!checkUserSignedIn(req.session)) {
-      logger.infoRequest(req, 'user not authenticated, status_code=401');
-      logger.infoRequest(req, 'redirecting to sign in page, status_code=302');
+      logger.infoRequest(req, 'User not authenticated, redirecting to sign in page, status_code=302');
       return res.redirect(`/signin?return_to=${PRESENTER_URL}`);
     }
 
--- a/src/middleware/authentication.middleware.ts
+++ b/src/middleware/authentication.middleware.ts
@@ -11,7 +11,8 @@ import {
 export const authentication = (req: Request, res: Response, next: NextFunction): void => {
   try {
     if (!checkUserSignedIn(req.session)) {
-      logger.infoRequest(req, 'User not authenticated, status_code=401, redirecting to sign in page');
+      logger.info("User not authenticated, status_code=401");
+      logger.infoRequest(req, 'redirecting to sign in page, status_code=302');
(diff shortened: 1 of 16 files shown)


src/middleware/authentication.middleware.ts
```diff
@@ -11,7 +11,8 @@ import {
 export const authentication = (req: Request, res: Response, next: NextFunction): void => {
   try {
     if (!checkUserSignedIn(req.session)) {
-      logger.infoRequest(req, 'User not authenticated, status_code=401, redirecting to sign in page');
+      logger.info("User not authenticated, status_code=401");
```
markpit, line 14: should the logger.info be logger.infoRequest on line 14? it should give us path info and I think request id

mouhajer-ch: There is already a `logger.infoRequest` on line 15 I did not want to log the same `req` twice

mouhajer-ch: Is just logging the `request.path` and `request.method`, I will push the fix shortly

src/middleware/authentication.middleware.ts
```diff
@@ -11,7 +11,8 @@ import {
...
+      logger.infoRequest(req, 'redirecting to sign in page, status_code=302');
```
mwestacott, line 15: I'm a bit confused - line 14 log message says status code is 401 and then this line says it's 302. Nothing's happened in between the two log messages so which one is right? Or, is the second one saying that the response/status code will be 302 because of the redirect? Can log messages be corrected/reworded to make clear what's going on here please?

test/middleware/authentication.middleware.spec.ts
```diff
@@ -55,7 +54,7 @@ describe('Authentication controller', () => {
     expect(res.redirect).toHaveBeenCalledWith(signinRedirectPath);
 
     expect(logger.infoRequest).toHaveBeenCalledTimes(1);
```
mwestacott, line 56: You could also test that the call to `logger.info` is made here (with that different status code output).

src/middleware/authentication.middleware.ts
mouhajer-ch: Both events happens and it is important to log them both. Do you have any suggestions?

test/middleware/authentication.middleware.spec.ts
mouhajer-ch: Working on that, thanks :)

src/middleware/authentication.middleware.ts
mouhajer-ch: As you suggested I will remove the 401 and add one single comments with the following msg error `User not authenticated, redirecting to sign in page, status_code=302`

mwestacott (approved):

PR #112, opened 2022-05-04
Feature/roe 420 add common address and date model

mouhajer-ch:
### JIRA link

[ROE-420](https://companieshouse.atlassian.net/browse/ROE-420)

### Change description

- Created Address and Date model
- Update BOs with Address and Date common model

### Work checklist


views/includes/inputs/address/principal-address-input.html
```diff
@@ -1,3 +1,5 @@
+{% set propertyNameNumberText = "For example, 'Saron House', '116' or 'Unit 37a'" %}
```
mwestacott, line 1: Just wondering what the value in doing this is? It seems to be only referenced the once, in this file only.

So, why do it for this bit of text and not all the others within the file?

mwestacott: Sorry, where's the other place - in which file?

mwestacott: 👍  thanks. Now I see it. I guess the only thing to be aware of is that if you use that `service-address-include.html` file you need to have the text constants defined in the files where it's being included.

mwestacott (approved):

PR #117, opened 2022-05-06
roe-413 Update Entity in web model

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-413

### Change description

Added field: is_on_register_in_country_formed_in

Removed address fields from model and re-reference common address models

Ensure that all fields in the model and template match the api spec

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/views/entity.html
+++ b/views/entity.html
@@ -82,6 +82,7 @@
       {% endset -%}
 
       {{ govukRadios({
+        classes: "govuk-radios",
         idPrefix: "is_on_register_in_country_formed_in",
         name: "is_on_register_in_country_formed_in",
         fieldset: {
@@ -97,11 +98,13 @@
             text: "Yes",
             conditional: {
                html: entityPublicRegisterHtml
-            }
+            },
+            checked: is_on_register_in_country_formed_in === 1
           },
           {
             value: 0,
-            text: "No"
+            text: "No",
+            checked: is_on_register_in_country_formed_in === 0
           }
         ]
       }) }}
--- a/views/entity.html
+++ b/views/entity.html
@@ -81,33 +81,8 @@
         {% include "includes/inputs/fields/registration-number-input.html" %}
       {% endset -%}
 
-      {{ govukRadios({
-        classes: "govuk-radios",
-        idPrefix: "is_on_register_in_country_formed_in",
-        name: "is_on_register_in_country_formed_in",
-        fieldset: {
-          legend: {
-            text: "Is the overseas entity already on a public register in the country it was formed in?",
-            isPageHeading: false,
-            classes: "govuk-fieldset__legend--m"
-          }
-        },
-        items: [
-          {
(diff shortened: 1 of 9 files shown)


views/entity.html
```diff
@@ -96,13 +97,11 @@ <h1 class="govuk-heading-xl">Tell us about the overseas entity</h1>
             text: "Yes",
             conditional: {
                html: entityPublicRegisterHtml
-            },
-            checked: is_on_sanctions_list === 1
+            }
```
mouhajer-ch, line 100: Hi Matthew, you should not remove the `checked` field, but you should update the controller. You can find example in other fields

mouhajer-ch (commented):
I would advice you to update first the SDK with the new field on Entity model and after do all the WEB changes

views/entity.html
mattch1: Put in a pr for the sdk: https://github.com/companieshouse/api-sdk-node/pull/384

Added checked fields to two templates as this was based on bo other and that was also missing them controller updated as well.

mouhajer-ch (commented):
Thanks for the update, may I ask you please to add an include for `is_on_register_in_country_formed_in` field and remove the condition with the phoneHtml that is not necessary. It is used on BO other and Entity page. Because the text will change add as you know the` {% set registerInCountryFormedText = "Is the other legal entity already on a public register in the country it was formed in?" %}`

Should be something like this

```

        {{ govukRadios({
          classes: "govuk-radios",
          idPrefix: "is_on_register_in_country_formed_in",
          name: "is_on_register_in_country_formed_in",
          fieldset: {
            legend: {
              text: registerInCountryFormedText,
              isPageHeading: false,
              classes: "govuk-fieldset__legend--m"
            }
          },
          items: [
            {
              value: 1,
              text: "Yes",
              conditional: {
                html: entityPublicRegisterHtml
              },
              checked: is_on_register_in_country_formed_in === 1
            },
            {
              value: 0,
              text: "No",
              checked: is_on_register_in_country_formed_in === 0
            }
          ]
        }) }}
```

mattch1:
I have had a look at the includes and what I would like to do is raise a separate ticket.

The reason for this is I think it would be best to include all the registration fields in this includes file together

public_register_name
registration_number
and is_on_register_in_country_formed_in

However, in order to do this we'd need some consistencey in bo other the name "register_name" is used and in the entity it is "public_register_name". A discussion on this came up yesterday in 

https://github.com/companieshouse/overseas-entities-api/pull/44

We are thinking of adding a ticket to do this work and we should probably keep it together. I can raise a ticket if you want. I'll discuss this with @markpit and @mwestacott as well.

mattch1:
What I can do for now is as suggested have an includes file just for the radio buttons but with a view to add those other two feilds as part of this other change.

mouhajer-ch (approved):

PR #129, opened 2022-05-16
Feature/roe 550 add payment service api

mouhajer-ch:
### JIRA link

Resolves [ROE-78](https://companieshouse.atlassian.net/browse/ROE-78), [ROE-247](https://companieshouse.atlassian.net/browse/ROE-247) and [ROE-527](https://companieshouse.atlassian.net/browse/ROE-527)

### Change description

- Integrate GOV Pay Service

### Work checklist

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -20,7 +20,7 @@ export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentValue("SHOW_SERVICE_OFFLI
 export const API_URL = getEnvironmentValue("API_URL");
 export const CHS_URL = getEnvironmentValue("CHS_URL");
 export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
-export const FEATURE_FLAG_PAYMENT = getEnvironmentValue("FEATURE_FLAG_PAYMENT");
+export const FEATURE_FLAG_ENABLE_PAYMENT_16052022 = getEnvironmentValue("FEATURE_FLAG_ENABLE_PAYMENT_16052022");
 
 export const SERVICE_NAME = 'Register an overseas entity and tell us about its beneficial owners';
 export const DESCRIPTION = "Overseas Entities Transaction";
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -65,6 +65,6 @@ export const PAYMENT = "payment";
 export const TRANSACTION = "transaction";
 export const OVERSEAS_ENTITY = "overseas-entity";
 
-const TRANSACTION_PATH = "/" + TRANSACTION + "/:transactionId";
-const OVERSEAS_ENTITY_PATH = "/" + OVERSEAS_ENTITY + "/:overseaEntityId/";
+const TRANSACTION_PATH = TRANSACTION + "/:transactionId/";
+const OVERSEAS_ENTITY_PATH = OVERSEAS_ENTITY + "/:overseaEntityId/";
 export const PAYMENT_WITH_TRANSACTION_URL = REGISTER_AN_OVERSEAS_ENTITY_URL + TRANSACTION_PATH + OVERSEAS_ENTITY_PATH + PAYMENT;
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -20,6 +20,7 @@ export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentValue("SHOW_SERVICE_OFFLI
 export const API_URL = getEnvironmentValue("API_URL");
 export const CHS_URL = getEnvironmentValue("CHS_URL");
 export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
+export const FEATURE_FLAG_PAYMENT = getEnvironmentValue("FEATURE_FLAG_PAYMENT");
 
 export const SERVICE_NAME = 'Register an overseas entity and tell us about its beneficial owners';
 export const DESCRIPTION = "Overseas Entities Transaction";
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -23,6 +23,7 @@ export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
 export const SERVICE_NAME = 'Register an overseas entity and tell us about its beneficial owners';
 export const DESCRIPTION = "Overseas Entities Transaction";
 export const REFERENCE = "OverseasEntitiestReference";
+export const PAYMENT_REQUIRED = "x-payment-required";
 
 // TEMPLATEs
 export const BENEFICIAL_OWNER_GOV_PAGE = "beneficial-owner-gov";
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -24,7 +24,7 @@ export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
 export const SERVICE_NAME = 'Register an overseas entity and tell us about its beneficial owners';
 export const DESCRIPTION = "Overseas Entities Transaction";
 export const REFERENCE = "OverseasEntitiestReference";
-export const PAYMENT_REQUIRED_HEADER = "x-payment-required";
+export const PAYMENT_REQUIRED_HEADER = "X-Payment-Required";
 export const PAYMENT_PAID = "paid";
 
 // TEMPLATEs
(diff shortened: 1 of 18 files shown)


src/config/index.ts
```diff
@@ -18,11 +18,15 @@ export const COOKIE_DOMAIN = getEnvironmentValue("COOKIE_DOMAIN");
 export const CACHE_SERVER = getEnvironmentValue("CACHE_SERVER");
 export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentValue("SHOW_SERVICE_OFFLINE_PAGE");
 export const API_URL = getEnvironmentValue("API_URL");
+export const CHS_URL = getEnvironmentValue("CHS_URL");
 export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
+export const FEATURE_FLAG_ENABLE_PAYMENT_16052022 = getEnvironmentValue("FEATURE_FLAG_ENABLE_PAYMENT_16052022");
```
mwestacott, line 23: Not sure if you've done this already, but we need to create a story on our backlog to remove this feature flag when it's no longer required.

mouhajer-ch: Thanks, will do for the WEB and API

src/controllers/payment.controller.ts
```diff
@@ -0,0 +1,43 @@
+import { NextFunction, Request, Response } from "express";
+
+import { logger, createAndLogErrorRequest } from "../utils/logger";
+import { CHECK_YOUR_ANSWERS_URL, CONFIRMATION_URL, PAYMENT_PAID } from "../config";
+import { ApplicationData } from "../model";
+import { getApplicationData } from "../utils/application.data";
+import { Payment, PaymentKey } from "../model/payment.type.model";
+
+// The Payment Platform will redirect the user's browser back to the `redirectUri` supplied when the payment session was created,
```
mwestacott, line 9: 👍  liking the comments. The whole payment process is so decoupled with redirects that these will help with maintenance.

src/controllers/payment.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+    // sent by the application to the Payment Platform. Protection against CSRF
+    if ( !savedPayment.state || savedPayment.state !== state) {
+      return next(createAndLogErrorRequest(req, `Rejecting redirect, state does not match. Payment Request:  ${ JSON.stringify(savedPayment) } `));
+    }
+
+    // Validate the status of the payment
+    if (status === PAYMENT_PAID) {
+      logger.debugRequest(req, `OE id: ${ savedPayment.overseasEntityId }, Payment status: ${status},Redirecting to: ${CONFIRMATION_URL}`);
+
+      // Payment Successfull, redirect to confirmation page
```
mwestacott, line 30: Successful (single 'l'). Sorry.

src/controllers/payment.controller.ts
mouhajer-ch: Thanks, following https://github.com/companieshouse/dapperdox.developer.ch.gov.uk/blob/master/assets/sections/payment-api/templates/guides/overview.md

src/controllers/payment.controller.ts
mouhajer-ch: will do

src/service/payment.service.ts
```diff
@@ -0,0 +1,87 @@
...
+const setPaymentRequest = (transactionId, overseasEntityId): CreatePaymentRequest => {
+
+  const paymentResourceUri = `${API_URL}/transactions/${transactionId}/${PAYMENT}`;
+
+  const baseURL = `${CHS_URL}${REGISTER_AN_OVERSEAS_ENTITY_URL}`;
+  const reference = `${REFERENCE}_${transactionId}`;
+
+  // Once payment has been taken, the platform redirects the user back to the application,
+  // using the application supplied `redirectUri`.
+  const redirectUri = `${baseURL}${TRANSACTION}/${transactionId}/${OVERSEAS_ENTITY}/${overseasEntityId}/${PAYMENT}`;
```
mwestacott, line 79: It looks like this URL closely matches the URL you defined in your new constant above - PAYMENT_WITH_TRANSACTION_URL - albeit with substituted values. Just wondering, rather than repeat the format of it here, could you somehow reference and use that constant and substitute the transaction and OE ids in it (because if the format did ever change, it would need updating in two places)?

mouhajer-ch: Hi Martin, could you please be more clear, I do not see any URL closely match

mwestacott: So that constant is built up from:

```
const TRANSACTION_PATH = TRANSACTION + "/:transactionId/";
const OVERSEAS_ENTITY_PATH = OVERSEAS_ENTITY + "/:overseaEntityId/";
export const PAYMENT_WITH_TRANSACTION_URL = REGISTER_AN_OVERSEAS_ENTITY_URL + TRANSACTION_PATH + OVERSEAS_ENTITY_PATH + PAYMENT;
```

Which I think resolves to: '**/register-an-overseas-entity/transaction/:transactionId/overseas-entity/:overseasEntityId/payment**'

So it looks like it's the same format as the '**redirectUri**' you build.

Give me a call if still unclear and we can have a chat...

mouhajer-ch: The `PAYMENT_WITH_TRANSACTION_URL` is the pattern used for our endpoint, `redirectUri` will have instead the field correctly populated with the `transactionId` and `overseaEntityId` and has to match the endpoint.
The first is always the same, the second is changing every time.

Lets have a call, you know I love them :D

src/service/payment.service.ts
```diff
@@ -0,0 +1,87 @@
...
+      // To initiate the web journey through which the user will interact to make the payment
+      return paymentResource.links.journey;
+    }
+  } else {
+    // Only if transaction does not have a fee or PAYMENT flagged.
+    return CONFIRMATION_URL;
+  }
+};
+
+const setPaymentRequest = (transactionId, overseasEntityId): CreatePaymentRequest => {
```
markpit, line 70: line 70, the params need types eg. transactionId: string

src/controllers/check.your.answers.controller.ts
```diff
@@ -37,10 +38,13 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
     const overseaEntity: OverseasEntityCreated = await createOverseasEntity(req, req.session as Session, transaction.id as string);
     logger.infoRequest(req, `Overseas Entity Created, ID: ${overseaEntity.id}`);
 
-    await closeTransaction(req, req.session as Session, transaction.id as string, overseaEntity.id);
+    const transactionClosedResponse = await closeTransaction(req, req.session as Session, transaction.id as string, overseaEntity.id);
     logger.infoRequest(req, `Transaction Closed, ID: ${transaction.id}`);
 
-    return res.redirect(config.CONFIRMATION_URL);
+    const redirectPath = await startPaymentsSession(req, req.session as Session, transaction.id as string, overseaEntity.id, transactionClosedResponse);
+    logger.infoRequest(req, `Payments Session ended with, Trans_ID: ${transaction.id}, OE_ID: ${overseaEntity.id}. Redirect to: ${redirectPath}`);
```
markpit, line 45: line 45, I think the 'Payments Session ended' is misleading, as it makes me think that the payment has been taken. I think it would read better if it was 'Payments Session created' which would also match up with the function name startPaymentsSession

src/controllers/payment.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const appData: ApplicationData = getApplicationData(req.session);
+    const savedPayment = appData[PaymentKey] || {} as Payment;
+
+    logger.debugRequest(req, `Returned state: ${ state }, saved state: ${savedPayment.state}`);
+
+    // The application must ensure that the returned `state` matches the nonce
+    // sent by the application to the Payment Platform. Protection against CSRF
+    if ( !savedPayment.state || savedPayment.state !== state) {
+      return next(createAndLogErrorRequest(req, `Rejecting redirect, state does not match. Payment Request:  ${ JSON.stringify(savedPayment) } `));
```
markpit, line 23: line 23 , in the log message , i'd add a few 'payment's to it -  
Rejecting payment redirect, payment state does not match. Payment Request: etc..

src/controllers/payment.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    // Validate the status of the payment
+    if (status === PAYMENT_PAID) {
+      logger.debugRequest(req, `OE id: ${ savedPayment.overseasEntityId }, Payment status: ${status},Redirecting to: ${CONFIRMATION_URL}`);
```
markpit, line 28: line 28 and 33, Change OE id to Overseas Entity id, as someone who has not worked on teh project might have to look at the logs and might be unsure what an OE id was

mwestacott (approved):

PR #151, opened 2022-05-30
Roe 432 create secure stop page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-432

### Change description

Added stop page for paper filing.
Navigation to be handled in case https://companieshouse.atlassian.net/browse/ROE-431

### Work checklist

- [x] Tests added where applicable
- [x] UI changes meet accessibility criteria

### Merge instructions

We are committed to keeping commit history clean, consistent and linear. To achieve this, this commit should be structured as follows:

```
<type>[optional scope]: <description>
```

and contain the following structural elements:

- fix: a commit that patches a bug in your codebase (this correlates with PATCH in semantic versioning),
- feat: a commit that introduces a new feature to the codebase (this correlates with MINOR in semantic versioning),
- BREAKING CHANGE: a commit that has a footer `BREAKING CHANGE:` introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type,
- types other than `fix:` and `feat:` are allowed, for example `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`, and others,
- footers other than `BREAKING CHANGE: <description>` may be provided.

PR #152, opened 2022-05-30
ROE-555 add sold land filter and cannot use page with validation

SVarallo:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-555

### Change description
Covers works for ROE-555 to add the sold land filter page as well as work for ROE-595 to add the stop screen. Also adds basic validation for no radio button selected. In addition I renamed the 'cannot-use' page added in https://github.com/companieshouse/overseas-entities-web/pull/151/files to 'use-paper' to match the file structure present in the prototype. It may be possible in future to make a more generic cannot-use page that we can use for both cases.

### Work checklist

### Merge instructions

```
```


src/controllers/sold.land.filter.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+  } catch (error) {
+    logger.errorRequest(req, error);
+    next(error);
+  }
+};
+
+export const post = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `POST ${config.SOLD_LAND_FILTER_PAGE}`);
+
```
mouhajer-ch, line 22: To improve readability I would have used 
`const { has_sold_land } = req.body`

src/controllers/sold.land.filter.controller.ts
```diff
@@ -0,0 +1,34 @@
...
+    });
+  } catch (error) {
+    logger.errorRequest(req, error);
+    next(error);
+  }
+};
+
+export const post = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `POST ${config.SOLD_LAND_FILTER_PAGE}`);
```
mouhajer-ch, line 21: We need to store the information saved, it is not a blocker can be done in a separate story.

views/sold-land-filter.html
```diff
@@ -0,0 +1,60 @@
...
+          items: [
+            {
+              value: 1,
+              text: "Yes",
+              checked: has_sold_land == 1
+            },
+            {
+              value: 0,
+              text: "No",
+              checked: has_sold_land == 0
```
mouhajer-ch, line 43: We need to return the value in the `render` call if we want to visualise what has been selected

test/controllers/sold.land.filter.controller.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    jest.clearAllMocks();
+  });
+
+  describe("GET tests", () => {
+    test(`renders the ${config.SOLD_LAND_FILTER_PAGE} page`, async () => {
+      const resp = await request(app).get(config.SOLD_LAND_FILTER_URL);
+
+      expect(resp.status).toEqual(200);
+      expect(resp.text).toContain(SOLD_LAND_FILTER_PAGE_TITLE);
+    });
```
mouhajer-ch, line 22: Add a simple unit test if you want to catch errors during the GET, something like
```
    test(`catch error when renders the ${config.SOLD_LAND_FILTER_PAGE} page on GET method`, async () => {
      mock_ANY_CALL_ON_GET_METHOD.mockImplementationOnce( () => { throw new Error(ANY_MESSAGE_ERROR); });
      const resp = await request(app).get(config.SOLD_LAND_FILTER_URL);

      expect(resp.status).toEqual(500);
      expect(resp.text).toContain(SERVICE_UNAVAILABLE);
    });
```

test/controllers/sold.land.filter.controller.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    test("renders the current page with error message", async () => {
+      const resp = await request(app)
+        .post(config.SOLD_LAND_FILTER_URL);
+
+      expect(resp.status).toEqual(200);
+      expect(resp.text).toContain(SOLD_LAND_FILTER_PAGE_TITLE);
+      expect(resp.text).toContain(ErrorMessages.SELECT_IF_ENTITY_HAS_SOLD_LAND);
+    });
```
mouhajer-ch, line 51: Add a simple unit test to cover the catch, similar to the example above

mouhajer-ch (commented):
Important note, now we need to integrate the navigation checking as soon as possible. We should have already the story, postponed for MVP

src/controllers/sold.land.filter.controller.ts
mouhajer-ch: It is work for another story :)

test/controllers/sold.land.filter.controller.spec.ts
SVarallo: Done 👍

views/sold-land-filter.html
mouhajer-ch: We can leave it, the logic will be completed when we store the info into the session and returned on the `render`

mouhajer-ch (approved):

PR #157, opened 2022-05-31
Roe 431 protect personal info page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-431

### Work checklist

### Merge instructions

```
```


src/controllers/secure.register.filter.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  } catch (error) {
+    logger.errorRequest(req, error);
+    next(error);
+  }
+};
+
+export const post = (req: Request, res: Response, next: NextFunction) => {
+  logger.debugRequest(req, `POST ${config.SECURE_REGISTER_FILTER_PAGE}`);
+  try {
+    if (req.body.secure_register === '1') {
```
SVarallo, line 20: Similar to what Moh said on my PR you could put this into it's own variable to use in the if statements

src/controllers/secure.register.filter.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+    next(error);
+  }
+};
+
+export const post = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `POST ${config.SECURE_REGISTER_FILTER_PAGE}`);
+    const isSecureRegister = req.body.secure_register;
+    if (isSecureRegister === '1') {
+      return res.render(config.USE_PAPER_PAGE, {
```
markpit, line 22: line 22 - i think this will need to be a redirect to the paper page. If we use render, then the page url won't change and the matomo journey tracking won't pick up that they have essentially moved to a new page (we had to change the stop pages to redirects in confirmation-statement for matomo)

SVarallo: Ah right, we'll have to do something similar for the other page I added too. Think it's possible to make a generic cannot use page? Otherwise it could result in a whole load of different controllers

mattch1: It's not a lot of code, it's probably more readable for anyone picking this up in the future to keep them separate.

markpit:
package-lock.json should be removed, i think it was accidentally included

test/controllers/use.paper.controller.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { expect, test } from "@jest/globals";
+import * as config from "../../src/config";
+import request from "supertest";
+import app from "../../src/app";
+
```
markpit, line 5: similar comment here , needs an outer 'describe' to describe what controller the tests in this file are for

PR #180, opened 2022-06-10
ROE-704 Validation for presenter full name characters

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-704

### Change description

Added the otherDataString regex and appended it to the express validation.

Had to put the regex string in the same file; for some reason if you try to export it from elsewhere the validation no longer works.

In the regex string had to remove '\' before the '[' character as sonar flagged it as an unnecessary escape character.

### Work checklist

### Merge instructions

```
```


src/validation/presenter.validation.ts
```diff
@@ -2,7 +2,9 @@ import { body } from "express-validator";
 
 import { ErrorMessages } from "./error.messages";
 
+const VALID_CHARACTERS = /^[-,.:; 0-9A-Z&@$£¥€'"«»''""?!/\\()[\]{}<>*=#%+ÀÁÂÃÄÅĀĂĄÆǼÇĆĈĊČÞĎÐÈÉÊËĒĔĖĘĚĜĞĠĢĤĦÌÍÎÏĨĪĬĮİĴĶĹĻĽĿŁÑŃŅŇŊÒÓÔÕÖØŌŎŐǾŒŔŖŘŚŜŞŠŢŤŦÙÚÛÜŨŪŬŮŰŲŴẀẂẄỲÝŶŸŹŻŽa-zÀÖØſƒǺǿẀẅỲỳ]*$/;
```
mwestacott, line 5: Having this here is probably fine for this PR as it's the first occurrence of its use, but in the next PR, we don't want to keep duplicating this string so we'll have to then move it somewhere central and work out how to get validation working with the imported constant value.

src/validation/error.messages.ts
```diff
@@ -56,5 +56,7 @@ export enum ErrorMessages {
   MAX_PUBLIC_REGISTER_NAME_LENGTH = "Name of register must be 4000 characters or less",
   MAX_PUBLIC_REGISTER_NUMBER_LENGTH = "Entity’s registration number must be 32 characters or less",
   MAX_OCCUPATION_LENGTH = "Occupation must be 100 characters or less",
-  MAX_ROLE_LENGTH = "Role and responsibilities must be 4000 characters or less"
+  MAX_ROLE_LENGTH = "Role and responsibilities must be 4000 characters or less",
+  // Invalid characters
+  INVALID_CHARACTERS = "Full name must only include letters a to z, numbers, and special characters such as hyphens, spaces and apostrophes"
```
markpit, line 61: I would name this INVALID_CHARACTERS_FULL_NAME as the text relates to the Full name field

mattch1: Have changed this, however, it could be applied to other fields.

markpit (dismissed):

src/validation/regex/regex.validation.ts
```diff
@@ -0,0 +1 @@
+export const INVALID_CHARACTERS_FULL_NAME = /^[-,.:; 0-9A-Z&@$£¥€'"«»''""?!/\\()[\]{}<>*=#%+ÀÁÂÃÄÅĀĂĄÆǼÇĆĈĊČÞĎÐÈÉÊËĒĔĖĘĚĜĞĠĢĤĦÌÍÎÏĨĪĬĮİĴĶĹĻĽĿŁÑŃŅŇŊÒÓÔÕÖØŌŎŐǾŒŔŖŘŚŜŞŠŢŤŦÙÚÛÜŨŪŬŮŰŲŴẀẂẄỲÝŶŸŹŻŽa-zÀÖØſƒǺǿẀẅỲỳ]*$/;
```
mwestacott, line 1: If this has been moved somewhere central to be generic and for reuse then suggest changing constant name to remove _FULL_NAME at the end.

markpit: yes, I missed that, I thought the regex was for valid chars? so should just be VALID_CHARACTERS

mattch1: More or less said the same above, so changed this.

src/validation/presenter.validation.ts
```diff
@@ -1,8 +1,9 @@
 import { body } from "express-validator";
 
 import { ErrorMessages } from "./error.messages";
+import { VALID_CHARACTERS } from "./regex/regex.validation";
 
 export const presenter = [
-  body("full_name").not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.FULL_NAME).isLength({ max: 160 }).withMessage(ErrorMessages.MAX_FULL_NAME_LENGTH),
+  body("full_name").not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.FULL_NAME).isLength({ max: 160 }).withMessage(ErrorMessages.MAX_FULL_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.INVALID_CHARACTERS),
```
mwestacott: .matches(VALID_CHARACTERS).withMessage(ErrorMessages.INVALID_CHARACTERS)

doesn't make sense and I think the VALID_CHARACTERS constant should change its name back to INVALID_CHARACTERS.

mwestacott (approved):

PR #182, opened 2022-06-13
Roe 587 web char validation bo individual

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-587

### Change description
Introduced char validation to 9 fields and custom validation tot he 6 service address fields.

First name
Last name
Nationality
Property name or number
Address line 1
Address line 2 
City or town
County, state, province or region
Postcode or ZIP code 

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/beneficial.owner.individual.validation.ts
+++ b/src/validation/beneficial.owner.individual.validation.ts
@@ -7,6 +7,7 @@ import { VALID_CHARACTERS } from "./regex/regex.validation";
 export const beneficialOwnerIndividual = [
   body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_PREFIX + ErrorMessages.INVALID_CHARACTERS),
   body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_PREFIX + ErrorMessages.INVALID_CHARACTERS),
+  body("nationality").matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_PREFIX + ErrorMessages.INVALID_CHARACTERS),
 
   ...usual_residential_address_max_validations,
   ...usual_residential_service_address_max_validations
--- a/src/validation/beneficial.owner.individual.validation.ts
+++ b/src/validation/beneficial.owner.individual.validation.ts
@@ -5,8 +5,12 @@ import { usual_residential_service_address_beneficial_owner_validations, usual_r
 import { VALID_CHARACTERS } from "./regex/regex.validation";
 
 export const beneficialOwnerIndividual = [
-  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_INVALID_CHARACTERS),
-  body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
+  body("first_name").isLength({ max: 50 })
+    .withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH)
+    .matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_INVALID_CHARACTERS),
+  body("last_name").isLength({ max: 160 })
+    .withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH)
+    .matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
   body("nationality").matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_INVALID_CHARACTERS),
 
   ...usual_residential_address_beneficial_owner_validations,
--- a/src/validation/beneficial.owner.individual.validation.ts
+++ b/src/validation/beneficial.owner.individual.validation.ts
@@ -5,9 +5,9 @@ import { usual_residential_service_address_beneficial_owner_validations, usual_r
 import { VALID_CHARACTERS } from "./regex/regex.validation";
 
 export const beneficialOwnerIndividual = [
-  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_PREFIX + ErrorMessages.INVALID_CHARACTERS),
-  body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_PREFIX + ErrorMessages.INVALID_CHARACTERS),
-  body("nationality").matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_PREFIX + ErrorMessages.INVALID_CHARACTERS),
+  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_INVALID_CHARACTERS),
+  body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
+  body("nationality").matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_INVALID_CHARACTERS),
 
   ...usual_residential_address_beneficial_owner_validations,
   ...usual_residential_service_address_beneficial_owner_validations
--- a/src/validation/beneficial.owner.individual.validation.ts
+++ b/src/validation/beneficial.owner.individual.validation.ts
@@ -2,10 +2,11 @@ import { body } from "express-validator";
 
 import { ErrorMessages } from "./error.messages";
 import { usual_residential_service_address_max_validations, usual_residential_address_max_validations } from "./fields/address.validation";
+import { VALID_CHARACTERS } from "./regex/regex.validation";
 
 export const beneficialOwnerIndividual = [
-  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH),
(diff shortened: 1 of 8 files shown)


src/validation/error.messages.ts
```diff
@@ -58,5 +58,15 @@ export enum ErrorMessages {
   MAX_OCCUPATION_LENGTH = "Occupation must be 100 characters or less",
   MAX_ROLE_LENGTH = "Role and responsibilities must be 4000 characters or less",
   // Invalid characters
-  INVALID_CHARACTERS_FULL_NAME = "Full name must only include letters a to z, numbers, and special characters such as hyphens, spaces and apostrophes"
+  INVALID_CHARACTERS = " must only include letters a to z, numbers, and special characters such as hyphens, spaces and apostrophes",
+  FULL_NAME_PREFIX = "Full name",
```
mwestacott, line 62: Rather than do the string constant validation in all the .validation files I'd suggest defining the strings (using concatenation) here and just referencing them. E.g.

FIRST_NAME_INVALID_CHARACTERS = "First name" + INVALID_CHARACTERS,

src/validation/fields/address.validation.ts
```diff
@@ -81,25 +86,31 @@ export const principal_service_address_max_validations = [
 ];
 
 export const usual_residential_address_max_validations = [
```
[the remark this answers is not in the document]
mattch1: The name was already taken to opted for "usual_residential_address_beneficial_owner_validations" as this is used specifically for beneficial owner validation.

src/validation/error.messages.ts
mwestacott: That's a shame. Should we consider changing this to just a list of string constants then (where presumably string concatenation would be allowed)? It's just that FULL_NAME_PREFIX on its own isn't really an error message and neither is INVALID_CHARACTERS. Or, are we happy that the `ErrorMessages` enum is just a bucket of strings for error messages which can define the entire message or sometimes need to be concatenated together.

test/controllers/presenter.controller.spec.ts
```diff
@@ -91,7 +91,7 @@ describe("PRESENTER controller", () => {
 
       expect(resp.status).toEqual(200);
       expect(resp.text).toContain(PRESENTER_PAGE_TITLE);
-      expect(resp.text).toContain(ErrorMessages.INVALID_CHARACTERS_FULL_NAME);
+      expect(resp.text).toContain( ErrorMessages.FULL_NAME_PREFIX + ErrorMessages.INVALID_CHARACTERS);
```
mwestacott, line 94: Again, space can go.

test/__mocks__/validation.mock.ts
```diff
@@ -83,6 +104,30 @@ export const BENEFICIAL_OWNER_INDIVIDUAL_WITH_MAX_LENGTH_FIELDS_MOCK = {
...
+  is_on_sanctions_list: "0",
+  is_service_address_same_as_usual_residential_address: "1",
+  ...RESIDENTIAL_ADDRESS_WITH_INVALID_CHAR_FIELDS_MOCK,
+  ...SERVICE_ADDRESS_WITH_INVALID_CHAR_FIELDS_MOCK,
+  ...DATE_OF_BIRTH_MOCK,
+  ...START_DATE_MOCK
+};
+
+export const BENEFICIAL_OWNER_INDIVIDUAL_WITH_INVALID_CHARS_SERVICE_ADDRESS_MOCK = {
+  first_name: MAX_50 + "1",
```
mwestacott, line 120: Do you really want to exceed max length for these fields? This data structure is concerned with testing invalid chars (and only in the service address).

src/validation/custom.validation.ts
```diff
@@ -14,6 +16,13 @@ export const checkMaxFieldIfRadioButtonSelected = (selected: boolean, errMsg: st
   return true;
 };
 
+export const checkInvalidCharactersIfRadioButtonSelected = (selected: boolean, errMsg: string, value: string) => {
+  if (value && !(selected || value.match(VALID_CHARACTERS))) {
```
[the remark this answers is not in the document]
mattch1: You have to check it the value is truthy if you done it will throw an error we don't want this we only want to throw it when something invalid has been entered.

test/__mocks__/validation.mock.ts
mattch1: No this should be changed.

src/validation/error.messages.ts
mattch1: I had similar thoughts but I'd prefer to raise a ticket and link to this one as it seems like a wider issue to have come out of this work.

src/validation/fields/address.validation.ts
```diff
@@ -80,26 +85,32 @@ export const principal_service_address_max_validations = [
...
+  body("usual_residential_address_line_2").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_ADDRESS_LINE2_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.ADDRESS_LINE_2_PREFIX + ErrorMessages.INVALID_CHARACTERS),
+  body("usual_residential_address_town").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_CITY_OR_TOWN_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.CITY_OR_TOWN_PREFIX + ErrorMessages.INVALID_CHARACTERS),
+  body("usual_residential_address_county").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_COUNTY_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.COUNTY_STATE_PROVINCE_REGION_PREFIX + ErrorMessages.INVALID_CHARACTERS),
+  body("usual_residential_address_postcode").isLength({ max: 20 }).withMessage(ErrorMessages.MAX_POSTCODE_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.POSTCODE_ZIPCODE_PREFIX + ErrorMessages.INVALID_CHARACTERS)
 ];
 
-export const usual_residential_service_address_max_validations = [
+export const usual_residential_service_address_beneficial_owner_validations = [
```
SVarallo, line 97: is there a risk that with making this common validation method specific for BOs that we're going to have to do this with every page that we has address fields in it? I'm just concerned this file will get far bigger than necessary

src/validation/custom.validation.ts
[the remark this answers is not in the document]
mattch1: Not sure it it were falsey then its length would be < max value so it is covered by that test.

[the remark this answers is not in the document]
markpit: I think the logic is the wrong way around. Looking at checkMaxFieldIfRadioButtonSelected, it should throw the error message if the string is too long. 
For checkInvalidCharactersIfRadioButtonSelected - 
checkInvalidCharactersIfRadioButtonSelected(true, "err", "Путин") - returns true (as in it the text is ok)
checkInvalidCharactersIfRadioButtonSelected(false, "err", "Путин") - throws error - (selected is false so we shouldn't be checking the field)

src/validation/error.messages.ts
markpit: I think to keep to follow the enum standard, you could have the prefix and INVALID_CHARACTERS in the same file but moved outside of the enum (and don't export them), then in the enum, you could add extra error message types by concatenating the strings like
INVALID_CHARACTERS_FIRST_NAME = FIRST_NAME_PREFIX + INVALID_CHARACTERS.
That way INVALID_CHARACTERS_FIRST_NAME would still be a'type' of errorMessage

src/validation/custom.validation.ts
```diff
@@ -16,8 +16,8 @@ export const checkMaxFieldIfRadioButtonSelected = (selected: boolean, errMsg: st
   return true;
 };
 
-export const checkInvalidCharactersIfRadioButtonSelected = (selected: boolean, errMsg: string, value: string) => {
-  if (value && !(selected || value.match(VALID_CHARACTERS))) {
+export const checkInvalidCharactersIfRadioButtonSelected = (noSelected: boolean, errMsg: string, value: string) => {
+  if (noSelected && !value.match(VALID_CHARACTERS)) {
```
mwestacott, line 20: I still don't know why this function isn't just implemented in a similar way to the `checkMaxFieldIfRadioButtonSelected` function above. Surely it's contract is the same - throw an error if radio button selected and invalid characters are found, return true otherwise...

test/controllers/beneficial.owner.individual.controller.spec.ts
```diff
@@ -165,7 +164,7 @@ describe("BENEFICIAL OWNER INDIVIDUAL controller", () => {
       expect(resp.text).toContain(ErrorMessages.COUNTY_STATE_PROVINCE_REGION_PREFIX + ErrorMessages.INVALID_CHARACTERS);
       expect(resp.text).toContain(ErrorMessages.POSTCODE_ZIPCODE_PREFIX + ErrorMessages.INVALID_CHARACTERS);
     });
-
+    /*
```
mwestacott, line 167: Wondering why this test is now commented out?

src/validation/beneficial.owner.individual.validation.ts
```diff
@@ -1,12 +1,14 @@
...
 
 import { ErrorMessages } from "./error.messages";
-import { usual_residential_service_address_max_validations, usual_residential_address_max_validations } from "./fields/address.validation";
+import { usual_residential_service_address_beneficial_owner_validations, usual_residential_address_beneficial_owner_validations } from "./fields/address.validation";
+import { VALID_CHARACTERS } from "./regex/regex.validation";
 
 export const beneficialOwnerIndividual = [
-  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH),
-  body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH),
+  body("first_name").isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_INVALID_CHARACTERS),
```
mwestacott, line 8: This is minor and not strictly related to your change, but I find that indenting the validations a bit like this make them a lot more readable:

body("first_name")
  .isLength({ max: 50 }).withMessage(ErrorMessages.MAX_FIRST_NAME_LENGTH)
.matches(VALID_CHARACTERS).withMessage(ErrorMessages.FIRST_NAME_INVALID_CHARACTERS),

Formatting a bit odd in the GitHub comment window, but basically putting each validation check/message pair on a separate indented line. What do you think?

src/validation/beneficial.owner.individual.validation.ts
```diff
@@ -1,12 +1,14 @@
...
+  body("last_name").isLength({ max: 160 }).withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH).matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
+  body("nationality").matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_INVALID_CHARACTERS),
 
-  ...usual_residential_address_max_validations,
-  ...usual_residential_service_address_max_validations
+  ...usual_residential_address_beneficial_owner_validations,
```
mwestacott, line 12: Again, minor. Sorry I'm being picky!

Could we just name this `usual_residential_address_beneficial_owner_validation` (and same for other). Pluralisation isn't needed - 'validation' can mean one or more checks being applied.

markpit (approved):

PR #188, opened 2022-06-15
Delete session data on confirmation page

mouhajer-ch:
### JIRA link

Resolves [ROE-775](https://companieshouse.atlassian.net/browse/ROE-775)

### Change description

Delete session data on confirmation page

### Work checklist

### DIFF
--- a/src/controllers/confirmation.controller.ts
+++ b/src/controllers/confirmation.controller.ts
@@ -3,7 +3,7 @@ import { Request, Response } from "express";
 import { logger } from "../utils/logger";
 import { CONFIRMATION_PAGE } from "../config";
 import { getLoggedInUserEmail } from "../utils/session";
-import { getApplicationData } from "../utils/application.data";
+import { deleteApplicationData, getApplicationData } from "../utils/application.data";
 import { ApplicationData } from "../model/application.model";
 import { Transactionkey } from "../model/data.types.model";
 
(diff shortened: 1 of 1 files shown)


src/controllers/confirmation.controller.ts
```diff
@@ -3,17 +3,20 @@ import { Request, Response } from "express";
...
 import { ApplicationData } from "../model/application.model";
 import { Transactionkey } from "../model/data.types.model";
 
 export const get = (req: Request, res: Response) => {
   logger.debugRequest(req, `GET ${CONFIRMATION_PAGE}`);
 
   const appData: ApplicationData = getApplicationData(req.session);
+  const referenceNumber = appData[Transactionkey];
+
+  deleteApplicationData(req.session);
```
markpit, line 16: could you add a test to expect that the deleteApplicationData gets called? or test that data has been removed from session on a call to 'get'?

mouhajer-ch: Which kind of test do you want to do?

markpit: i would probably mock the  deleteApplicationData and use expect to test it gets called. if deleteApplicationData is in the same module as other functions that you don't want to mock like getApplicationData, then I think you can use spy to mock just the one function

mouhajer-ch: I do not think checking that the mock of `deleteApplicationData` is called will improve test, but I will add it tomorrow

markpit: I think we need it, I've just removed  deleteApplicationData(req.session); and its import and the tests still pass. If there is a requirement for the get function to delete the data then we should have a test that confirms it does that.

mouhajer-ch: Thanks mate, added missing unit test

PR #198, opened 2022-06-16
Feature/trust page

rsachdevch:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-504

### Change description
Add trust page to allow users to select the beneficial owner associated with the trust
and upload trust information as JSON within the textarea

### Work checklist


markpit:
the package-lock.json file isn't needed, are you able to remove it?

test/controllers/beneficial.owner.type.controller.spec.ts
```diff
@@ -12,11 +12,14 @@ import * as config from "../../src/config";
 import { getApplicationData } from '../../src/utils/application.data';
 import {
   BENEFICIAL_OWNER_TYPE_PAGE_HEADING,
+  CHECK_YOUR_ANSWERS_LINK,
+  SERVICE_UNAVAILABLE,
```
markpit, line 16: just running the tests, it has highlighted that SERVICE_UNAVAILABLE is a duplicate import

test/controllers/beneficial.owner.type.controller.spec.ts
```diff
@@ -12,11 +12,14 @@ import * as config from "../../src/config";
...
+  TRUST_INFORMATION_LINK
```
markpit, line 17: comma is needed after TRUST_INFORMATION_LINK. if you run 'npm run lint' it should show that - we probably forgot to mention the lint tool is available

src/controllers/check.your.answers.controller.ts
```diff
@@ -18,9 +20,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
+    }
+
+    let backLinkUrl: string = config.BENEFICIAL_OWNER_TYPE_URL;
+    if (hasTrusts) {
+      backLinkUrl = config.TRUST_INFO_PAGE;
+    }
+
     return res.render(config.CHECK_YOUR_ANSWERS_PAGE, {
-      backLinkUrl: config.BENEFICIAL_OWNER_TYPE_URL,
+      backLinkUrl: backLinkUrl,
```
markpit, line 35: if the param name is same as variable name eg backLinkUrl: backLinkUrl, you can just use backLinkUrl on it's own, same for hasTrusts  eg
 return res.render(config.CHECK_YOUR_ANSWERS_PAGE, {
      backLinkUrl,
      templateName: config.CHECK_YOUR_ANSWERS_PAGE,
      hasTrusts,
      appData
    }

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -16,9 +18,16 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
+    let hasTrusts: boolean = false;
+
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUST_INFO_16062022)) {
+      hasTrusts = checkEntityHasTrusts(appData);
+    }
+
     return res.render(config.BENEFICIAL_OWNER_TYPE_PAGE, {
       backLinkUrl: config.BENEFICIAL_OWNER_STATEMENTS_URL,
       templateName: config.BENEFICIAL_OWNER_TYPE_PAGE,
+      hasTrusts: hasTrusts,
```
markpit, line 30: can replace `hasTrusts: hasTrusts,`   with just `hasTrusts,`

src/utils/navigation.ts
```diff
@@ -86,5 +86,10 @@ export const NAVIGATION: Navigation = {
     currentPage: config.MANAGING_OFFICER_CORPORATE_PAGE,
     previousPage: config.BENEFICIAL_OWNER_TYPE_URL,
     nextPage: config.BENEFICIAL_OWNER_TYPE_URL
+  },
+  [config.TRUST_INFO_URL]: {
+    currentPage: config.TRUST_INFO_PAGE,
+    previousPage: config.BENEFICIAL_OWNER_TYPE_URL,
+    nextPage: config.CHECK_YOUR_ANSWERS_URL
```
mwestacott, line 93: Just wondering if you need to change this section above:

```
[config.BENEFICIAL_OWNER_TYPE_URL]: {
    currentPage: config.BENEFICIAL_OWNER_TYPE_PAGE,
    previousPage: config.BENEFICIAL_OWNER_STATEMENTS_URL,
    nextPage: config.CHECK_YOUR_ANSWERS_URL
  },
```

Or does it stay as is since trust data may or may not be present.

src/utils/trusts.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (appData.beneficial_owners_corporate !== undefined) {
+      appData.beneficial_owners_corporate.forEach(element => {
+        if (element.trustees_nature_of_control_types !== undefined && element.trustees_nature_of_control_types.length > 0) {
+          hasTrusts = true;
+        }
+      });
+    }
+  }
+  return hasTrusts;
```
mwestacott, line 24: Minor, but blank line before return statement rather than at end of block maybe.

test/controllers/beneficial.owner.type.controller.spec.ts
```diff
@@ -37,13 +40,26 @@ describe("BENEFICIAL OWNER TYPE controller", () => {
...
   describe("GET tests", () => {
-    test("renders the beneficial owner type page for beneficial owners with all options", async () => {
+    test("renders the beneficial owner type page for beneficial owners with trusts", async () => {
       mockGetApplicationData.mockReturnValueOnce(APPLICATION_DATA_MOCK);
       const resp = await request(app).get(config.BENEFICIAL_OWNER_TYPE_URL);
 
       expect(resp.status).toEqual(200);
       expect(resp.text).toContain(BENEFICIAL_OWNER_TYPE_PAGE_HEADING);
       expect(resp.text).toContain(config.BENEFICIAL_OWNER_STATEMENTS_URL); // back button
+      expect(resp.text).not.toContain(CHECK_YOUR_ANSWERS_LINK); // continue button
```
mwestacott, line 50: Just checking this is right - both the 'check your answers' link and 'enter trust info' link should be present?

views/trust-information.html
```diff
@@ -0,0 +1,105 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Add trust information for the beneficial owners – {{ serviceName }} – GOV.UK
+{% endblock %}
+
+{% block beforeContent %}
```
markpit, line 7: need to change the `block beforeContent ` to `block backLink`
when you navigate through the page you'll notice the 'Alpha' banner dissappears on the new page as you are overriding the block beforeContent in layout.html. Just need to override the block backLink instead

views/trust-information.html
```diff
@@ -0,0 +1,105 @@
...
+        }) }}
+        {{ govukButton({
+          text: "Add another trust",
+          href: "/register-an-overseas-entity/trust-information"
+        }) }}
+
+        <p>
+          {{ govukButton({
+            text: "No more to add",
+            href: "/register-an-overseas-entity/check-your-answers",
```
markpit, line 95: could you add an 'id' attribute of 'submit' to the 'no more to add' button? (example here - https://github.com/companieshouse/overseas-entities-web/blob/main/views/beneficial-owner-type.html#L102)
It will allow our auto tests in the pipeline to pick up the button on the screen.

views/trust-information.html
```diff
@@ -0,0 +1,105 @@
...
+        </details>
+
+        {{ govukInsetText({
+          html: '
+            <h2 class="govuk-heading-m">Information shown on the public register</h2>
+            <p>None of the information you provide about the trust will be shown on the public Register of Overseas Entities. However, it may be shared with HMRC.</p>
+          '
+        }) }}
+        {{ govukButton({
+          text: "Add another trust",
```
markpit, line 88: similar to the comment for the submit button, could you add an 'id' attribute of "add" (see https://github.com/companieshouse/overseas-entities-web/blob/main/views/beneficial-owner-type.html#L86)
to the "Add another trust' button

src/utils/navigation.ts
mwestacott: No response to this comment so not sure if you agree and a change was made or was missed (generally useful to respond to all comments so it's clear that they've been read). Thanks.

markpit: I think the 
nextPage: config.CHECK_YOUR_ANSWERS_URL
will have to change to be an array as it can go to 2 next pages depending on trusts or not
i would change it to 
`nextPage: [
   config.CHECK_YOUR_ANSWERS_URL,
   config.TRUST_INFO_PAGE ]`

I don't think we currently use the next page value but plan to in future. Perhaps @mouhajer-ch can confirm what I've written is correct

markpit: I think the check_your_answers_page previousPage would have to be an array as well as could have come from either trusts page and beneficial_owner_type page

mwestacott: 👍 not ideal that we have code in there for the future, that we're not using yet. Should be added at the point it's used.

src/controllers/check.your.answers.controller.ts
```diff
@@ -18,9 +20,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
 
     const appData: ApplicationData = getApplicationData(req.session);
 
+    let hasTrusts: boolean = false;
+
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUST_INFO_16062022)) {
+      hasTrusts = checkEntityHasTrusts(appData);
+    }
+
```
markpit, line 28: I would also log out whether any trusts are present or not (line 28) 
something like logger.infoRequest(req, `config.CHECK_YOUR_ANSWERS_PAGE hasTrusts=${hasTrusts}`);

src/utils/navigation.ts
rsachdevch: That doesn't seem to work, @markpit because the Navigation interface type expects a `string` not an `array`

mwestacott: Sounds like this is fine for now as nextPage not currently used but when it is, this may not work. Would suggest adding a comment where the array list would normally be defined, just to point out that what's currently configured is not entirely correct.

mwestacott (approved):

markpit (approved):

PR #212, opened 2022-06-22
ROE-186 Clear beneficial owner individual service address data

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-186

### Change description

Clear service address data from the session on POST if yes is selected for same address so that we do not see the data on the page if we revisit.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/test/controllers/beneficial.owner.individual.controller.spec.ts
+++ b/test/controllers/beneficial.owner.individual.controller.spec.ts
@@ -16,6 +16,7 @@ import {
 } from "../../src/config";
 import {
   getFromApplicationData,
+  mapFieldsToDataObject,
   prepareData,
   removeFromApplicationData,
   setApplicationData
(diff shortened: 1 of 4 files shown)


test/controllers/beneficial.owner.individual.controller.spec.ts
```diff
@@ -98,6 +109,15 @@ describe("BENEFICIAL OWNER INDIVIDUAL controller", () => {
       expect(resp.header.location).toEqual(BENEFICIAL_OWNER_TYPE_URL);
     });
 
+    test(`redirect to the ${BENEFICIAL_OWNER_TYPE_URL} page after a successful post from ${BENEFICIAL_OWNER_INDIVIDUAL_PAGE} page with service address data`, async () => {
```
mwestacott, line 112: I'm not sure this is testing what the story is saying - it's talking about choosing the 'yes' then 'no' address radio buttons and ensuring that any address data entered is removed from the session.

I'm therefore wondering if the implementation is delivering what the story describes?

markpit: PR Updated - Matt and I have been looking at the test and have now added checks to check that mapFieldsToDataObject function is called with service address or not, and made the mapFieldsToDataObject mock return some dummy data so we can test if the service address gets replaced with dummy data during the mapping or if it is replaced with empty object (when they have selected service address is same as residential)

mwestacott (approved):

SVarallo (approved):

PR #223, opened 2022-06-29
ROE-808 Validate start date on corporate Mo page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-808

### Change description

Validate the start date values on the corporate MO page to ensure that the days and months are within the correct range e.g. 31/10/2020 but not 32/20/2020. 

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -41,10 +41,13 @@ export const checkDateIsInPast = (errMsg: string, day: string = "", month: strin
   return true;
 };
 
-export const checkDateValueIsValid = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
-  const inputDate = DateTime.utc(Number(year), Number(month), Number(day));
-  if (!inputDate.isValid) {
+export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+  const day = parseInt(dayStr), month = parseInt(monthStr), year = parseInt(yearStr);
+
+  if (isNaN(day) || isNaN(month) || isNaN(year) || !DateTime.utc(year, month, day).isValid) {
     throw new Error(errMsg);
   }
+
   return true;
 };
+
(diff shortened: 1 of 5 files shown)


src/validation/fields/date.validation.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import { checkDate, checkDateIsInPast } from "../custom.validation";
+import { checkDate, checkDateIsInPast, checkDateValueIsValid } from "../custom.validation";
 import { ErrorMessages } from "../error.messages";
 
 export const start_date_validations = [
   body("start_date")
     .custom((value, { req }) => checkDate(ErrorMessages.MANAGING_OFFICER_START_DATE, req.body["start_date-day"], req.body["start_date-month"], req.body["start_date-year"]))
-    .custom((value, { req }) => checkDateIsInPast(ErrorMessages.MO_START_DATE_NOT_IN_PAST, req.body["start_date-day"], req.body["start_date-month"], req.body["start_date-year"])),
+    .custom((value, { req }) => checkDateIsInPast(ErrorMessages.MO_START_DATE_NOT_IN_PAST, req.body["start_date-day"], req.body["start_date-month"], req.body["start_date-year"]))
```
mouhajer-ch, line 9: Check is valid should be added before checking id it is in the pass

src/validation/custom.validation.ts
```diff
@@ -40,3 +40,11 @@ export const checkDateIsInPast = (errMsg: string, day: string = "", month: strin
   }
   return true;
 };
+
+export const checkDateValueIsValid = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
```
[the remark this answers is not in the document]
mouhajer-ch: Thanks for the update and unit test

src/validation/fields/date.validation.ts
mouhajer-ch: The order will not block the checks for going forward so we will end up with different message errors. We should use the [oneOf](https://express-validator.github.io/docs/check-api.html#oneofvalidationchains-message) that comes with [express-validator](https://express-validator.github.io/docs/), but it is part of another story.

We will need to fix the `checkDateIsInPast` method and validate the values passed, I will do it in another PR. Thanks Matthew

PR #245, opened 2022-07-07
ROE-712 Validation for the BO Gov screen

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-712

### Change description

* Implementation as per the story, although some messages are
different to be more generic (as agreed)
* Unit test changes to check the validation and deal with the
fact that some fields are now mandatory

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/error.messages.ts
+++ b/src/validation/error.messages.ts
@@ -17,6 +17,7 @@ export enum ErrorMessages {
   AGENT_CODE = "Enter the agent assurance code",
   PARTNER_NAME = "Enter the name of the person with overall responsibility for identity checks",
   SUPERVISORY_NAME = "Enter the name of the supervisory body",
+  BO_GOV_NAME = "Enter the other legal entity’s name",
   // Public Register
   PUBLIC_REGISTER_NAME = "Enter the name of the register",
   PUBLIC_REGISTER_NUMBER = "Enter the registration number",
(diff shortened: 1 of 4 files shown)


src/validation/error.messages.ts
```diff
@@ -45,7 +46,7 @@ export enum ErrorMessages {
   SELECT_IF_INDIVIDUAL_PERSON_HAS_FORMER_NAME = "Select yes if the individual person has any former names",
   SELECT_IF_REGISTER_IN_COUNTRY_FORMED_IN = "Select yes if the overseas entity is already on a public register in the country it was formed in",
   SELECT_IF_MANAGING_OFFICER_REGISTER_IN_COUNTRY_FORMED_IN = "Select yes if the corporate managing officer is already on a public register in the country it was formed",
-  SELECT_IF_SERVICE_ADDRESS_SAME_AS_PRINCIPAL_ADDRESS = "Select yes if the correspondence address the same as the principal or registered office address",
+  SELECT_IF_SERVICE_ADDRESS_SAME_AS_PRINCIPAL_ADDRESS = "Select yes if the correspondence address is the same as the principal or registered office address",
```
mwestacott, line 49: English in error message clearly wrong here. I've corrected it and will highlight this in the JIRA ticket when passing over to test.

test/controllers/beneficial.owner.gov.controller.spec.ts
```diff
@@ -104,27 +104,31 @@ describe("BENEFICIAL OWNER GOV controller", () => {
 
     test(`redirects to the ${config.BENEFICIAL_OWNER_TYPE_URL} page`, async () => {
       mockPrepareData.mockReturnValueOnce(BENEFICIAL_OWNER_GOV_OBJECT_MOCK);
```
mwestacott, line 106: @markpit you mentioned doing this causes an issue but seems to be ok in this test. We should investigate why.

markpit: It's because we are using the same mock data to send into the request as we are to return from mock functions. When we return the from the mock, some functions will then modify the data. In my case it was because the test was to check if service address got removed. The mock data had the radio button set appropriately but the after the test completes, the data had been changed during the test, meaning that when that same mock was used again, it failed validation.

markpit: in your case they are different mocks, but in the one I was fixing it was the same one

markpit (approved):

PR #259, opened 2022-07-11
ROE-848 Make confirmation screen txt change for agents and non agents…

markpit:
… who are filing

### JIRA link
https://companieshouse.atlassian.net/browse/ROE-848

### Change description
Make confirmation screen text change depending on who is filing, agent or overseas entity

### Work checklist

### Merge instructions

```
```


src/controllers/confirmation.controller.ts
```diff
@@ -16,6 +16,7 @@ export const get = (req: Request, res: Response) => {
   deleteApplicationData(req.session);
 
   return res.render(CONFIRMATION_PAGE, {
+    ...appData,
```
mouhajer-ch, line 19: I would have just created another variable for `entity.email` instead of passing the entire object.

views/confirmation.html
```diff
@@ -20,15 +20,27 @@ <h1 class="govuk-panel__title">Application submitted</h1>
...
+      <h3 class="govuk-heading-s">What happens if the application is accepted</h3>
+      <p class="govuk-body">We'll email the Overseas Entity ID to <strong>{{ userEmail }}</strong>. We'll send the notice of registration to <strong>{{ entity.email }}</strong>, which will also include the Overseas Entity ID. This will need to be provided to the Land Registry whenever the overseas entity buys, sells or transfers land or property in the UK. </p>
 
-      <p class="govuk-body">If the application is accepted, the overseas entity and its beneficial owners will be added to the <a href="#">Register of Overseas Entities.</a></p>
+      <p class="govuk-body">The overseas entity and its beneficial owners will be added to the Register of Overseas Entities.</p>
 
-      <p class="govuk-body">If it's rejected, we'll explain what to do next in the email.</p>
+      <h3 class="govuk-heading-s">What happens if the application is rejected</h3>
+      <p class="govuk-body">If it's rejected, we'll email you at <strong>{{ userEmail }}</strong> to explain what to do next. We'll also refund the £X registration fee.</p>
```
mouhajer-ch, line 41: What do you think to create an env variable for the registration fee, similar to the API?

markpit: the fee is on the start page as well so if we do the env var, teh story will have to change to test the start page too.

PR #275, opened 2022-07-18
ROE-1003 Allow today for due diligence check date

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1003

### Change description

* Code now allows today's date for date when checks were completed
* Tests added to check new validation

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -41,6 +41,16 @@ export const checkDateIsInPast = (errMsg: string, day: string = "", month: strin
   return true;
 };
 
+export const checkDateIsInPastOrToday = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
+  const inputDate = DateTime.utc(Number(year), Number(month), Number(day));
+  const now = DateTime.now();
+  const currentDate = DateTime.utc(now.year, now.month, now.day); // exclude time of day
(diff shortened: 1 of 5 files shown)


src/validation/custom.validation.ts
```diff
@@ -41,6 +41,16 @@ export const checkDateIsInPast = (errMsg: string, day: string = "", month: strin
   return true;
 };
 
+export const checkDateIsInPastOrToday = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
```
mouhajer-ch, line 44: Can we have `checkDateIsInPastOrToday` instead of `checkDateIsInPast`? and change the message to include `today`?

mwestacott: Not at the moment. We have other screens that don't allow today, although I have raised this as something we might want to change. So we still need both at the moment.

mouhajer-ch: You are right but it is likely that we will have both error messages in the Error list, one that says `The date must be in the past` and `The date must be today or in the past`. 

It is an edge case and for MVP I will approve. Will need an update, can be done later.

mouhajer-ch (approved):

PR #276, opened 2022-07-18
ROE-888 Fix to remove service address country and register info from …

markpit:
…entity screen when not selected

### JIRA link
https://companieshouse.atlassian.net/browse/ROE-888

### Change description
Fix to remove 'register in country formed in' data when register option is selected as 'no', any data previously entered in that section gets removed on submit. Also fix service address country, it was using the same value as entity country if loading the screen with entity data and no service address.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/entity.controller.ts
+++ b/src/controllers/entity.controller.ts
@@ -9,7 +9,7 @@ import {
 } from "../utils/application.data";
 import { EntityKey, EntityKeys } from "../model/entity.model";
 import { ApplicationData, ApplicationDataType } from "../model";
-import { AddressKeys, HasSamePrincipalAddressKey, IsOnRegisterInCountryFormedInKey } from "../model/data.types.model";
+import { AddressKeys, HasSamePrincipalAddressKey, IsOnRegisterInCountryFormedInKey, PublicRegisterNameKey, RegistrationNumberKey } from "../model/data.types.model";
 import { logger } from "../utils/logger";
 import * as config from "../config";
 import { PrincipalAddressKey, PrincipalAddressKeys, ServiceAddressKey, ServiceAddressKeys } from "../model/address.model";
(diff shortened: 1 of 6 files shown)


views/includes/inputs/address/service-address-input.html
```diff
@@ -63,7 +63,7 @@
     value: service_address_county
   }) }}
 
-  {% set countryField = service_address_country %}
+  {% set countryField = service_address_country if service_address_country %}
```
mouhajer-ch, line 66: We do not need this condition

src/controllers/entity.controller.ts
```diff
@@ -52,6 +52,10 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
       :  {};
     data[IsOnRegisterInCountryFormedInKey] = (data[IsOnRegisterInCountryFormedInKey]) ? +data[IsOnRegisterInCountryFormedInKey] : '';
 
+    // Wipe 'register in country formed in' data if IsOnRegisterInCountryFormedInKey is no or not selected
+    data[PublicRegisterNameKey] = (data[IsOnRegisterInCountryFormedInKey]) ? data[PublicRegisterNameKey] : '';
```
mouhajer-ch, line 56: Why checking the field `data[IsOnRegisterInCountryFormedInKey] `twice?

views/includes/inputs/address/service-address-input.html
markpit: it's weird, it seems to fix the problem, i'll see if there's another way

src/controllers/entity.controller.ts
markpit: i can tidy that up, was just following pattern of that function

mouhajer-ch (approved):

mwestacott (approved):

PR #286, opened 2022-07-19
Add accessibility statement page

mouhajer-ch:
### JIRA link

Resolves [ROE-294](https://companieshouse.atlassian.net/browse/ROE-294)

### Change description

- Add accessibility statement view, controller, route and tests

### Work checklist


views/accessibility-statement.html
```diff
@@ -0,0 +1,112 @@
...
+                <li><a href="#reporting">Reporting accessibility problems with this service</a></li>
+                <li><a href="#enforcement">Enforcement procedure</a></li>
+                <li><a href="#technical-information">Technical information about this service’s accessibility</a></li>
+                <li><a href="#tested">How we tested this service</a></li>
+                <li><a href="#were-doing">What we’re doing to improve accessibility</a></li>
+            </ul>
+
+            <h2 class="govuk-heading-m" id="about-service">About this service</h2>
+
+            <p>This accessibility statement applies to the 'Register an overseas entity and tell us about its beneficial owners' service.</p>
```
markpit, line 28: I think the <p> tags will need to have the govuk-body styling to keep the font the same as the other pages
<p class="govuk-body">

views/accessibility-statement.html
```diff
@@ -0,0 +1,112 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Accessibility statement – {{ SERVICE_NAME }} - GOV.UK
```
SVarallo, line 4: Small thing but are the hyphens the correct ones? I remember we had to use '-' to avoid issues

views/accessibility-statement.html
SVarallo: Yes this will need to be done for all the <p> tags by the looks of it

views/accessibility-statement.html
mouhajer-ch: It is not an issue from a technical point of view, the browser will render correctly the hyphen (-) or dashes (— or –). For consistency I will update it to match other pages

views/accessibility-statement.html
mouhajer-ch: Yep, will do

markpit (approved):

PR #292, opened 2022-07-20
Trust Validation - mandatory fields

emead:
### JIRA link
[ROE-1031](https://companieshouse.atlassian.net/browse/ROE-1031)

### Change description
Add validation for mandatory fields within the Trust Data JSON.
This validation has to happen after the JSON is parsed, so happens in the controller.

### Work checklist


src/validation/trust.information.validation.ts
```diff
@@ -1,7 +1,43 @@
...
 ];
+
+export type TrustValidationError = {
+  param: any;
+  msg: any;
+};
+
+export const checkMandatoryTrustFields = (trusts: trustType.Trust[]): TrustValidationError[] => {
+  const errors: TrustValidationError[] = [];
+  trusts.forEach(trust => {
```
markpit, line 18: forEach won't wait for each item to be processed so may return errors as empty before each trust => {.. } function has a chance to run. 
Mo has recommended to me before to use the .map function when iterating over an array, or failing that use the for(item of items)

src/controllers/trust.information.controller.ts
```diff
@@ -38,6 +39,19 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
       trusts: trustData
     };
 
+    const appData: ApplicationData = getApplicationData(req.session);
+
+    const trustErrors = checkMandatoryTrustFields(trustData);
+    if ( trustErrors.length > 0 ) {
+      return res.render(config.TRUST_INFO_PAGE, {
```
markpit, line 46: it looks like appData is only needed by the res.render so I would move
inside the 
if ( trustErrors.length > 0 ) {

so that we only make the call if it is needed

src/validation/trust.information.validation.ts
mwestacott: So this comment applies to change on line 127 of `trust.information.controller.ts` too @markpit ?

test/__mocks__/session.mock.ts
```diff
@@ -711,8 +743,26 @@ const hasSoldLandKey = '0';
...
+};
+
+export const TRUST_NO_NAME: Trust = {
   trust_id: "",
   trust_name: "",
+  creation_date_day: "31",
+  creation_date_month: "12",
+  creation_date_year: "1999",
+  unable_to_obtain_all_trust_info: "No"
+};
```
mwestacott, line 761: Perhaps worth adding a mock and test when just one of the date field component fields is empty/undefined.

src/validation/trust.information.validation.ts
markpit: yes it'll apply to 127 as well

src/validation/custom.validation.ts
```diff
@@ -89,3 +90,24 @@ export const checkAtLeastOneFieldHasValue = (errMsg: string, ...fields: any[]) =
...
+  trusts.forEach(trust => {
+    if (
+      trust.creation_date_day === undefined ||
+      trust.creation_date_day === "" ||
+      trust.creation_date_month === undefined ||
+      trust.creation_date_month === "" ||
+      trust.creation_date_year === undefined ||
+      trust.creation_date_year === ""
+    ) {
+      console.log(dateErrMsg);
```
markpit, line 105: remove the console.log or use logger.error / logger.info to log

src/validation/custom.validation.ts
```diff
@@ -89,3 +90,23 @@ export const checkAtLeastOneFieldHasValue = (errMsg: string, ...fields: any[]) =
   }
   throw new Error(errMsg);
 };
+
+export const checkMandatoryTrustFields = (nameErrMsg, dateErrMsg, trusts_json: string) => {
+  const trusts: trustType.Trust[] = JSON.parse(trusts_json);
+  trusts.forEach(trust => {
```
markpit, line 96: i'd use for(trust of trusts) {   here instead of forEach

views/trust-information.html
```diff
@@ -19,42 +19,24 @@ <h1 class="govuk-heading-xl">Add trust information for the beneficial owners</h1
...
-                      <label class="govuk-label govuk-checkboxes__label" for="{{ boc.id }}">
-                        {{ boc.name }}
-                      </label>
-                    </div>
-                  {% endif %}
-                {% endfor %}
-              {% endif %}
-            </div>
-          </fieldset>
+            {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
```
markpit, line 22: need to remove the import, it should already be in the layout.html which this file extends

src/validation/trust.information.validation.ts
```diff
@@ -1,7 +1,12 @@
...
+import { checkAtLeastOneFieldHasValue, checkMandatoryTrustFields } from "./custom.validation";
 
 import { ErrorMessages } from "./error.messages";
 
 export const trustInformation = [
-  body("trusts").not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.TRUST_DATA_EMPTY)
+  body("trusts").not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.TRUST_DATA_EMPTY),
+  body("beneficialOwners").custom((value, { req }) =>
+    checkAtLeastOneFieldHasValue(ErrorMessages.TRUST_BO_CHECKBOX, req.body.beneficialOwners)),
+  body("trusts").custom((value, { req }) =>
```
markpit, line 10: you should be able to just chain the 'trust' checks,
body("trusts")
  .not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.TRUST_DATA_EMPTY)
  .custom((value, { req }) => .....

src/validation/custom.validation.ts
```diff
@@ -89,3 +90,23 @@ export const checkAtLeastOneFieldHasValue = (errMsg: string, ...fields: any[]) =
   }
   throw new Error(errMsg);
 };
+
+export const checkMandatoryTrustFields = (nameErrMsg, dateErrMsg, trusts_json: string) => {
```
markpit, line 94: i'd stick to camel case for var names - trusts_json -> trustsJson

src/validation/custom.validation.ts
```diff
@@ -89,3 +90,23 @@ export const checkAtLeastOneFieldHasValue = (errMsg: string, ...fields: any[]) =
   }
   throw new Error(errMsg);
 };
+
+export const checkMandatoryTrustFields = (nameErrMsg, dateErrMsg, trustsJson: string) => {
+  const trusts: trustType.Trust[] = JSON.parse(trustsJson);
```
markpit, line 95: just thinking, if you enter un-parseable json, I imagine JSON.parse will throw an error here, and error thrown from this function will get displayed on the screen. So I think you'd have to wrap the JSON.parse line in it's own try-catch and log the parse error using logger.error and throw a 'friendly' error message - 'invalid json' or something.

emead: ROE-789 will deal with un-parseable json

src/middleware/validation.middleware.ts
```diff
@@ -56,6 +57,29 @@ export function checkValidations(req: Request, res: Response, next: NextFunction
...
+    if (!errorList.isEmpty()) {
+      const errors = formatValidationError(errorList.array());
+      const routePath = req.route.path;
+      const appData: ApplicationData = getApplicationData(req.session);
+
+      return res.render(NAVIGATION[routePath].currentPage, {
+        backLinkUrl: NAVIGATION[routePath].previousPage(appData),
+        ...req.body,
+        beneficialOwners: getBeneficialOwnerList(appData),
+        trusts_input: req.body.trusts.toString(),
```
mouhajer-ch, line 72: It should be `trusts_input: req.body.trusts?.toString()`

src/validation/custom.validation.ts
markpit: ok, just be aware that if that work had already been tested, will need to be retested when we add this check in

test/__mocks__/session.mock.ts
```diff
@@ -751,3 +806,33 @@ export const APPLICATION_DATA_NO_TRUSTS_MOCK: ApplicationData = {
   [OverseasEntityKey]: OVERSEAS_ENTITY_ID,
   [Transactionkey]: TRANSACTION_ID
 };
+
+export const APPLICATION_DATA_NO_TRUST_NAME_MOCK: ApplicationData = {
```
mouhajer-ch, line 810: These can be written as:

```
export const APPLICATION_DATA_NO_TRUST_NAME_MOCK: ApplicationData = {
  ...APPLICATION_DATA_NO_TRUSTS_MOCK,
  [TrustKey]: [TRUST_NO_NAME]
};

export const APPLICATION_DATA_NO_TRUST_DATE_MOCK: ApplicationData = {
  ...APPLICATION_DATA_NO_TRUSTS_MOCK,
  [TrustKey]: [TRUST_NO_DATE]
};
```

views/trust-information.html
```diff
@@ -67,7 +47,8 @@ <h2 class="govuk-heading-m">Enter trust information from the Excel document</h2>
               text: "If you've already completed the trust Excel document, follow the instructions in the document to paste the data into the box:",
               classes: "govuk-body",
               isPageHeading: false
-            }
+            },
+            value: trusts_input
```
mouhajer-ch, line 51: It seems the `trusts_input` is not passed in the trust page in the get controller.

emead: This isn't needed for the `get` as there isn't data to be displayed.

mouhajer-ch: I think the user wants to visualise what has been added after saving it

emead: I've tested this locally and that's not the case. The agreed user journey is that no saved data will be displayed until the Check Your Answers page.
This trusts_input is to pre-populate the field with the submitted data, only when validation has failed.

markpit: just to confirm, the trust input box should be empty if the user navigates back to the page after entering valid json?  but we would need to show the json if there is an error.

emead: Yes, that is correct

mouhajer-ch: Something like `trusts_input: appData[TrustKey]?.toString() ` probably will do, but for MVP can be done in another task

mouhajer-ch (approved):

markpit (approved):

PR #341, opened 2022-08-09
ROE-1089 Add audible hidden text to beneficial owner types page

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1089

### Change description
Added hidden text to the:
add button
no more button
subheading

Added three new includes templates and a new variable for each of the three possible identification scenarios.

### Work checklist

### Merge instructions

```
```


views/beneficial-owner-type.html
```diff
@@ -87,37 +90,24 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ boTypesTitle }}</h1>
           items: boTypes
         }) }}
 
-        {{ govukButton({
-          text: "Add",
-          attributes: {
-            "id": "add"
-          }
-        }) }}
+        <button class="govuk-button" data-module="govuk-button" id="add">Add <span class="govuk-visually-hidden"> {{ whatToAdd }} </span></button>
```
markpit, line 93: the examples James Francis gives look like they are taken from the page source rather than the nunjucks components. Does it work the same if you add a govuk-visually-hidden attribute to the govukButton component?

views/beneficial-owner-type.html
```diff
@@ -87,37 +90,24 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ boTypesTitle }}</h1>
...
-
-          {{ govukButton({
-            text: "No more to add",
-            attributes: {
-              "id": "submit"
-            },
-            classes: "govuk-button--secondary govuk-!-margin-top-6",
-            href: "/register-an-overseas-entity/" + nextPage
-          }) }}
+          %}
```
markpit, line 98: it looks like you just want to determine if the button text should include one of these 3
beneficial owners
managing officers
beneficial owners or managing officers

We should be able to do it without the includes - if you look at the original version before this PR,
we can create 2 new variables in the template
hasBeneficialOwners
hasManagingOfficers

(this is psuedo code)
set hasBeneficialOwners = (beneficial_owners_individual and beneficial_owners_individual.length > 0)
          or (beneficial_owners_corporate and beneficial_owners_corporate.length > 0)
          or (beneficial_owners_government_or_public_authority and beneficial_owners_government_or_public_authority.length > 0)

hasManagingOfficers = (managing_officers_individual and managing_officers_individual.length > 0) 
         or (managing_officers_corporate and managing_officers_corporate.length > 0)

then you can use those 2 variables to replace the previous "5 line" if statement that was previously on line 97 to be
if (hasBeneficialOwners or hasManagingOfficers)

then you can use those 2 'has' variables to set another variable that will be used to set the hidden text on the button, one of these 3
beneficial owners
managing officers
beneficial owners or managing officers

you could call that variable hiddenTextValue
eg 
if (hasBeneficialOwners) then 
    set hiddenTextValue = "beneficial owners"
if (hasManagingOfficers) then 
    set hiddenTextValue = "managing officers"
if (hasBeneficialOwners AND hasManagingOfficers)
    set hiddenTextValue = "beneficial owners or managing officers"

then use that 'hiddenTextValue' var in the buttons
<span class="govuk-visually-hidden"> {{hiddenTextValue}} </span>

views/beneficial-owner-type.html
mattch1: The problem is that the govuk-visually-hidden attribute is attached to a nested element not directly to the button.

markpit: ok, just wondered if there was an easier way to do it with nunjucks. Stick with the html buttons.

views/beneficial-owner-type.html
mattch1: Set the and/or values based on what is on the page.
Check to find the correct hidden value text - most specific bo and mo first.
Then if there is anything on the page display the subheading with the bo types below it.

markpit (approved):

PR #352, opened 2022-08-16
ROE-1123 adding matomo tracking to change links

SVarallo:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1123

### Change description
Added an option to the CREATE_CHANGE_LINK function to allow a data-event-id attribute to be added for matomo tracking. Then added data for this new parameter to each CREATE_CHANGE_LINK call.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/change.link.ts
+++ b/src/utils/change.link.ts
@@ -1,10 +1,13 @@
 import { FEATURE_FLAG_ENABLE_CHANGE_LINKS_02082022 } from "../config";
 import { isActiveFeature } from "./feature.flag";
 
-export const createChangeLinkConfig = (href: string, text: string) => {
+export const createChangeLinkConfig = (href: string, text: string, dataEventId: string) => {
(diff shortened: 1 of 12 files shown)


mouhajer-ch (commented):
We should update the unit tests [here](https://github.com/companieshouse/overseas-entities-web/blob/main/test/utils/change.link.spec.ts). Fix the params on the `createChangeLinkConfig` and add a new expect related to the `attributes`, something like 
```
  test('should check if the object returned is correct, and contains the change link configs', () => {
    const testChangeLinkConfig = createChangeLinkConfig(PRESENTER_CHANGE_FULL_NAME, CHANGE_LINK_NAME_PRESENTER, DATA_EVENT_ID) as any;
    ...
    expect(testChangeLinkConfig.attributes).toEqual(DATA_EVENT_ID);
    ...
  });
```

mouhajer-ch (approved):

PR #411, opened 2022-10-12
Feature/roe 1217 add are you sure you want to sign out page

mouhajer-ch:
### JIRA link

Resolves [ROE-1217](https://companieshouse.atlassian.net/browse/ROE-1217) and [ROE-1213](https://companieshouse.atlassian.net/browse/ROE-1213)

### Change description

- Create 'sign out' function
- Create 'are you sure you want to sign out?' page

### Work checklist


src/config/index.ts
```diff
@@ -59,6 +60,7 @@ export const SOLD_LAND_FILTER_PAGE = "sold-land-filter";
 export const TRUST_INFO_PAGE = "trust-information";
 export const USE_PAPER_PAGE = "use-paper";
 export const WHO_IS_MAKING_FILING_PAGE = "who-is-making-filing";
+export const SIGN_OUT_PAGE = "sign-out";
```
mwestacott, line 63: I think this represents the page where the user confirms they wish to sign out? If yes, suggest a minor name change here (and possibly to other code artefacts below) to make this clear. Could name the constant CONFIRM_SIGN_OUT_PAGE.

src/controllers/beneficial.owner.gov.controller.ts
```diff
@@ -40,7 +40,7 @@ export const getById = (req: Request, res: Response, next: NextFunction) => {
 
     return res.render(BENEFICIAL_OWNER_GOV_PAGE, {
       backLinkUrl: BENEFICIAL_OWNER_TYPE_URL,
-      templateName: BENEFICIAL_OWNER_GOV_PAGE,
+      templateName: `${BENEFICIAL_OWNER_GOV_PAGE}/${id}`,
```
mwestacott, line 43: Just wondering why this file and the one below have changed here? Surely nothing to do with sign out?

src/middleware/authentication.middleware.ts
```diff
@@ -14,8 +14,9 @@ export const authentication = (req: Request, res: Response, next: NextFunction):
       logger.infoRequest(req, 'User not authenticated, redirecting to sign in page, status_code=302');
       return res.redirect(`/signin?return_to=${SOLD_LAND_FILTER_URL}`);
     }
-
-    logger.infoRequest(req, `User (${getLoggedInUserEmail(req.session)}) is signed in`);
+    const userEmail = getLoggedInUserEmail(req.session);
+    logger.infoRequest(req, `User (${ userEmail }) is signed in`);
+    res.locals.userEmail = userEmail;
```
mwestacott, line 19: Assume this is somehow needed for sign out to work? Worth a code comment explaining why maybe as it's not immediately obvious why this is done.

views/sign-out.html
```diff
@@ -0,0 +1,57 @@
+{% extends "layout.html" %}
```
mwestacott, line 1: Just wondering, does the sign-out link appear on the 'confirm sign out' page too? Should it?

views/sign-out.html
```diff
@@ -0,0 +1,57 @@
...
+        name: "sign_out",
+        fieldset: {
+          legend: {
+            text: "Are you sure you want to sign out?",
+            isPageHeading: true,
+            classes: "govuk-fieldset__legend--l"
+          }
+        },
+        hint: {
+            text: "Your answers will not be saved. You will need to start again if you want to register an overseas entity’s beneficial owner."
```
mwestacott, line 31: Text here does not match that shown in the prototype image in the story. Guessing this is just the template from the prototype but need to check with analyst which is correct.

src/config/index.ts
mouhajer-ch: Same name in the url and file name. The same as for the other template name, so I will keep it consistent.

src/controllers/beneficial.owner.gov.controller.ts
mouhajer-ch: `templateName` will be part of the redirect url and needs to be complete with IDs in the getById when MOs or BOs are selected for change/delete. We can have a call

src/middleware/authentication.middleware.ts
mouhajer-ch: You are right, sorry, I will add a comment. I am using thehttps://expressjs.com/en/5x/api.html#res.locals to make sure that the email is available within a single request-response cycle and visible in the template. This way I will avoid changing different part of the code and is more maintainable and clean

views/sign-out.html
mouhajer-ch: It does not appear on the new sign out page.

views/sign-out.html
mouhajer-ch: During the refinement has been decided to change similar to confirmation statement. I have added a note on the story

src/controllers/beneficial.owner.gov.controller.ts
mwestacott: But is this change needed (now) for sign out? I can't see how it relates to a requirement in the stories you're working on, but yes, let's have a call if required.

src/middleware/authentication.middleware.ts
mwestacott: 👍 nice. Yes, a comment would help here.

views/sign-out.html
mwestacott: Ok. If agreed during the refinement then we need to ask analyst to put a note on the story 'Description' to say how the text varies. As it stands, delivering this will (should) result in a defect.

mwestacott: The text change is odd though - you'd need to start again if you wanted to register an OE's beneficial owner OR, indeed, an Overseas Entity...

mouhajer-ch: Thanks Martin, I will rise it during the standup

src/controllers/beneficial.owner.gov.controller.ts
markpit: templateName is used by the matomo tracking so we don't want to alter the value passed into the template. Perhaps pass id as a separate parameter?

mwestacott: Would still like to know which 'sign out' requirement this change relates to.

mouhajer-ch: Thanks @markpit, this way will correctly track the real page that the user is selecting, probably I need to discuss it with Sam.

I could concatenate the ID to the template when present as a second implementation option instead of changing the getByID method.

views/sign-out.html
mouhajer-ch: Karen has suggested a new text. It will be updated to `Your answers will not be saved. You will need to start again if you want to register an overseas entity and tell us about its beneficial owners.`

src/controllers/beneficial.owner.gov.controller.ts
SVarallo: Had a talk with Mou and looked at the matomo dashboard and it seems to be able to differentiate between the ids while also putting them all together under managing officer in the tracking. Can't confirm if it's 100% necessary for the 'sign out' functionality but it certainly seems like a useful addition 👍

views/sign-out.html
mwestacott: Nice one.

src/controllers/beneficial.owner.gov.controller.ts
mwestacott: Sounds useful, but if not sign out functionality can we move this change to another story, so that it can be tested against requirements?

mouhajer-ch: Just for clarity I posted the snippet related to the sign out link, and you can see the the `href` for the `a` link is constructed using the `templateName`, so we need to have the full and correct path including the ID if you are signing out from MOs or BOs selected. The changes will improve as well the matzo tracking 
```
{% if userEmail %}
   <ul id="navigation">
     <li class="content-email">{{ userEmail }}</li>
     <li>
       <a href="{{ OE_CONFIGS.REGISTER_AN_OVERSEAS_ENTITY_URL }}sign-out?page={{ templateName }}" data-event-id="signout">Sign out</a>
     </li>
   </ul>
   <hr class="govuk-section-break govuk-section-break--visible"> 
{% endif %}
```

mwestacott: Right, I'm getting there slowly. Maybe. @mouhajer-ch  Is this needed now (for sign out) because if the user navigates to the 'confirm sign out' page from a BO or MO page, the navigation back to that page - if they abort sign out - won't work without this addition of the Id?

src/config/index.ts
mouhajer-ch: Thanks Martin

views/sign-out.html
mouhajer-ch: Updated

src/middleware/authentication.middleware.ts
mouhajer-ch: Done

mwestacott (approved):

PR #423, opened 2022-10-21
Add OE update call to the presenter and check your answers page

mouhajer-ch:
### JIRA link

Resolves [ROE-1441](https://companieshouse.atlassian.net/browse/ROE-1441)

### Change description

Add OE update call to the presenter and check your answers page

### Work checklist

### DIFF
--- a/src/controllers/check.your.answers.controller.ts
+++ b/src/controllers/check.your.answers.controller.ts
@@ -1,7 +1,7 @@
 import { Session } from "@companieshouse/node-session-handler";
 import { NextFunction, Request, Response } from "express";
 
-import { createOverseasEntity } from "../service/overseas.entities.service";
+import { createOverseasEntity, updateOverseasEntity } from "../service/overseas.entities.service";
 import { closeTransaction, postTransaction } from "../service/transaction.service";
 
 import * as config from "../config";
@@ -66,8 +66,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
       ? appData[OverseasEntityKey] as string
       : await createOverseasEntity(req, session, transactionID);
 
-    // TODO: Missing last put call to submit OE, it will be done on ROE-1441.
-    // Note: this will be removed in the future when all PUT calls have been set correctly on the others pages.
+    await updateOverseasEntity(req, session);
 
     const transactionClosedResponse = await closeTransaction(req, session, transactionID, overseasEntityID);
     logger.infoRequest(req, `Transaction Closed, ID: ${transactionID}`);
--- a/src/controllers/check.your.answers.controller.ts
+++ b/src/controllers/check.your.answers.controller.ts
@@ -57,16 +57,15 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
       logger.infoRequest(req, `New access token: ${accessToken}`);
     }
 
-    const isSaveAndResumeFeatureActive = isActiveFeature(config.FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022);
-    const transactionID = (isSaveAndResumeFeatureActive)
-      ? appData[Transactionkey] as string
-      : await postTransaction(req, session);
-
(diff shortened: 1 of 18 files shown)


test/service/overseas.entities.service.spec.ts
```diff
@@ -63,3 +69,33 @@ describe('Overseas Entity Service test suite', () => {
   });
 
 });
+
+describe('Update Overseas Entity Service test suite', () => {
```
mattch1, line 73: Aren't we using double quotes for these?

test/service/overseas.entities.service.spec.ts
```diff
@@ -1,30 +1,35 @@
 jest.mock("@companieshouse/api-sdk-node");
 jest.mock("@companieshouse/api-sdk-node/dist/services/overseas-entities");
 jest.mock('../../src/utils/logger');
+jest.mock('../../src/utils/application.data');
```
mattch1, line 4: See comment below

src/controllers/check.your.answers.controller.ts
```diff
@@ -66,8 +66,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
       ? appData[OverseasEntityKey] as string
       : await createOverseasEntity(req, session, transactionID);
 
-    // TODO: Missing last put call to submit OE, it will be done on ROE-1441.
-    // Note: this will be removed in the future when all PUT calls have been set correctly on the others pages.
+    await updateOverseasEntity(req, session);
```
mwestacott, line 69: I think we only need to do this PUT call if the 'save and resume' feature is active.

Looking at the above logic might read better if we just have one 'if' check on the feature flag and two execution blocks - one for when we're in save and resume mode and the other when not.

test/__mocks__/text.mock.ts
```diff
@@ -91,3 +91,4 @@ export const CHANGE_LINK_NAME_PRESENTER = "Full name - who can we contact about
 export const DATA_EVENT_ID = "data-event-id";
 export const PAGE_TITLE_ERROR = "Error:";
 export const PAGE_NOT_FOUND_TEXT = "Page not found";
+export const UPDATE_OE__MSG_ERROR = "Something went wrong with updating Overseas Entity";
```
mwestacott, line 94: Minor, but two underscores __ present here...

src/controllers/check.your.answers.controller.ts
mouhajer-ch: Good spot, but I ill not use a single if condition. It is always better using the `cost` instead of `let`

test/service/overseas.entities.service.spec.ts
mouhajer-ch: Better use template literals here.

test/service/overseas.entities.service.spec.ts
mouhajer-ch: For consistency I will use the double quote, but it is the same here

src/controllers/check.your.answers.controller.ts
mwestacott: 👍 I suppose what I'm saying is that using the ternary operator here means multiple checks - an if ... else block would make flow easier to read I think, but will wait to see the next updates.

test/service/overseas.entities.service.spec.ts
mouhajer-ch: done

test/service/overseas.entities.service.spec.ts
mouhajer-ch: done

src/controllers/check.your.answers.controller.ts
mouhajer-ch: Done, it will be refactored anyway soon

test/__mocks__/text.mock.ts
mouhajer-ch: done

mwestacott (approved):

PR #458, opened 2022-11-08
ROE-1582: add TRUST_WEB feature flag; refactor FeatureFlag functional…

dgolubevs:
…ity;

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -9,14 +9,12 @@ export const COOKIE_SECRET = getEnvironmentValue("COOKIE_SECRET");
 export const COOKIE_DOMAIN = getEnvironmentValue("COOKIE_DOMAIN");
 export const DEFAULT_SESSION_EXPIRATION = getEnvironmentValue("DEFAULT_SESSION_EXPIRATION", "3600");
 export const CACHE_SERVER = getEnvironmentValue("CACHE_SERVER");
-export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentValue("SHOW_SERVICE_OFFLINE_PAGE");
 export const API_URL = getEnvironmentValue("API_URL");
 export const CHS_URL = getEnvironmentValue("CHS_URL");
 export const CHS_API_KEY = getEnvironmentValue("CHS_API_KEY");
 export const PIWIK_URL = getEnvironmentValue("PIWIK_URL");
 export const PIWIK_SITE_ID = getEnvironmentValue("PIWIK_SITE_ID");
 export const PIWIK_START_GOAL_ID = getEnvironmentValue("PIWIK_START_GOAL_ID");
-export const FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022 = getEnvironmentValue("FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022", "false");
 export const PAYMENT_FEE = getEnvironmentValue("PAYMENT_FEE");
 export const LANDING_PAGE_URL = getEnvironmentValue("LANDING_PAGE_URL");
 export const OAUTH2_CLIENT_ID = getEnvironmentValue(`OAUTH2_CLIENT_ID`);
@@ -29,6 +27,11 @@ export const DESCRIPTION = "Overseas Entities Transaction";
 export const REFERENCE = "OverseasEntitiesReference";
 export const PAYMENT_REQUIRED_HEADER = "x-payment-required";
 export const PAYMENT_PAID = "paid";
+
+//  FEATURE FLAGS
+export const SHOW_SERVICE_OFFLINE_PAGE = getEnvironmentValue("SHOW_SERVICE_OFFLINE_PAGE");
+export const FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022 = getEnvironmentValue("FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022", "false");
+export const FEATURE_FLAG_ENABLE_TRUSTS_WEB = <any>getEnvironmentValue('FEATURE_FLAG_ENABLE_TRUSTS_WEB_07112022', 'false');
 
 // TEMPLATEs
 export const ACCESSIBILITY_STATEMENT_PAGE = "accessibility-statement";
--- a/src/utils/feature.flag.ts
+++ b/src/utils/feature.flag.ts
@@ -1,15 +1,8 @@
 /**
  * Feature flags will be determined by environment variables and all environment variables in nodejs are
- * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
+ * either string or undefined.
+ * This function will ensure that 'true', '1', 'on' will be Truthy
  */
 export const isActiveFeature = (flag: string | undefined): boolean => {
-
-  if (flag === undefined) {
-    return false;
-  }
-  const featureFlag = flag.toLowerCase();
-  return !(featureFlag === "false" ||
-    featureFlag === "0" ||
-    featureFlag === "off" ||
-    featureFlag === "");
+  return ['1', 'true', 'on'].includes(flag?.toString().toLowerCase() ?? '');
 };
--- a/test/utils/feature.flag.spec.ts
+++ b/test/utils/feature.flag.spec.ts
@@ -1,60 +1,32 @@
-import { describe, expect, test } from '@jest/globals';
+import { describe, expect } from '@jest/globals';
 import { isActiveFeature } from "../../src/utils/feature.flag";
 
-describe("feature flag tests", function() {
+describe("feature flag tests", function () {
+  const dataExpectFalse = [
+    'false',
+    'faLse',
+    '0',
+    'off',
+    'oFF',
+    'kdjhskjf',
+  ];
 
-  test("should return false if variable is 'false'", function() {
-    const active = isActiveFeature("false");
+  test.each(dataExpectFalse)('should return false if variable is %p', (val) => {
+    const active = isActiveFeature(val as any);
     expect(active).toBeFalsy;
   });
 
-  test("should return false if variable is '0'", function() {
-    const active = isActiveFeature("0");
-    expect(active).toBeFalsy;
-  });
+  const dataExpectTrue = [
+    'true',
+    'trUe',
+    'TRUE',
+    'on',
+    'oN',
+    '1',
+  ];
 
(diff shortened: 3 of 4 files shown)


SteveHicksCH (approved):
Nice improvement for feature flags

src/config/index.ts
```diff
@@ -31,6 +29,15 @@ export const REFERENCE = "OverseasEntitiesReference";
 export const PAYMENT_REQUIRED_HEADER = "x-payment-required";
 export const PAYMENT_PAID = "paid";
 
+//  FEATURE FLAGS
+const envFlagSfx = 'FEATURE_FLAG_';
+enum FeatureFlag {
```
mouhajer-ch, line 34: This enum is not needed, feature flags are short lived, they will be removed probably in the next two sprints. Please keep things simple

src/utils/feature.flag.ts
```diff
@@ -2,14 +2,8 @@
...
 
-  if (flag === undefined) {
-    return false;
-  }
-  const featureFlag = flag.toLowerCase();
-  return !(featureFlag === "false" ||
-    featureFlag === "0" ||
-    featureFlag === "off" ||
-    featureFlag === "");
+export const isActiveFeature = (flag: FeatureFlag): boolean => {
```
mouhajer-ch, line 7: Nice one, but any refactoring, outside Trust work, can be please agreed with toro to avoid conflicts.

test/utils/feature.flag.spec.ts
```diff
@@ -1,60 +1,32 @@
...
+  const dataExpectFalse = [
+    'false',
+    'faLse',
+    '0',
+    'off',
+    'oFF',
+    'kdjhskjf',
+  ];
+
+  it.only.each(dataExpectFalse)('should return false if variable is %p', (val) => {
```
mouhajer-ch, line 14: Nice one, but could you please use `test` (not its alias) and add `test` on the `import`

mouhajer-ch (commented):
Could you please make sure that the `Change description` and `Jira link` are correctly reported on your PRs

src/utils/feature.flag.ts
```diff
@@ -2,14 +2,8 @@
  * Feature flags will be determined by environment variables and all environment variables in nodejs are
  * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
```
mwestacott, line 3: Worth updating this comment I think to match the new implementation - so saying something like "'1', 'true' and 'on' indicate that a feature is active, any other value means it is not".

src/config/index.ts
```diff
@@ -31,6 +29,15 @@ export const REFERENCE = "OverseasEntitiesReference";
 export const PAYMENT_REQUIRED_HEADER = "x-payment-required";
 export const PAYMENT_PAID = "paid";
 
+//  FEATURE FLAGS
+const envFlagSfx = 'FEATURE_FLAG_';
```
mwestacott, line 33: I think this represents a prefix not a suffix (if that's what 'Sfx' is meant to represent)?

Suggest changing name to `featureFlagNamePrefix` if this remains.

src/config/index.ts
```diff
@@ -31,6 +29,15 @@ export const REFERENCE = "OverseasEntitiesReference";
...
+enum FeatureFlag {
+  ENABLE_SHOW_SERVICE_OFFLINE_PAGE = <any>getEnvironmentValue("SHOW_SERVICE_OFFLINE_PAGE"),
+  ENABLE_REFRESH_TOKEN_29092022 = <any>getEnvironmentValue(`${envFlagSfx}ENABLE_REFRESH_TOKEN_29092022`, "false"),
```
mwestacott, line 36: If remaining (and whilst we're refactoring this), what do we think about removing the date at the end of the Node JS constant names? e.g. say:

`ENABLE_REFRESH_TOKEN = <any>getEnvironmentValue(`${envFlagSfx}ENABLE_REFRESH_TOKEN_29092022`, "false"),`

The date is needed at the end of the feature flag name in config files (according to our standards) but I don't see the need for its presence and the increased verbosity when used throughout the code.

src/utils/feature.flag.ts
markpit: yes, we'd like to keep refactoring to a minimum and to keep things as simple as possible as nearly everyone here has very little nodejs skills

src/config/index.ts
dgolubevs: Feature Flag will we always, just some will be removed and some will be added

src/utils/feature.flag.ts
dgolubevs: @mouhajer-ch 
could you please drop me full name of Toro? Thank you

@markpit 
It is good chance for ppl to learn TS
Also previous implementation do not use Typescript at all. Need to use TS power

@mouhajer-ch @markpit 
me need you decision - revert all changes or leave them with changes requested by @mwestacott ?
Thank you

src/utils/feature.flag.ts
dgolubevs: fixed

src/config/index.ts
dgolubevs: fixed

src/utils/feature.flag.ts
mouhajer-ch: Keep this code but you need to change the params expected

src/config/index.ts
mouhajer-ch: Revert this part, please

test/utils/feature.flag.spec.ts
```diff
@@ -1,60 +1,32 @@
...
+    'TRUE',
+    'on',
+    'oN',
+    '1',
+  ];
 
-  test("should return false if variable is 'OFF'", function() {
-    const active = isActiveFeature("OFF");
+  it.only.each(dataExpectTrue)('should return true if variable is %p', (val) => {
+    const active = isActiveFeature(<any>val);
```
mouhajer-ch, line 29: You do not need `<any>`. Use `any` when strictly necessary

src/config/index.ts
dgolubevs: Agree fixed

Could you please explain why need DATE at the end of feature flag env variable?

test/utils/feature.flag.spec.ts
dgolubevs: there is need casting to `any` because of
>  Argument of type 'string' is not assignable to parameter of type 'FeatureFlag'.

src/config/index.ts
markpit: It is the companies house standard - https://github.com/companieshouse/styleguides/blob/main/feature_flag.md

test/utils/feature.flag.spec.ts
dgolubevs: changed to test, import not required

src/config/index.ts
dgolubevs: Thank you, but this do not explain why need date
We have git blame to see when evn variable appear in project if will be need this information.

src/config/index.ts
dgolubevs: I revert all related existing feature flags
Keep this enum for future feature flags. This is more strict approach, that using now. FeatureFlag enum can be empty.

src/config/index.ts
mwestacott: Not sure why the date is part of the standard but can see it could be useful, e.g. having date in the name makes it immediately obvious how old a feature (flag) is, especially when viewing that feature flag in a config flag in a remote environment file that doesn't have any Git history behind it.

src/config/index.ts
dgolubevs: reverted

mwestacott: This now looks ok to me.

mouhajer-ch: Do not know why the feature flag does no have the date but it is just minor change. Fine to me

src/utils/feature.flag.ts
```diff
@@ -1,15 +1,8 @@
...
-  if (flag === undefined) {
-    return false;
-  }
-  const featureFlag = flag.toLowerCase();
-  return !(featureFlag === "false" ||
-    featureFlag === "0" ||
-    featureFlag === "off" ||
-    featureFlag === "");
+export const isActiveFeature = (flag: string): boolean => {
+  return ['1', 'true', 'on'].includes(flag.toString().toLowerCase());
```
markpit, line 7: the original version had a check for undefined, do we still need that check as undefined.toString() will throw error

test/utils/feature.flag.spec.ts
```diff
@@ -1,60 +1,32 @@
...
+  const dataExpectFalse = [
+    'false',
+    'faLse',
+    '0',
+    'off',
+    'oFF',
+    'kdjhskjf',
+  ];
+
+  test.only.each(dataExpectFalse)('should return false if variable is %p', (val) => {
```
markpit, line 14: just reading up on the use of .only - https://stackoverflow.com/a/44446669
it looks like it is used to only run one specified test in a file, i guess when you have a large amount of tests you can run one in isolation. I think we can remove the .only from these tests.

markpit:
i think it will need the updated 'main' branch merged in and the CH feature branch standard is to use lower case separated with hyphens. I think feature/ is ok, so something like feature/roe-1582-add-etc...  You don't need to change this one, just a guideline for future branches.

dgolubevs:
> i think it will need the updated 'main' branch merged in and the CH feature branch standard is to use lower case separated with hyphens. I think feature/ is ok, so something like feature/roe-1582-add-etc... You don't need to change this one, just a guideline for future branches.

I made by this manual https://companieshouse.atlassian.net/wiki/spaces/DEVOPS/pages/78151945/GitFlow+at+CH

test/utils/feature.flag.spec.ts
mouhajer-ch: Nope, but it is not a blocker

test/utils/feature.flag.spec.ts
mouhajer-ch: Nope, the IDE will complain `Cannot find name 'test'. Do you need to install type definitions for a test runner? Try ...` but it is not a blocker

src/utils/feature.flag.ts
mouhajer-ch: Just passing an empty string as default value in the flag argument will have improved readability and coverage

mwestacott (approved):

PR #463, opened 2022-11-14
Feature/roe 1607 refresh token before save and resume

mouhajer-ch:
### JIRA link

Resolves [ROE-1607](https://companieshouse.atlassian.net/browse/ROE-1607)

### Change description

Temporary fix for handling unauthorised response from OE calls

### Work checklist


src/service/unauthorised.handler.service.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param otherParams Parameters passed to the callback:
+ *        For the postOverseasEntity call we have (transactionID, appData, isSaveAndResumeFeatureActive)
+ *        For the putOverseasEntity call we have (transactionID, overseasEntityID, appData)
+ *
+ * @returns Promise< Resource<OverseasEntityCreated> | Resource<HttpStatusCode> | ApiErrorResponse >
+ */
+export const unauthorisedResponseHandler = async ( fnName: string, req: Request, session: Session, ...otherParams: any[] ) => {
+  let client = createOAuthApiClient(session);
+
+  let response = await client.overseasEntity[fnName](...otherParams);
```
markpit, line 24: As the function is going to be for the overseas-entities-api only (using client.overseasEntity), i was wondering if this function should live in the overseas.entities.service.ts and rename the function to something like 'makeOverseasEntitiesApiCallWithRetry'

mwestacott: Yes, agree. Even if it doesn't move, renaming the function name makes sense as that's what it is doing.

src/service/overseas.entities.service.ts
```diff
@@ -1,23 +1,23 @@
...
   session: Session,
   transactionId: string,
   isSaveAndResumeFeatureActive: boolean = false
 ): Promise<string> => {
-  const client = createOAuthApiClient(session);
-
-  const response = await client.overseasEntity.postOverseasEntity(
+  const response = await unauthorisedResponseHandler(
+    "postOverseasEntity",
+    req, session,
```
mwestacott, line 16: Suggest putting 'session' parameter on a separate line.

src/service/overseas.entities.service.ts
```diff
@@ -30,17 +30,18 @@ export const createOverseasEntity = async (
...
-  const client = createOAuthApiClient(session);
   const appData = getApplicationData(session);
 
   const transactionID = appData[Transactionkey] as string;
   const overseasEntityID = appData[OverseasEntityKey] as string;
 
-  const response = await client.overseasEntity.putOverseasEntity(
+  const response = await unauthorisedResponseHandler(
+    "putOverseasEntity",
+    req, session,
```
mwestacott, line 40: Yes, each param on a separate line as you've done for the others.

src/service/unauthorised.handler.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    logger.debugRequest(req, `${responseMsg} - ${JSON.stringify(response)}`);
+
+    const accessToken = await refreshToken(req, session);
+    logger.infoRequest(req, `New access token: ${accessToken}`);
+
+    client = createOAuthApiClient(session);
+    response = await client.overseasEntity[fnName](...otherParams);
+
+  }
+
```
mwestacott, line 38: Just wonder if it's worth having a debug log message something like "OE call successful" at this point?

src/service/unauthorised.handler.service.ts
mouhajer-ch: I could make it more generic by adding the sdk object after the client call, and call it as you suggested makeApiCallWithRetryOnUnauthorisedResponse. Something like this and call the function

```
export const makeApiCallWithRetryOnUnauthorisedResponse = async ( sdkObjName: string, fnName: string, req: Request, session: Session, ...otherParams: any[] ) => {
  let client = createOAuthApiClient(session);

  let response = await client[sdkObjName][fnName](...otherParams);

  if (response && response.httpStatusCode === 401){

    const responseMsg = `Retrying ${fnName} call after unauthorised response`;
    logger.debugRequest(req, `${responseMsg} - ${JSON.stringify(response)}`);

    const accessToken = await refreshToken(req, session);
    logger.infoRequest(req, `New access token: ${accessToken}`);

    client = createOAuthApiClient(session);
    response = await client[sdkObjName][fnName](...otherParams);

  }

  return response;

};
```

src/service/overseas.entities.service.ts
```diff
@@ -1,23 +1,23 @@
 import { Session } from "@companieshouse/node-session-handler";
 import { Request } from "express";
-import { createOAuthApiClient } from "./api.service";
 import { createAndLogErrorRequest, logger } from "../utils/logger";
 import { getApplicationData } from "../utils/application.data";
 import { Transactionkey, OverseasEntityKey } from "../model/data.types.model";
+import { unauthorisedResponseHandler } from "./unauthorised.handler.service";
```
mouhajer-ch, line 6: As suggested by the team, the file will be named `retry.handler.service.ts`

src/service/unauthorised.handler.service.ts
mwestacott: 👍 not sure it needs to be more generic though as in this context (within OE web) it will always be 'overseasEntity', no?

mwestacott: Ah right. So don't we need it on the POST call to create the OE too actually (for when save and resume feature is turned off)? Even when it's not turned off - when we call POST to create the initial OE with just the presenter info?

Basically, for all our remote calls from web to the OE API?

mouhajer-ch: The changes are already covering the POST and PUT for the OE calls

mwestacott: Shouldn't it be done as part of this PR though? Not the removal of the feature flag but the redundant refresh call.

mouhajer-ch: Nope

mwestacott: But this PR means that this line is now immediately redundant - hence why I think we remove it now.

mouhajer-ch: well, as I said it is part of the feature flag
```
    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REFRESH_TOKEN_29092022)) {
      const accessToken = await refreshToken(req, session);
      logger.infoRequest(req, `New access token: ${accessToken}`);
    }
```

mouhajer-ch: You do not like it, you disable the feature flag

mwestacott (approved):

PR #472, opened 2022-11-23
Feature/roe 1644 add controller to retrieve oe form from api

mouhajer-ch:
### JIRA link

Resolves [ROE-1644](https://companieshouse.atlassian.net/browse/ROE-1644)

### Change description

Add controller for the resume OE and GET service call to fetch form from API

### Work checklist


src/controllers/save.and.resume.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+ * related checkboxes for different pages
+ * @param session
+ * @param appData
+ * @param transactionId
+ * @param overseaEntityId
+ */
+const setWebApplicationData = (session: Session, appData: ApplicationData, transactionId: string, overseaEntityId: string) => {
+
+  appData[BeneficialOwnerIndividualKey] = (appData[BeneficialOwnerIndividualKey] as BeneficialOwnerIndividual[])
+    .map( e => { return { ...e, [ID]: uuidv4() }; } );
```
markpit, line 61: won't the uuid4() generate a different uuid to the one returned from mongo?

markpit, line 61: could you also rename the 'e's to be eg, boi or boo, bog etc..

src/service/overseas.entities.service.ts
```diff
@@ -55,3 +57,29 @@ export const updateOverseasEntity = async (req: Request, session: Session) => {
 
   logger.debugRequest(req, `Updated Overseas Entity, ${JSON.stringify(response)}`);
 };
+
+export const resumeOverseasEntity = async (
```
markpit, line 61: i'd call this function getOverseasEntity, it will be consistent then with the create and update functions in this file - and update the log messages to say 'get' instead of 'resume'

src/controllers/save.and.resume.controller.ts
mouhajer-ch: Hi, mongo does not return IDs for any OE object even BOs or MOs

mouhajer-ch: Will do

src/service/overseas.entities.service.ts
mouhajer-ch: Will do

src/controllers/save.and.resume.controller.ts
markpit: ah ok, i forgot we use them for adding/removing in the web app only.

src/service/overseas.entities.service.ts
```diff
@@ -55,3 +57,29 @@ export const updateOverseasEntity = async (req: Request, session: Session) => {
...
+  );
+
+  const infoMsg = `Transaction ID: ${transactionId}, OverseasEntity ID: ${overseasEntityId}`;
+
+  if (response.httpStatusCode !== 200) {
+    const errorMsg = `Something went wrong resuming Overseas Entity - ${infoMsg} - ${JSON.stringify(response)}`;
+    throw createAndLogErrorRequest(req, errorMsg);
+  }
+
+  logger.debugRequest(req, `Overseas Entity Resumed - ${infoMsg}`);
```
mwestacott, line 82: 'Retrieved' not 'Resumed'.

mwestacott (approved):

PR #494, opened 2022-12-05
ROE-1725 Add jurisdiction field to only the entity page

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1725

### Change description

Introduced separate public register component templates so the extra field on the entity page can be kept separate.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/views/beneficial-owner-other.html
+++ b/views/beneficial-owner-other.html
@@ -37,7 +37,7 @@
 
         {% set registerInCountryFormedText = "Is the other legal entity already on a public register in the country it was formed in?" %}
         {% set registrationNumberText = "Entity's registration number" %}
-        {% include "includes/inputs/fields/is-on-register-in-country-formed-input.html" %}
+        {% include "includes/inputs/fields/is-on-register-in-country-formed-input-no-jurisdiction-field.html" %}
 
         {% set dateText = "When did it become a beneficial owner for the overseas entity?" %}
         {% include "includes/inputs/date/start-date-input.html" %}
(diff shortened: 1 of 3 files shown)


views/includes/inputs/fields/is-on-register-in-country-formed-input.html
```diff
@@ -1,29 +1,26 @@
+{% if title == "Tell us about the overseas entity" %}
```
markpit, line 1: rather than have this include work out the hint text based on the title of the including page, it shoul d be possible to set hintText in the including page before you call this includes

eg
`{% set registerInCountryFormedText = "Is the other legal entity already on a public register in the country it was formed in?" %}
        {% set registrationNumberText = "Entity's registration number" %}
        {% set hintText = "Include the full name. For example, Jersey Financial Services Commission." %}
        {% include "includes/inputs/fields/is-on-register-in-country-formed-input.html" %}`

views/includes/inputs/fields/is-on-register-in-country-formed-input.html
```diff
@@ -1,29 +1,26 @@
...
-    name: "public_register_jurisdiction",
-    value: public_register_jurisdiction,
-    hint: {
-      text: "For example, Jersey."
-    },
-    label: {
-      text: "Jurisdiction"
-    }
-  }) }}
+  {% if title == "Tell us about the overseas entity" %}
```
markpit, line 21: for this include, you could set a boolean in the including page to indicate whether or not to include Jurisdiction, maybe have an 'includeJurisdiction = true/false' set before calling the is-on-register-in-country-formed-input

mouhajer-ch: Probably you do not need to set it to false, but it makes it more readable on MOC and BOC

mouhajer-ch (approved):

markpit (approved):

PR #514, opened 2022-12-13
Feature/roe 1749 what is the name of the overseas entity page

mouhajer-ch:
### JIRA link

Resolves [ROE-1749](https://companieshouse.atlassian.net/browse/ROE-1749)

### Change description

- Add MVC for `what is the name of the overseas entity` page
- Add navigation checks
- Move creation and update of OE and creation of transaction from presenter page to `what is the name of the overseas entity` page

### Work checklist


SVarallo (approved):
Looks ok to me 👍

src/controllers/presenter.controller.ts
```diff
@@ -29,26 +25,14 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
   }
 };
 
-export const post = async (req: Request, res: Response, next: NextFunction) => {
+export const post = (req: Request, res: Response, next: NextFunction) => {
```
mwestacott, line 28: Haven't you got to do an `await saveAndContinue(req, session);` in this method now/still?

src/middleware/navigation/check.condition.ts
```diff
@@ -27,8 +27,12 @@ export const checkIsSecureRegisterDetailsEntered = (appData: ApplicationData): b
   return checkHasSoldLandDetailsEntered(appData) && appData[IsSecureRegisterKey] === "0";
 };
 
+export const checkOverseasNameDetailsEntered = (appData: ApplicationData): boolean => {
+  return checkIsSecureRegisterDetailsEntered(appData) && (appData[EntityNameKey] || "").length  !== 0;
```
mwestacott, line 31: Minor - two spaces before !==

src/controllers/presenter.controller.ts
mouhajer-ch: Good spot, I will add it

test/controllers/overseas.name.controller.spec.ts
```diff
@@ -0,0 +1,140 @@
+jest.mock("ioredis");
+jest.mock('../../src/middleware/authentication.middleware');
+jest.mock('../../src/utils/application.data');
+jest.mock('../../src/middleware/navigation/is.secure.register.middleware');
+jest.mock('../../src/service/transaction.service');
+jest.mock('../../src/service/overseas.entities.service');
```
markpit, line 6: do you need to mock the service.availibility.middleware? it might be relying on the test setup process env defaults, just thinking might be better to mock it.

src/config/index.ts
```diff
@@ -67,6 +67,7 @@ export const SOLD_LAND_FILTER_PAGE = "sold-land-filter";
 export const TRUST_INFO_PAGE = "trust-information";
 export const USE_PAPER_PAGE = "use-paper";
 export const WHO_IS_MAKING_FILING_PAGE = "who-is-making-filing";
+export const OVERSEAS_NAME_PAGE = "overseas-name";
```
mwestacott, line 70: Would have probably called this constant ENTITY_NAME_PAGE (even though URL differs) and then used 'entity name' in all the code artefacts but not a problem.

test/controllers/overseas.name.controller.spec.ts
mouhajer-ch: As you said we have already `SHOW_SERVICE_OFFLINE_PAGE` set to false on test setup and the `FEATURE_FLAG_ENABLE_ROE_UPDATE`  is not part of our journey. I think is better not to mock it

src/config/index.ts
markpit: it does seem that the prototype page is called overseas-name so it is in keeping with the prototype ( i think our other page names follow this too)

PR #528, opened 2022-12-19
Add web validation to what is the name of the overseas entity page

mouhajer-ch:
### JIRA link

Resolves part of [ROE-1671](https://companieshouse.atlassian.net/browse/ROE-1671)

### Change description

Add web validation to `what is the name of the overseas entity` page

### Work checklist

### DIFF
--- a/src/validation/overseas.name.validation.ts
+++ b/src/validation/overseas.name.validation.ts
@@ -1,0 +1,7 @@
+import { body } from "express-validator";
+import { EntityNameKey } from "../model/data.types.model";
+import { checkEntityName } from "./custom.validation";
+
+export const overseasName = [
+  body(EntityNameKey).custom((value) => checkEntityName(value))
+];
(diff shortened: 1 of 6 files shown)


src/validation/overseas.name.validation.ts
```diff
@@ -0,0 +1,7 @@
+import { body } from "express-validator";
+import { EntityNameKey } from "../model/data.types.model";
+import { checkEntityName } from "./custom.validation";
+
+export const overseasName = [
+  body(EntityNameKey).custom((value) => checkEntityName(value))
```
mwestacott, line 6: I actually prefer 'entityName' as the term/name to use for all the code artefacts when referring to the field on this new screen - is it too much work to refactor and use this across the board in the web app?

If yes, then we should then really change this function name to be consistent and call it `checkOverseasName`.

mouhajer-ch: Could you please be more specific

mwestacott (approved):

PR #554, opened 2023-01-04
Roe 1240 dates one error message only

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1240

### Change description

Check date messages such that only the relevant one is returned both in the list and on the field.
All date checks now in custom verification where each error is sought in order.

Updated unit tests to check only one date error message is present
New tests to cover empty date field and the empty individual field scenarios.

### Work checklist

### Merge instructions

```
```

[ROE-1240]: https://companieshouse.atlassian.net/browse/ROE-1240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/validation/custom.validation.ts
```diff
@@ -78,11 +78,75 @@ export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", month
...
+          checkDateIsWithinLast3Months(ErrorMessages.IDENTITY_CHECK_DATE_NOT_WITHIN_PAST_3_MONTHS, dayStr, monthStr, yearStr);
+        }
+      }
+    }
+  }
+  return true;
+};
+
+export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+  const isStartDateFields = checkStartDate(dayStr, monthStr, yearStr);
```
mwestacott, line 96: It reads a bit odd that the `checkIdentityDate` function calls a `checkStartDate` function. Suggest that the `checkStartDate` function is renamed to be more generic - just `checkDate` perhaps (since it doesn't seem to have to take a 'start date' as input)?

Also, the variable `isStartDateFields` is confusing (at least to me!). How about `isDateValid` (assuming you rename that other function too)?

src/validation/custom.validation.ts
```diff
@@ -78,11 +78,75 @@ export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", month
...
+      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
+      if (isDateValid) {
+        checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
+      }
+    }
+  }
+  return true;
+};
+
+export const checkDateFieldsPresent = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
mwestacott, line 117: Look at the naming of the other methods in this file - they have 'Is' within the function name, so to be consistent, suggest this is renamed to `checkDateFieldsArePresent`.

src/validation/custom.validation.ts
```diff
@@ -78,11 +78,75 @@ export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", month
...
+      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE_OF_BIRTH, dayStr, monthStr, yearStr);
+      if (isDateValid) {
+        checkDateIsInPast(ErrorMessages.DATE_NOT_IN_PAST, dayStr, monthStr, yearStr);
+      }
+    }
+  }
+  return true;
+};
+
+export const checkDateOfBirthFieldsPresent = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
mwestacott, line 142: Suggest renaming to `checkDateOfBirthFieldsArePresent`.

src/validation/custom.validation.ts
```diff
@@ -78,11 +78,75 @@ export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", month
 
 export const checkOptionalDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
   if ( dayStr !== "" || monthStr !== "" || yearStr !== "" ) {
-    checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
-    checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
+    const areDateFieldsPresent = checkDateFieldsPresent(dayStr, monthStr, yearStr);
```
markpit, line 81: I think line 80 is redundant now as it checks if one of the 3 str is present, but then the next line will throw error if any of them are missing. So you can probably remove line 80 and it's closing bracket, as the line 81 checks will overrule them.

markpit: actually, i can see now that because it is for optional dates, if all fields are missing then it jumps to the return true; - so seems ok, ignore previous comment

src/validation/custom.validation.ts
mattch1: Renamed this to "mandatory date" to compliment the existing optional date.

src/validation/custom.validation.ts
mattch1: Renamed as above

src/validation/custom.validation.ts
mattch1: Renamed as above

src/validation/custom.validation.ts
```diff
@@ -78,11 +78,75 @@ export const checkDateValueIsValid = (errMsg: string, dayStr: string = "", month
...
+          checkDateIsWithinLast3Months(ErrorMessages.IDENTITY_CHECK_DATE_NOT_WITHIN_PAST_3_MONTHS, dayStr, monthStr, yearStr);
+        }
+      }
+    }
+  }
+  return true;
+};
+
+export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+  const areMandatoryDateFields = checkMandatoryDate(dayStr, monthStr, yearStr);
```
mwestacott, line 96: Variable name still a bit confusing (to me). How about `isDateValid` as I think that's what the `checkMandatoryDate` function is doing?

`areMandatoryDateFields` makes me think 'are mandatory date fields.... what?

mattch1: Renamed once more.

mwestacott: Thanks

mwestacott: I was expecting variable name (`areMandatoryDateFields`) and not function name (`checkMandatoryDate`) to change. I think function name is fine - most of the other functions in that file start with 'check...'.

mattch1: Reverted function name back.

mwestacott: Thanks.

mwestacott (approved):

markpit (approved):

PR #563, opened 2023-01-06
ROE-1240 Focus field problem depends on field errors being kept separate

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1240

### Change description

This addresses the loss of focus, that is to say clicking the error and it taking us directly to the dd mm yyyy field when an error is thrown against one one of them.

Had to ensure that each of the the date fields are dealt with separately to get the focus back and yet still ensure only one error is displayed in order to do this the checks have to be performed twice first on each field separately then a second time when we check for other errors so that a second error is not thrown when a single empty field is detected

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -93,14 +93,23 @@ export const checkOptionalDate = (dayStr: string = "", monthStr: string = "", ye
 };
 
 export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
-  const isMandatoryDateValid = checkMandatoryDate(dayStr, monthStr, yearStr);
-  if (isMandatoryDateValid) {
-    checkDateIsWithinLast3Months(ErrorMessages.IDENTITY_CHECK_DATE_NOT_WITHIN_PAST_3_MONTHS, dayStr, monthStr, yearStr);
+  const isDatePresent = checkDateIsNotCompletelyEmpty(ErrorMessages.ENTER_DATE, dayStr, monthStr, yearStr);
+  if (isDatePresent) {
+    const areAllDateFieldsPresent = checkAllDateFieldsArePresent(dayStr, monthStr, yearStr);
+    if (areAllDateFieldsPresent) {
+      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
+      if (isDateValid) {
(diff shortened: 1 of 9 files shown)


src/validation/custom.validation.ts
```diff
@@ -93,14 +93,23 @@ export const checkOptionalDate = (dayStr: string = "", monthStr: string = "", ye
 };
 
 export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
-  const isMandatoryDateValid = checkMandatoryDate(dayStr, monthStr, yearStr);
-  if (isMandatoryDateValid) {
-    checkDateIsWithinLast3Months(ErrorMessages.IDENTITY_CHECK_DATE_NOT_WITHIN_PAST_3_MONTHS, dayStr, monthStr, yearStr);
+  const isDatePresent = checkDateIsNotCompletelyEmpty(ErrorMessages.ENTER_DATE, dayStr, monthStr, yearStr);
```
markpit, line 96: the checks in here look the same as those in checkStartDate plus the extra last 3 months check. You could rename checkStartDate to something like checkDateIsValidAndInPast and call that from within checkIdentityDate.
Or if you want to keep the checkStartDate name, you could create a new function like checkDateIsValidAndInPast and have checkStartDate just call checkDateIsValidAndInPast, then have checkIdentityDate call checkDateIsValidAndInPast and checkDateIsWithinLast3Months

mattch1: Trying to reuse code like this will present a complication here because of checkAllDateFieldsArePresent returning true or false and so checkDateIsValidAndInPast would have to as well as we would require a check from the result of that. This may not be obvious from just looking at the code, you may need to check the branch out and try this proposal to see what I mean here.

src/validation/custom.validation.ts
```diff
@@ -141,13 +171,35 @@ export const checkDateOfBirth = (dayStr: string = "", monthStr: string = "", yea
   return true;
 };
 
-export const checkDateOfBirthFieldsArePresent = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+export const checkDateOfBirthFieldDay = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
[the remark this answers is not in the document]
mattch1: Can reuse code here, updated.

src/validation/custom.validation.ts
markpit: i see now, just tried it out locally, if the areAllDateFieldsPresent returns false then the last line of the method we are re-using will return true instead. ok, ignore this comment.

markpit (approved):

PR #576, opened 2023-01-10
ROE-1814 entity page content update

SVarallo:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1814

### Change description
Updating the public register declaration on the entity page

### Work checklist

### Merge instructions

```
```


views/entity.html
```diff
@@ -80,9 +80,9 @@ <h1 class="govuk-heading-xl">{{ title }}</h1>
 
       {{ govukInsetText({
         html: '
-         <h2 class="govuk-heading-m">Information shown on the public register</h2>
-         <p>The email address will not be shown.</p>
-         <p>All other information provided in this section will be shown on the public Register of Overseas Entities.</p>
+         <h2 class="govuk-heading-m">What information we’ll show on the public register</h2>
```
mattch1, line 83: Did you merge from main? We should use the  {% include "includes/inset-text.html" %} and set the value for paragraph one, the second paragraph and head are in the include file.

SVarallo: Wasn't aware this existed, I'll try it out 👍

SVarallo: This will require a change for paragraphTwo as well as paragraphOne, I think it will be simpler to keep it separate than use the includes

markpit: the inset-text will set the second paragraph to 
`We will show all other information you provide on this screen on the public Register of Overseas Entities.` if you supply a paragraph one. Looks like the only difference compared to yours is the word 'the', so might be better if they were all consistent. I'd add a question/comment on teh story and say other insetText boxes show that sentence without the 'the'.

markpit: and i'd probably use the inset-text as well for consistency

SVarallo: Seems they want the version of the paragraph containing the 'the' and will add another story to bring the others in line with that

markpit: ok, leave as is and then we can bring it inline by using the inset-text in the new story

markpit (approved):

PR #600, opened 2023-01-17
ROE-1811 Update text for presenter page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1811

### Change description

Add hint text to the top of the page.
Update public register text.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/views/presenter.html
+++ b/views/presenter.html
@@ -43,12 +43,8 @@
           value: email
         }) }}
 
-        {{ govukInsetText({
-          html: '
-            <h2 class="govuk-heading-m">What information we’ll show on the public register</h2>
-            <p>We will not show any of the information you provide on this screen on the public Register of Overseas Entities.</p>
-          '
-        }) }}
+        {% set paragraphTwo = '<p>We will not show any of the information you provide on this screen on the public Register of Overseas Entities.</p>' %}
+        {% include "includes/inset-text.html" %}
 
         {% include "includes/save-and-continue-button.html" %}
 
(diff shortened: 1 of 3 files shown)


views/presenter.html
```diff
@@ -44,8 +45,8 @@ <h1 class="govuk-heading-xl">Who can we contact about this application?</h1>
 
         {{ govukInsetText({
           html: '
-            <h2 class="govuk-heading-m">Information shown on the public register</h2>
-            <p>None of the information provided in this section will be shown on the public Register of Overseas Entities.</p>
+            <h2 class="govuk-heading-m">What information we’ll show on the public register</h2>
```
mouhajer-ch, line 48: Why you do not want to use the `inset-text` as [here](https://github.com/companieshouse/overseas-entities-web/blob/main/views/includes/inset-text.html)?

mattch1: Inset text will add a bar to the left hand side, which is not what has been requested. Also this is more consistent with top page hint text for other pages such as due diligence.

mouhajer-ch: Could you please check the screenshot on your story. Anyway they should be all in the same style, so eventually check with UX and confirm that all the inset text related to the information shown on the public register should have all the same style.

mattch1: Sorry thought you were talking about something else , yes we can use inset text here.

mattch1: Updated this.

mouhajer-ch (approved):

PR #603, opened 2023-01-18
ROE-1105 Ensure single email error

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1105

### Change description

As with the date fields we want to display only one relevant mail error message at a time

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -1,6 +1,6 @@
 // Custom validation utils - For now checking is not empty
 
-import { VALID_CHARACTERS } from "./regex/regex.validation";
+import { VALID_CHARACTERS, VALID_EMAIL_FORMAT } from "./regex/regex.validation";
 import { DateTime } from "luxon";
 import { ErrorMessages } from "./error.messages";
 import { trustType } from "../model";
@@ -294,4 +294,36 @@ const checkCorporatesAddress = (trust: trustType.Trust, addressMaxLength: number
       }
     }
   }
+};
+
+export const validateEmail = (email: string, maxLength: number) => {
+  const isEmailPresent = checkEmailIsPresent(email);
+  if (isEmailPresent) {
+    const isWithinLengthLimit = checkIsWithinLengthLimit(email, maxLength);
+    if (isWithinLengthLimit) {
+      checkCorrectIsFormat(email);
+    }
+  }
+  return true;
+};
+
+const checkEmailIsPresent = (email: string) => {
+  if (email === undefined || email === "") {
+    throw new Error(ErrorMessages.EMAIL);
+  }
+  return true;
+};
+
+const checkIsWithinLengthLimit = (email: string, maxLength: number) => {
+  if (email.length > maxLength) {
+    throw new Error(ErrorMessages.MAX_EMAIL_LENGTH);
+  }
+  return true;
+};
+
+const checkCorrectIsFormat = (email: string) => {
+  if (!email.match(VALID_EMAIL_FORMAT)) {
+    throw new Error(ErrorMessages.EMAIL_INVALID_FORMAT);
+  }
+  return true;
 };
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -297,13 +297,9 @@ const checkCorporatesAddress = (trust: trustType.Trust, addressMaxLength: number
 };
 
 export const validateEmail = (email: string, maxLength: number) => {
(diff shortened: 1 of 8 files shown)


src/validation/custom.validation.ts
```diff
@@ -295,3 +295,35 @@ const checkCorporatesAddress = (trust: trustType.Trust, addressMaxLength: number
     }
   }
 };
+
+export const validateEmail = (email: string, maxLength: number) => {
+  const isEmailPresent = checkEmailIsPresent(email);
```
mouhajer-ch, line 300: You do not need to check it with ifs something like the following will do
```
export const validateEmail = (email: string, maxLength: number) => {
    checkEmailIsPresent(email);
    checkIsWithinLengthLimit(email, maxLength);
    checkCorrectIsFormat(email);
    
    return true;
}
```

src/validation/custom.validation.ts
```diff
@@ -295,3 +295,35 @@ const checkCorporatesAddress = (trust: trustType.Trust, addressMaxLength: number
...
+    const isWithinLengthLimit = checkIsWithinLengthLimit(email, maxLength);
+    if (isWithinLengthLimit) {
+      checkCorrectIsFormat(email);
+    }
+  }
+  return true;
+};
+
+const checkEmailIsPresent = (email: string) => {
+  if (email === undefined || email === "") {
```
mouhajer-ch, line 311: Probably email needs to be trimmed before checking it.

src/validation/fields/email.validation.ts
```diff
@@ -1,17 +1,12 @@
...
-    .matches(VALID_EMAIL_FORMAT).withMessage(ErrorMessages.EMAIL_INVALID_FORMAT)
+    .custom((value, { req }) => validateEmail(req.body["email"], 256))
 ];
 
 export const contact_email_validations = [
   body("contact_email").trim()
-    .not().isEmpty().withMessage(ErrorMessages.EMAIL)
-    .isLength({ max: 250 }).withMessage(ErrorMessages.MAX_EMAIL_LENGTH)
-    .matches(VALID_EMAIL_FORMAT).withMessage(ErrorMessages.EMAIL_INVALID_FORMAT)
+    .custom((value, { req }) => validateEmail(req.body["contact_email"], 250))
```
mouhajer-ch, line 11: Probably I will confirm with analyst why the difference between email length, I can not find the file related to the length of related chips columns

src/validation/custom.validation.ts
mattch1: This seems to work so removed the conditions

src/validation/custom.validation.ts
mattch1: Trim the email as it is passed in.

src/validation/fields/email.validation.ts
mattch1: That's a wider question about emails in the web app; this does not relate tot he work for this.

src/validation/custom.validation.ts
```diff
@@ -295,3 +295,31 @@ const checkCorporatesAddress = (trust: trustType.Trust, addressMaxLength: number
     }
   }
 };
+
+export const validateEmail = (email: string, maxLength: number) => {
+  checkEmailIsPresent(email.trim());
+  checkIsWithinLengthLimit(email.trim(), maxLength);
```
mouhajer-ch, line 301: I would not trim it for every call, just create single variable and pass it to the three functions, furthermore we do not need to return true from those function, their values is not used

src/validation/fields/email.validation.ts
mouhajer-ch: Ok, agreed probably we should rise it to the standup

src/validation/custom.validation.ts
mattch1: I'm not sure the trim call really matters either way. We cannot remove return true as the validation that calls this requires it. You can verify this by running the unit tests without returning true.

mattch1: Updated the trim to use a variable.

mouhajer-ch: `return true` should be removed from `checkEmailIsPresent`, `checkIsWithinLengthLimit` and `checkCorrectIsFormat`, it is not used, we need to `return true` for the main email custom validator,  if successfu,l as expected by `express-validator`.

mattch1: Updated this.

mouhajer-ch (approved):

PR #630, opened 2023-01-27
roe-1550 fixing non display of page

okoye-companieshouse:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1550

### Change description

Changed the view, route, test and controller

### Work checklist

### Merge instructions

- fix: fixing the non display of the trust-interrupt page

### DIFF
--- a/src/controllers/trust.interrupt.controller.ts
+++ b/src/controllers/trust.interrupt.controller.ts
@@ -5,6 +5,7 @@ import { getApplicationData } from '../utils/application.data';
 import { mapCommonTrustDataToPage } from '../utils/trust/common.trust.data.mapper';
 import * as PageModel from '../model/trust.page.model';
 import { safeRedirect } from '../utils/http.ext';
+import { TrustKey } from '../model/trust.model';
 
 const TRUST_INTERRUPT_TEXTS = {
   title: 'You now need to submit trust information',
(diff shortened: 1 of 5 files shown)


src/controllers/trust.interrupt.controller.ts
```diff
@@ -24,12 +25,12 @@ type TrustDetailPageProperties = {
 const getPageProperties = (
   req: Request,
 ): TrustDetailPageProperties => {
-  const trustId = req.params[config.ROUTE_PARAM_TRUST_ID];
 
   const appData = getApplicationData(req.session);
+  const trustId = `${(appData[TrustKey] ?? []).length + 1}`;
```
JoshFen95, line 30: Would we ever need the trustId for this page? If this page is at the very start of the trust journey then no trust data will be displayed or recorded in the session.  Same goes for the references to `trustData` and `mapCommonTrustDataToPage`. I don't think we'll ever have it here. 
Correct me if I'm wrong

test/controllers/trust.interrupt.controller.spec.ts
```diff
@@ -63,12 +62,15 @@ describe('Trust Interrupt controller', () => {
...
+      // When no session increment by 1
+      const noSessionTrustId = '1';
+
       get(mockReq, mockRes, mockNext);
 
       expect(mockRes.redirect).not.toBeCalled();
 
       expect(mapCommonTrustDataToPage).toBeCalledTimes(1);
-      expect(mapCommonTrustDataToPage).toBeCalledWith(mockAppData, trustId);
+      expect(mapCommonTrustDataToPage).toBeCalledWith(mockAppData, noSessionTrustId);
```
JoshFen95, line 73: If you agree with us not using trustData in the controller we shouldn't need any `mapCommonTrustDataToPage ` checks

JoshFen95 (approved):
Thanks for doing the changes. Looks good to me

PR #635, opened 2023-01-31
Feature/uar 170 oe number validate

AndreKramer7654:
UAR-170 https://companieshouse.atlassian.net/browse/UAR-170

Move the lookup of company profile to the oe number query controller. This allows error handling without loading twice.

Show how an Update specific Model can be used. The Update model in ApplicationData holds the date of creation of the company - this allows error handling on company lookup without loading the company profile twice. While adding the date_of_creation so that it can be displayed on the next (confirmation) page.

The model that is dependent on the entity being updated  is reset if the entity number changes. This ensures that any data that was entered for one entity is not used for another entity.

Probably easier to look at the controller code than the diffs as the code moved over.

Tested with second entity added to assets in overseas-entity-web service.


src/controllers/update/overseas.entity.query.controller.ts
```diff
@@ -14,25 +17,54 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
 
-export const post = (req: Request, res: Response, next: NextFunction) => {
+export const post = async (req: Request, res: Response, next: NextFunction) => {
   try {
     logger.debugRequest(req, `POST ${config.OVERSEAS_ENTITY_QUERY_PAGE}`);
-    const oeNumber = req.body[OeNumberKey];
 
-    setExtraData(req.session, { ...getApplicationData(req.session), [OeNumberKey]: oeNumber });
-    return res.redirect(config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL);
+    const entityNumber = req.body[EntityNumberKey];
```
jkayyala, line 32: maybe we can add getCompanyProfile and creating the error into validation.middleware(separate method), so that the error is created in validation.middleware, not sure if its right to call companyProfile from validation.middleware

AndreKramer7654: I did try this approach but found that the middleware uses a different request type (in meta data passed in) so could not call out to get the company profile. Also, it seems wrong to call out to services in the validation middleware which presumably should just be checking format and other data integrity / validations.

AndreKramer7654: Also this avoids calling getProfile twice which itself has downsides

bantao-ch: > I did try this approach but found that the middleware uses a different request type (in meta data passed in) so could not call out to get the company profile. Also, it seems wrong to call out to services in the validation middleware which presumably should just be checking format and other data integrity / validations.

I agree -- Validation middleware should just do the validations 👍

src/middleware/navigation/check.condition.ts
```diff
@@ -65,7 +66,15 @@ export const checkBOsOrMOsDetailsEntered = (appData: ApplicationData): boolean =
...
+  return checkHasAppData(appData) && (appData[EntityNumberKey] || "").length !== 0;
+};
+
+export const checkHasOverseasEntity = (appData: ApplicationData): boolean => {
+  return checkHasAppData(appData) && appData[EntityKey] !== undefined;
+};
+
+export const checkHasDateOfCreation = (appData: ApplicationData): boolean => {
+  return checkHasAppData(appData) && appData[UpdateKey]?.date_of_creation !== undefined;
 };
```
corybudd-ch, line 78: I think the first condition of each of these should be calling the previous check like the register ones?
e.g. should `checkOverseasEntityNumberEntered`'s first condition to `checkHasOverseasEntity` rather than `checkHasAppData`

AndreKramer7654: This check is combined (using &&) with checkHasEntity in the middleware so that is checking the previous page looked up the entity & date of creation.

corybudd-ch: Cool, sorry maybe left comment in wrong place .. what about the other one, `checkOverseasEntityNumberEntered`, is that ok? Should it be updated to the following now you have added new `checkHasOverseasEntity`:

```
export const checkOverseasEntityNumberEntered = (appData: ApplicationData): boolean => {
  return checkHasOverseasEntity(appData) && (appData[EntityNumberKey] || "").length !== 0;
};
```

AndreKramer7654: No, getting the entity requires the number. The hasOverseasEntity check is just a more detailed check for the confirmation page (checks entity was loaded) with others then checking that the entity number is present. It's probably more than is needed as you now can't get forward from the query page unless the entity is found (error handling ensures this). But I've added an explicit check for the entity number to the hasOverseasEntity validation (and also added some more simple check tests).

corybudd-ch (approved):
I've approved and read through the Model re-work ODP, still not 100% sure if that work is required under this ticket UAR-170 or not, as it's a fair amount of work it should maybe have its own ticket/PR but I'll leave that up to others!

jkayyala (approved):

src/model/update.reset.ts
```diff
@@ -0,0 +1,29 @@
+
+import { ApplicationData } from "./application.model";
```
markpit, line 2: I'd move this file under /utils as it's more a utility function to reset the model rather than the model itself.

AndreKramer7654: done

markpit (approved):

PR #641, opened 2023-02-01
roe-1734 display individual trustees ...

okoye-companieshouse:
### JIRA link

Display individual trustees in what you have added so far page

### Change description

- Modified the template file
- Modified the controller
- Fixed the form fields in the individual beneficial page field

### Work checklist

### DIFF
--- a/src/controllers/trust.involved.controller.ts
+++ b/src/controllers/trust.involved.controller.ts
@@ -10,7 +10,7 @@ import { getApplicationData } from '../utils/application.data';
 import { mapCommonTrustDataToPage } from '../utils/trust/common.trust.data.mapper';
 import { mapTrustWhoIsInvolvedToPage } from '../utils/trust/who.is.involved.mapper';
 import { FormattedValidationErrors, formatValidationError } from '../middleware/validation.middleware';
-import { GeneralTrustee } from '../model/trust.model';
+import { IndividualTrustee } from '../model/trust.model';
 import { getTrusteesFromTrust } from '../utils/trusts';
 
 const TRUST_INVOLVED_TEXTS = {
@@ -36,7 +36,7 @@ type TrustInvolvedPageProperties = {
     beneficialOwnerTypeTitle: Record<string, string>;
     trusteeTypeTitle: Record<string, string>;
     trusteeType: typeof TrusteeType;
-    individualTrusteeData: GeneralTrustee[];
+    individualTrusteeData: IndividualTrustee[];
     checkYourAnswersUrl: string;
     beneficialOwnerUrlDetach: string;
     trustData: CommonTrustData,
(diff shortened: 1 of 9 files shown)


views/trust-involved.html
```diff
@@ -45,21 +45,53 @@ <h1 class="govuk-heading-xl">{{ pageParams.title }}</h1>
...
+                "detach-trustee-from-trust-button"
+              )]
             }
           }),
           trustees
         ) %}
       {% endfor %}
 
+      {% for trustee in pageData.trusteeData %}
+        {% set type = trustee.type %}
```
JoshFen95, line 73: trustee.type will bring the wrong data from the trust. type is linked to `roleWithinTrustType` on the `GeneralTrustee` model. That type is relating to the type of indiviudal trustee so whether it's settlor, beneficiary, grantor or interested person. 
We are after the actual type of trustee whether it's a individual, legal entity or historical. GeneralTrustee doesn't contain that data. In my PR I have created TrusteeItem which contains the trustee type and the trustee name. 

I've mapped data from the session object to this. You would need to join the first and last name of the individual so it fits in the name field. 

The way the code is currenly implemented, we wouldn't be getting the `trustee type` displaying on the page and it would also stop the legal entity work from displaying. 

I think we need to do something similar for `individual` like i have for `legal entity` in my PR. 

https://github.com/companieshouse/overseas-entities-web/pull/640

Happy to have a call if you want

JoshFen95:
The code looks good but I feel we have both done a similar task in a different way. We also have the "displaying former bo" story to complete at some point too and I feel that these 3 tasks should be implemented in a  consistent way as they are so similar in purpose and function. 
I'm happy to go either way tbh, I'll ask Steve to have a look and see what he thinks would be the best approach

src/controllers/trust.involved.controller.ts
```diff
@@ -34,6 +36,7 @@ type TrustInvolvedPageProperties = {
     beneficialOwnerTypeTitle: Record<string, string>;
     trusteeTypeTitle: Record<string, string>;
     trusteeType: typeof TrusteeType;
+    trusteeData: GeneralTrustee[];
```
JoshFen95, line 39: Should `trusteeData` maybe be called `individualTrusteeData` or something similar because the other 2 trustee types (legal entity(corporate) and historical(former) don't fit into `GeneralTrustee `

okoye-companieshouse: Yes it should forgot that change thanks

views/trust-involved.html
```diff
@@ -44,6 +44,31 @@ <h1 class="govuk-heading-xl">{{ pageParams.title }}</h1>
         ) %}
       {% endfor %}
 
+      {% set trusteeList = [] %}
```
JoshFen95, line 47: Similar thing here, this is just a list of individualTrustees.

JoshFen95 (approved):

PR #652, opened 2023-02-07
Roe 1935 - TL -validate submission for some identified

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1935

### Change description

Replaced submit anchor link with button. 
Therefore, trust link now handled in code rather than template.
Added new url and submit POST function.

### Work checklist

### Merge instructions

```
```


src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -15,24 +15,17 @@ import { isActiveFeature } from '../utils/feature.flag';
...
 
     logger.infoRequest(req, `${config.BENEFICIAL_OWNER_TYPE_PAGE} hasTrusts=${hasTrusts}`);
 
     return res.render(config.BENEFICIAL_OWNER_TYPE_PAGE, {
       backLinkUrl: config.BENEFICIAL_OWNER_STATEMENTS_URL,
       templateName: config.BENEFICIAL_OWNER_TYPE_PAGE,
+      submitUrl: config.BENEFICIAL_OWNER_TYPE_SUBMIT_URL,
       hasTrusts,
       ...appData,
-      pageParams: {
```
mwestacott, line 30: I'm wondering why trusts related code had to change - at this point, why is this block removed?

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -46,6 +39,18 @@ export const post = (req: Request, res: Response) => {
   return res.redirect(getNextPage(req.body[BeneficialOwnerTypeKey]));
 };
 
+export const postSubmit = (req: Request, res: Response) => {
```
mwestacott, line 42: Again, confused by this change... maybe it's not yours? In which case I would do a local rebase and re-push, which will hopefully help GtHub display only your changes.

src/validation/custom.validation.ts
```diff
@@ -279,6 +280,31 @@ export const checkBeneficialOwnerType = (beneficialOwnersStatement: string, valu
   return true;
 };
 
+export const checkBeneficialOwnerSubmission = (req) => {
```
mwestacott, line 283: I'd make this method name plural - `checkBeneficialOwnersSubmission`

src/validation/custom.validation.ts
```diff
@@ -279,6 +280,31 @@ export const checkBeneficialOwnerType = (beneficialOwnersStatement: string, valu
...
+  const appData: ApplicationData = getApplicationData(req.session);
+  if (appData.beneficial_owners_statement === BeneficialOwnersStatementType.SOME_IDENTIFIED_ALL_DETAILS) {
+    if (!hasBeneficialOwners(appData)) {
```
mwestacott, line 286: Does this work for AC5 in the story? It says we should show the error even if only one Corporate BO is added.

test/controllers/beneficial.owner.type.controller.spec.ts
```diff
@@ -241,4 +235,150 @@ describe("BENEFICIAL OWNER TYPE controller", () => {
...
+        [ManagingOfficerKey]: [],
+        [ManagingOfficerCorporateKey]: []
+      });
+
+      const resp = await request(app)
+        .post(config.BENEFICIAL_OWNER_TYPE_SUBMIT_URL);
+
+      expect(resp.status).toEqual(200);
+      expect(resp.text).toContain(BENEFICIAL_OWNER_TYPE_PAGE_HEADING_SOME_IDENTIFIED);
+      expect(resp.text).toContain(ErrorMessages.MUST_ADD_MANAGING_OFFICER);
```
mwestacott, line 294: I think this is testing AC5 but error message in that scenario is "You need to add at least one beneficial owner", so that would be the string to check for. If this test then starts failing it backs up my comment above about the wrong validation logic.

src/controllers/beneficial.owner.type.controller.ts
mattch1: No longer needed it was linked to the old anchor tag:


 {% if (hasTrusts) %}
            {% set nextPage = pageParams.urlToTrust %}
          {% else %}
            {% set nextPage = pageParams.urlToCheckYourAnswers %}
          {% endif %}

          <a href= "{{ nextPage }}"
             role="button"
             draggable="false"
             class="govuk-button govuk-button--secondary govuk-!-margin-top- 6"
             data-module="govuk-button"
             id="submit"
          >No more <span class="govuk-visually-hidden"> {{ hiddenTextValue }} </span> to add</a>

src/controllers/beneficial.owner.type.controller.ts
mattch1: It is mine, what's the matter with it?

src/validation/custom.validation.ts
mattch1: AC 5 should read corporate MO not BO as AC2 covers this.

test/controllers/beneficial.owner.type.controller.spec.ts
mattch1: See comment above - it is testing AC 2.

src/validation/custom.validation.ts
mattch1: Updated.

src/controllers/beneficial.owner.type.controller.ts
mwestacott: Clarified that this logic has moved (from being related to the anchor tag to now being on a button).

src/validation/custom.validation.ts
mwestacott: AC updated after discussion with Karen. Code now matches requirements.

views/beneficial-owner-type.html
```diff
@@ -103,27 +103,15 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ boTypesTitle }}</h1>
           items: boTypes
         }) }}
 
-        <button class="govuk-button" data-module="govuk-button" id="add">Add<span class="govuk-visually-hidden"> {{ whatToAdd }} </span></button>
+        <button class="govuk-button" data-module="govuk-button" id="add" formaction="/register-an-overseas-entity/beneficial-owner-type" >Add<span class="govuk-visually-hidden"> {{ whatToAdd }} </span></button>
```
mwestacott, line 106: Just as the `submitUrl` is passed in to the template as a param, I think this URL should be passed in too (or, we pass neither in). Needs to be a consistent approach though. I think passing the URLs in is the right way to go though.

markpit (approved):

views/beneficial-owner-type.html
mattch1: Tried that and it won't work because the submit version of the page url does not have a get, so there is nothing to pass that value in and so there is only one way to guarantee that this original url will be present for when press that add button in cases where we have tried to press no more to add and an error has been thrown with the "/submit" url.

mattch1: Not only is there no get for "/submit" but there is not point where we would be able to use a get, it is fundamentally as POST url.

mwestacott: Can't understand why this is not possible. I will check the branch out and investigate.

mwestacott: All clear now. A quirk of using 'formAction' and the validation middleware that doesn't add back in params when page is rerendered.

Agreed with @markpit that we'll hard-code both the URLs for consistency.

mwestacott (approved):

PR #656, opened 2023-02-08
ROE-1933-TL-Fix date error highlighting and only fire one date error at a time

markpit:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1933

### Change description
Added year length checks to the other validation checks to prevent > 1 date error being fired. This means that the date length check is a higher priority than the 'all in one' date check (invalid date etc...)

Had to add a for loop in the template includes to determine what part of the date to highlight red. Works of the field name being present in the errors list. Tried to make it generic but struggled with variable scopes when used as an include. So had to repeat it for each of the 3 date includes.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/beneficial.owner.individual.validation.ts
+++ b/src/validation/beneficial.owner.individual.validation.ts
@@ -20,6 +20,9 @@ export const beneficialOwnerIndividual = [
     .isLength({ max: 160 })
     .withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH)
     .matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
+
+  ...date_of_birth_validations,
+
   body("nationality")
     .not().isEmpty({ ignore_whitespace: true }).withMessage(ErrorMessages.NATIONALITY)
     .matches(VALID_CHARACTERS).withMessage(ErrorMessages.NATIONALITY_INVALID_CHARACTERS),
@@ -31,12 +34,10 @@ export const beneficialOwnerIndividual = [
   body("is_service_address_same_as_usual_residential_address")
     .not().isEmpty().withMessage(ErrorMessages.SELECT_IF_SERVICE_ADDRESS_SAME_AS_USER_RESIDENTIAL_ADDRESS),
 
-  ...start_date_validations,
-
   ...usual_residential_address_validations,
   ...usual_residential_service_address_validations,
 
-  ...date_of_birth_validations,
+  ...start_date_validations,
 
   ...nature_of_control_validations
 ];
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -65,12 +65,10 @@ export const checkDateIsWithinLast3Months = (errMsg: string, day: string = "", m
   return true;
 };
 
-export const checkDateValueIsValid = (invalidDateErrMsg: string, yearLengthErrMsg: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+export const checkDateValueIsValid = (invalidDateErrMsg: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
   const day = parseInt(dayStr), month = parseInt(monthStr), year = parseInt(yearStr);
   if (isNaN(day) || isNaN(month) || isNaN(year) || !DateTime.utc(year, month, day).isValid) {
     throw new Error(invalidDateErrMsg);
-  } else if (yearStr.length !== 4) {
-    throw new Error(yearLengthErrMsg);
   }
 
   return true;
@@ -80,7 +78,7 @@ export const checkOptionalDate = (dayStr: string = "", monthStr: string = "", ye
   if ( dayStr !== "" || monthStr !== "" || yearStr !== "" ) {
     const areDateFieldsPresent = checkAllDateFieldsArePresent(dayStr, monthStr, yearStr);
     if (areDateFieldsPresent) {
-      const isOptionalDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, ErrorMessages.YEAR_LENGTH, dayStr, monthStr, yearStr);
+      const isOptionalDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
       if (isOptionalDateValid) {
         const isDateInThePast = checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
         if (isDateInThePast) {
@@ -97,7 +95,7 @@ export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", ye
   if (isDatePresent) {
     const areAllDateFieldsPresent = checkAllDateFieldsArePresent(dayStr, monthStr, yearStr);
     if (areAllDateFieldsPresent) {
-      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, ErrorMessages.YEAR_LENGTH, dayStr, monthStr, yearStr);
+      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
       if (isDateValid) {
         const isDatePastOrToday = checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
         if (isDatePastOrToday) {
@@ -114,7 +112,7 @@ export const checkStartDate = (dayStr: string = "", monthStr: string = "", yearS
   if (isDatePresent) {
     const areAllDateFieldsPresent = checkAllDateFieldsArePresent(dayStr, monthStr, yearStr);
     if (areAllDateFieldsPresent) {
-      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, ErrorMessages.YEAR_LENGTH, dayStr, monthStr, yearStr);
+      const isDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, dayStr, monthStr, yearStr);
       if (isDateValid) {
         checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
       }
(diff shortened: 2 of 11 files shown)


src/validation/custom.validation.ts
```diff
@@ -65,26 +65,26 @@ export const checkDateIsWithinLast3Months = (errMsg: string, day: string = "", m
...
 export const checkOptionalDate = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
-  if ( dayStr !== "" || monthStr !== "" || yearStr !== "" ) {
-    const areDateFieldsPresent = checkAllDateFieldsArePresent(dayStr, monthStr, yearStr);
-    if (areDateFieldsPresent) {
-      const isOptionalDateValid = checkDateValueIsValid(ErrorMessages.INVALID_DATE, ErrorMessages.YEAR_LENGTH, dayStr, monthStr, yearStr);
-      if (isOptionalDateValid) {
-        const isDateInThePast = checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
-        if (isDateInThePast) {
-          checkDateIsWithinLast3Months(ErrorMessages.IDENTITY_CHECK_DATE_NOT_WITHIN_PAST_3_MONTHS, dayStr, monthStr, yearStr);
+  if (yearStr.length === 0 || yearStr.length === 4) {
```
mwestacott, line 78: Somebody reading this code/logic months (even years) later will probably wonder why this method doesn't check the date if the year is empty or fully populated.

Could we maybe put the `if (yearStr.length === 0 || yearStr.length === 4)` logic check in its own private function (also called by the other 'checkDate' functions below) and add a comment in that function why we only proceed if the conditions are met? I think it's because this check is done previously in the validation logic chain and we need to suppress duplicate messages if more than one condition fails?

src/validation/custom.validation.ts
```diff
@@ -110,36 +112,45 @@ export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", ye
...
+        if (isDateValid) {
+          checkDateIsInPastOrToday(ErrorMessages.DATE_NOT_IN_PAST_OR_TODAY, dayStr, monthStr, yearStr);
+        }
       }
     }
   }
   return true;
 };
 
 export const checkDateFieldDay = (message: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
mwestacott, line 130: Minor, but would rename param name to `dayMissingMessage` to match what you've done below.

src/validation/custom.validation.ts
```diff
@@ -110,36 +112,45 @@ export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", ye
...
-    throw new Error(message);
+  if (yearStr.length === 0 || yearStr.length === 4) {
+    if (dayStr === "" && monthStr !== "" && yearStr !== "") {
+      throw new Error(message);
+    }
   }
   return true;
 };
 
 export const checkDateFieldMonth = (message: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
mwestacott, line 139: `monthMissingMessage`

test/controllers/beneficial.owner.gov.controller.spec.ts
```diff
@@ -444,11 +444,11 @@ describe("BENEFICIAL OWNER GOV controller", () => {
       expect(mockSaveAndContinue).not.toHaveBeenCalled();
     });
 
-    test(`renders the current page ${config.BENEFICIAL_OWNER_GOV_PAGE} with only INVALID_DATE error when invalid characters are used`, async () => {
+    test(`renders the current page ${config.BENEFICIAL_OWNER_GOV_PAGE} with only INVALID_DATE error when invalid date is used`, async () => {
       const beneficialOwnerGov = { ...REQ_BODY_BENEFICIAL_OWNER_GOV_FOR_DATE_VALIDATION };
-      beneficialOwnerGov["start_date-day"] =  "a";
-      beneficialOwnerGov["start_date-month"] = "b";
-      beneficialOwnerGov["start_date-year"] = "c";
+      beneficialOwnerGov["start_date-day"] =  "11";
```
mwestacott, line 449: This looks like a valid test but testing when non-digits are supplied also looks valid. Should/could we keep the original test too?

src/validation/custom.validation.ts
markpit: yes this occurred to me too after raising this PR, the month check (eg.) will only run if year is partly populated. I'm going to look at moving the if (year) check outside of these custom functions and try and have it in the body("..." validators

test/controllers/beneficial.owner.gov.controller.spec.ts
markpit: i can add an extra test to cover the previous scenario of non-digits

views/includes/inputs/date/identity-date-input.html
```diff
@@ -1,5 +1,55 @@
...
+{% if isIdentityDateMonthError %}
+  {% set identity_date_classes_month = identity_date_classes_month + "--error" %}
+{% endif %}
+{% if isIdentityDateYearError %}
+  {% set identity_date_classes_year = identity_date_classes_year + "--error" %}
+{% endif %}
+
 {{ govukDateInput({
-  errorMessage: errors.identity_date if errors,
+  errorMessage: errorMess,
```
mattch1, line 52: Could we call this errorMsg?

markpit: went with identityDateErrorMsg so it doesn't clash with any other vars when included in templates

src/validation/beneficial.owner.individual.validation.ts
```diff
@@ -20,6 +20,9 @@ export const beneficialOwnerIndividual = [
     .isLength({ max: 160 })
     .withMessage(ErrorMessages.MAX_LAST_NAME_LENGTH)
     .matches(VALID_CHARACTERS).withMessage(ErrorMessages.LAST_NAME_INVALID_CHARACTERS),
+
+  ...date_of_birth_validations,
```
markpit, line 24: just moved the order the validators run so that the error messages are in field order in teh box at top of page

src/validation/custom.validation.ts
markpit: updated

src/validation/custom.validation.ts
markpit: updated

src/validation/custom.validation.ts
markpit: updated

test/controllers/beneficial.owner.gov.controller.spec.ts
markpit: updated

src/validation/custom.validation.ts
```diff
@@ -110,36 +118,42 @@ export const checkIdentityDate = (dayStr: string = "", monthStr: string = "", ye
...
   return true;
 };
 
-export const checkDateFieldYear = (message: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+export const checkDateFieldYear = (yearMissingMessage: string, yearLengthMessage: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
   if (yearStr === "" && dayStr !== "" && monthStr !== "") {
-    throw new Error(message);
+    throw new Error(yearMissingMessage);
+  }
+  if (yearStr.length !== 0 && yearStr.length !== 4) {
```
mwestacott, line 155: Rather than this line of logic, could you maybe make use of your new `isYearEitherMissingOrCorrectLength` function and NOT its result?

src/validation/fields/date.validation.ts
```diff
@@ -9,35 +9,47 @@ import {
 } from "../custom.validation";
 import { ErrorMessages } from "../error.messages";
 
+// to prevent more than 1 error reported on the date fields we check if the year is valid before doing some checks.
+// This means that teh year check is checked before some others
```
mwestacott, line 13: 'the year check'

src/validation/fields/date.validation.ts
```diff
@@ -9,35 +9,47 @@ import {
...
 export const start_date_validations = [
-  body("start_date")
+  body("start_date-day")
+    .if(body("start_date-year").isLength({ min: 4, max: 4 }))
```
mwestacott, line 16: As discussed, this 'if' here is a bit different to how we check it in the custom functions, but agreed to leave as is for now.

It seems odd that we do the year check when validating the month field (line 19) but we don't also check the day...

test/controllers/beneficial.owner.gov.controller.spec.ts
```diff
@@ -444,7 +444,7 @@ describe("BENEFICIAL OWNER GOV controller", () => {
       expect(mockSaveAndContinue).not.toHaveBeenCalled();
     });
 
-    test(`renders the current page ${config.BENEFICIAL_OWNER_GOV_PAGE} with only INVALID_DATE error when invalid characters are used`, async () => {
+    test(`renders the current page ${config.BENEFICIAL_OWNER_GOV_PAGE} with only YEAR_LENGTH error when invalid charcters are used`, async () => {
```
mwestacott, line 447: 'characters are used'

test/controllers/managing.officer.controller.spec.ts
```diff
@@ -591,11 +591,11 @@ describe("MANAGING_OFFICER controller", () => {
       expect(mockSaveAndContinue).not.toHaveBeenCalled();
     });
 
-    test(`renders the current page ${MANAGING_OFFICER_PAGE} with only INVALID_DATE_OF_BIRTH error when invalid characters are used`, async () => {
+    test(`renders the current page ${MANAGING_OFFICER_PAGE} with only INVALID_DATE_OF_BIRTH error when invalid date is used`, async () => {
       const managingOfficer = { ...REQ_BODY_MANAGING_OFFICER_FOR_DATE_VALIDATION };
-      managingOfficer["date_of_birth-day"] =  "a";
```
mwestacott, line 596: Would reinstate this original test, as you did above but also have the new one with invalid numbers.

views/includes/inputs/date/date-of-birth-input.html
```diff
@@ -1,5 +1,58 @@
+{% set dobErrorMessage = '' %}
+{% set isDobDayError = false %}
+{% set isDobMonthError = false %}
+{% set isDobYearError = false %}
+
+{% if errors %}
```
mwestacott, line 6: Maybe just add some comments to this code added to the templates - which bits are doing what? I think some bits are highlighting the fields in red, another bit is placing focus in the field etc.

src/validation/fields/date.validation.ts
markpit: updated

views/includes/inputs/date/date-of-birth-input.html
markpit: updated

src/validation/custom.validation.ts
markpit: updated

src/validation/fields/date.validation.ts
markpit: updated

test/controllers/beneficial.owner.gov.controller.spec.ts
markpit: updated

test/controllers/managing.officer.controller.spec.ts
markpit: updated

src/validation/fields/date.validation.ts
mwestacott: Not sure there was specifically an update for this. At least I can't see one and I thought we agreed to leave as is. Am I missing something?

src/validation/custom.validation.ts
mwestacott: Nice.

mwestacott (approved):

src/validation/fields/date.validation.ts
markpit: thought i updated everything, nothing to do for this one then

PR #671, opened 2023-02-16
Bugfix/roe 1775 fixbugs for 1775 and 1776

okoye-companieshouse:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1775
https://companieshouse.atlassian.net/browse/ROE-1776

### Change description

Addressing test failures


test/utils/trust/individual.trustee.mapper.spec.ts
```diff
@@ -44,71 +44,73 @@ describe('Individual Beneficial Owner page Mapper Service', () => {
...
+                               sa_address_line1: mockFormData.service_address_line_1,
+                               sa_address_line2: mockFormData.service_address_line_2,
+                               sa_address_locality: mockFormData.service_address_town,
+                               sa_address_region: mockFormData.service_address_county,
+                               sa_address_country: mockFormData.service_address_country,
+                               sa_address_postal_code: mockFormData.service_address_postcode,
+                               sa_address_care_of: '',
+                               sa_address_po_box: '',
+                             });
+                           });
```
JoshFen95, line 93: Formatting is off here

src/model/trust.page.model.ts
```diff
@@ -40,13 +40,13 @@ interface TrustHistoricalBeneficialOwnerFormCommon {
...
   type: RoleWithinTrustType,
   forename: string,
   surname: string,
-  date_of_birth_day: string,
-  date_of_birth_month: string,
-  date_of_birth_year: string,
+  dateOfBirthDay: string,
+  dateOfBirthMonth: string,
+  dateOfBirthYear: string,
   nationality: string,
```
JoshFen95, line 50: Why are we renaming these variables? 
If they do need renaming for whatever reason then the other vars with multiple words should follow the same standard. Here there is a miss of camelCase and snake_case

src/controllers/check.your.answers.controller.ts
```diff
@@ -27,8 +27,9 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     logger.infoRequest(req, `${config.CHECK_YOUR_ANSWERS_PAGE} hasTrusts=${hasTrusts}`);
 
     let backLinkUrl: string = config.BENEFICIAL_OWNER_TYPE_URL;
+
     if (hasTrusts) {
-      backLinkUrl = config.TRUST_INFO_PAGE;
+      backLinkUrl = `${config.TRUSTS_URL}`;
```
JoshFen95, line 32: I think the back link change is being tracked by a Jira ticket so might be best to make those changes on a seperate PR or see if you can bring that jira into this one. I would double check with Steve

src/model/trust.page.model.ts
```diff
@@ -92,28 +92,30 @@ type TrustLegalEntityForm = {
...
+service_address_country?: string;
+service_address_postcode?: string;
+governingLaw: string;
+legalForm: string;
+public_register_name?: string;
+public_register_jurisdiction?: string;
+registration_number?: string;
+is_service_address_same_as_principal_address: yesNoResponse;
+is_on_register_in_country_formed_in: yesNoResponse;
 };
```
JoshFen95, line 119: Why don't need these changes/fields for the historic/former trustees. They are not present on the web page or the data model.
https://roe-v13.herokuapp.com/register/trusts/trust-historical-beneficial-owner

views/includes/check-your-answers/beneficial-owner-legal.html
```diff
@@ -4,7 +4,7 @@
```
JoshFen95, line 4: maybe call this file 'legal-entity-trustee' so it is easier to tell what is used for and follows a similar naming to the 'individual-trustee' you created.

src/model/trust.page.model.ts
okoye-companieshouse: Steve mentioned the same issue. It is as a result of how the UI makes a post request with a mixture of both, as I reuse 2 modules that do both. This is a candidate for refactoring but will make the PR larger than needs to be and I will have to fix lots of models. Hence, used it the way it is. Many models have done same but that will require changing in a separate ticket.

src/model/trust.page.model.ts
okoye-companieshouse: In order to determine if the home address is same as the service address there is need to track the option clicked by the user in the page. same as the country of registration.

views/includes/check-your-answers/beneficial-owner-legal.html
okoye-companieshouse: Yep you are right my naming is not always the best 👍

src/model/trust.page.model.ts
okoye-companieshouse: Used as flags when radio buttons selected to track value

src/controllers/trust.individual.beneficial.owner.controller.ts
```diff
@@ -75,6 +75,7 @@ const post = (
     const trustId = req.params[config.ROUTE_PARAM_TRUST_ID];
     const url = `${config.TRUST_ENTRY_URL}/${trustId}${config.TRUST_INVOLVED_URL}`;
 
+
```
SteveHicksCH, line 78: remove this extra line which is not related to this PR

src/model/trust.page.model.ts
JoshFen95: image

Historical bo doesn't use the address fields or the register fields in the model and it is not captured on the web page. 

legal entity trustees capture both address types and the reg so i see the need there. Individual trustee capture both address type so I see the need there. Don't think we need it for former/historical.

https://roe-v13.herokuapp.com/register/trusts/trust-historical-beneficial-owner

JoshFen95 (approved):

PR #683, opened 2023-02-20
UAR-94: Add Auth Code page

jkayyala:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-94

### Change description
- Added Auth code page in Update journey, the scope is defaulted to admin.write-full, this will be changed in later ticket

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/middleware/company.authentication.middleware.ts
+++ b/src/middleware/company.authentication.middleware.ts
@@ -1,0 +1,32 @@
+import { Request, Response, NextFunction } from 'express';
+import { logger } from '../utils/logger';
+import { authMiddleware, AuthOptions } from "@companieshouse/web-security-node";
+import { CHS_URL, OVERSEAS_ENTITY_PRESENTER_URL } from '../config';
+import { getApplicationData } from "../utils/application.data";
+import { ApplicationData } from "../model";
+import { EntityNumberKey } from "../model/data.types.model";
+
+export const companyAuthentication = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `COMPANY AUTHENTICATION REQUEST`);
+
+    const appData: ApplicationData = getApplicationData(req.session);
+    const entityNumber: string = appData?.[EntityNumberKey] as string;
+
+    if (entityNumber) {
+      const authMiddlewareConfig: AuthOptions = {
+        chsWebUrl: CHS_URL,
+        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+        companyNumber: entityNumber
+      };
+      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+      return authMiddleware(authMiddlewareConfig)(req, res, next);
+    } else {
+      logger.errorRequest(req, "No entity number to authenticate against");
+      throw new Error(`OE number not found in session "${ entityNumber }"`);
+    }
+  } catch (err) {
+    logger.errorRequest(req, err);
+    next(err);
+  }
+};
--- a/src/middleware/company.authentication.middleware.ts
+++ b/src/middleware/company.authentication.middleware.ts
@@ -5,9 +5,6 @@ import { CHS_URL, OVERSEAS_ENTITY_PRESENTER_URL } from '../config';
 import { getApplicationData } from "../utils/application.data";
 import { ApplicationData } from "../model";
 import { EntityNumberKey } from "../model/data.types.model";
-import { SignInInfoKeys } from '@companieshouse/node-session-handler/lib/session/keys/SignInInfoKeys';
-import { ISignInInfo } from '@companieshouse/node-session-handler/lib/session/model/SessionInterfaces';
-import { SessionKey } from '@companieshouse/node-session-handler/lib/session/keys/SessionKey';
 
 export const companyAuthentication = (req: Request, res: Response, next: NextFunction) => {
   try {
@@ -15,19 +12,15 @@ export const companyAuthentication = (req: Request, res: Response, next: NextFun
 
     const appData: ApplicationData = getApplicationData(req.session);
     const entityNumber: string = appData?.[EntityNumberKey] as string;
-    const signInInfo: ISignInInfo = req.session?.get<ISignInInfo>(SessionKey.SignInInfo) || {};
 
     if (entityNumber) {
-      if (!isAuthorisedForCompany(entityNumber, signInInfo)) {
-        const authMiddlewareConfig: AuthOptions = {
-          chsWebUrl: CHS_URL,
-          returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
-          companyNumber: entityNumber
-        };
-        logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
-        return authMiddleware(authMiddlewareConfig)(req, res, next);
-      }
-      next();
+      const authMiddlewareConfig: AuthOptions = {
+        chsWebUrl: CHS_URL,
+        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+        companyNumber: entityNumber
+      };
+      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+      return authMiddleware(authMiddlewareConfig)(req, res, next);
     } else {
       logger.errorRequest(req, "No entity number to authenticate against");
       throw new Error("OE number not found in session");
@@ -37,11 +30,3 @@ export const companyAuthentication = (req: Request, res: Response, next: NextFun
     next(err);
   }
 };
-
-function isAuthorisedForCompany(companyNumber: string, signInInfo: ISignInInfo): boolean {
-  const authorisedCompany = signInInfo[SignInInfoKeys.CompanyNumber];
-  if (!authorisedCompany) {
-    return false;
-  }
-  return authorisedCompany.localeCompare(companyNumber) === 0;
-}
--- a/src/middleware/company.authentication.middleware.ts
+++ b/src/middleware/company.authentication.middleware.ts
@@ -5,28 +5,43 @@ import { CHS_URL, OVERSEAS_ENTITY_PRESENTER_URL } from '../config';
 import { getApplicationData } from "../utils/application.data";
 import { ApplicationData } from "../model";
 import { EntityNumberKey } from "../model/data.types.model";
+import { SignInInfoKeys } from '@companieshouse/node-session-handler/lib/session/keys/SignInInfoKeys';
+import { ISignInInfo } from '@companieshouse/node-session-handler/lib/session/model/SessionInterfaces';
+import { SessionKey } from '@companieshouse/node-session-handler/lib/session/keys/SessionKey';
 
 export const companyAuthentication = (req: Request, res: Response, next: NextFunction) => {
   try {
-    logger.debugRequest(req, `COMPANY AUTHENTICATION REQUEST`);
+    logger.debugRequest(req, `Company Authentication Request`);
 
     const appData: ApplicationData = getApplicationData(req.session);
     const entityNumber: string = appData?.[EntityNumberKey] as string;
+    const signInInfo: ISignInInfo = req.session?.get<ISignInInfo>(SessionKey.SignInInfo) || {};
 
     if (entityNumber) {
-      const authMiddlewareConfig: AuthOptions = {
-        chsWebUrl: CHS_URL,
-        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
-        companyNumber: entityNumber
-      };
-      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
-      return authMiddleware(authMiddlewareConfig)(req, res, next);
+      if (!isAuthorisedForCompany(entityNumber, signInInfo)) {
+        const authMiddlewareConfig: AuthOptions = {
+          chsWebUrl: CHS_URL,
+          returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+          companyNumber: entityNumber
+        };
+        logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+        return authMiddleware(authMiddlewareConfig)(req, res, next);
+      }
+      next();
     } else {
       logger.errorRequest(req, "No entity number to authenticate against");
-      throw new Error(`OE number not found in session "${ entityNumber }"`);
+      throw new Error("OE number not found in session");
     }
   } catch (err) {
     logger.errorRequest(req, err);
     next(err);
   }
 };
+
+function isAuthorisedForCompany(companyNumber: string, signInInfo: ISignInInfo): boolean {
+  const authorisedCompany = signInInfo[SignInInfoKeys.CompanyNumber];
+  if (!authorisedCompany) {
+    return false;
+  }
+  return authorisedCompany.localeCompare(companyNumber) === 0;
+}
--- a/src/routes/index.ts
+++ b/src/routes/index.ts
@@ -317,19 +317,21 @@ router.route(config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL)
 router.route(config.OVERSEAS_ENTITY_PRESENTER_URL)
   .all(
     authentication,
+    companyAuthentication,
     navigation.hasOverseasEntityNumber
   )
   .get(overseasEntityPresenter.get)
   .post(...validator.presenter, checkValidations, overseasEntityPresenter.post);
 
-router.get(config.UPDATE_CHECK_YOUR_ANSWERS_URL, authentication, updateCheckYourAnswers.get);
-router.post(config.UPDATE_CHECK_YOUR_ANSWERS_URL, authentication, updateCheckYourAnswers.post);
+router.get(config.UPDATE_CHECK_YOUR_ANSWERS_URL, authentication, companyAuthentication, updateCheckYourAnswers.get);
+router.post(config.UPDATE_CHECK_YOUR_ANSWERS_URL, authentication, companyAuthentication, updateCheckYourAnswers.post);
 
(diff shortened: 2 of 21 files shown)


src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (entityNumber) {
+      const authMiddlewareConfig: AuthOptions = {
+        chsWebUrl: CHS_URL,
+        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+        companyNumber: entityNumber
+      };
+      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+      return authMiddleware(authMiddlewareConfig)(req, res, next);
+    } else {
+      logger.errorRequest(req, "Invalid OE number");
```
AndreKramer7654, line 25: Maybe "No entity number to authenticate against"?

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    logger.debugRequest(req, `COMPANY AUTHENTICATION REQUEST`);
+
+    const appData: ApplicationData = getApplicationData(req.session);
+    const entityNumber: string = appData?.[EntityNumberKey] as string;
+
+    if (entityNumber) {
+      const authMiddlewareConfig: AuthOptions = {
+        chsWebUrl: CHS_URL,
+        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
```
AndreKramer7654, line 19: Maybe this should go back to the Overseas Entity Query page? To allow another OE number to be entered (assuming they somehow entered & confirmed the wrong one). At least it would show them the number again...

src/routes/index.ts
```diff
@@ -286,7 +287,7 @@ router.route(config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL)
     navigation.hasOverseasEntity
   )
   .get(confirmOverseasEntityDetails.get)
-  .post(confirmOverseasEntityDetails.post);
+  .post(companyAuthentication, confirmOverseasEntityDetails.post);
```
AndreKramer7654, line 290: Are we going to add it to all the routes after the auth code page?

AndreKramer7654: Should it not be after the confirmation page? i.e. user enters number; confirms they have right one; then asked for auth code.

test/__mocks__/text.mock.ts
```diff
@@ -64,6 +64,7 @@ export const TRANSACTION_ERROR = "Transaction API request returned no response";
 export const CREATE_OE__MSG_ERROR = "Something went wrong creating Overseas Entity, transactionId =";
 export const FOUND_REDIRECT_TO = "Found. Redirecting to";
 export const REDIRECT_TO_SIGN_IN_PAGE = "User not authenticated, redirecting to sign in page, status_code=302";
+export const REDIRECT_TO_COMPANY_SIGN_IN_PAGE = "Found. Redirecting to test/signin?return_to=/update-an-overseas-entity/update-presenter&company_number=SA000392";
```
AndreKramer7654, line 67: What is the company_number on the query string? Should that be OE111129 instead?

test/middleware/company.authentication.middleware.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  test('should catch error when oeNumber is not present in session ', () => {
+    req.session = undefined;
+
+    companyAuthentication(req, res, next);
+
+    expect(next).toHaveBeenCalledTimes(1);
+    expect(logger.errorRequest).toHaveBeenCalledTimes(2);
+  });
+
+  test('should catch error when calling companyAuthentication middleware and call next(err)', () => {
```
AndreKramer7654, line 58: "should call next when catching error"?

test/middleware/company.authentication.middleware.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    return next();
+  };
+});
+
+describe('Company Authentication middleware', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    res.locals = {};
+  });
+
```
AndreKramer7654, line 36: Do we need tests on some of the update pages with no auth code previously entered?

src/middleware/company.authentication.middleware.ts
jkayyala: @AndreKramer7654 , validation on this page should not allow this to happen , so ideally should never occur, just added security....

test/middleware/company.authentication.middleware.spec.ts
jkayyala: @AndreKramer7654  with navigation checks we won't be able to test this, and it will need an IT if we need to check if the page is accessed without auth code...

src/routes/index.ts
bantao-ch, line 290: We will need to add the companyAuthentication middleware on all subsequent pages starting from this one.
For the same reason the login authentication is there on each page aa well.

bantao-ch (changes requested):

src/middleware/company.authentication.middleware.ts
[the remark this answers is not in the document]
jkayyala: @bantao-ch , if req.originalUrl is added it lands you back to the confirmation page, should it not land you to the next page , the UX of the journey does not feel good.

src/routes/index.ts
jkayyala: can add it to all subsequent pages, but with the navigation checks a user cannot skip the Auth code page, same with login it only needs to be added to the first page , but it does make it clear to understand on what is needed, so will add it to all subsequent pages

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+        companyNumber: entityNumber
+      };
+      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+      return authMiddleware(authMiddlewareConfig)(req, res, next);
+    } else {
+      logger.errorRequest(req, "No entity number to authenticate against");
```
AndreKramer7654, line 25: Should this redirect to OE number query page?

src/routes/index.ts
AndreKramer7654: I think it should. What if we get navigation wrong anywhere. Authentication middleware is everywhere too.

test/middleware/company.authentication.middleware.spec.ts
AndreKramer7654: You could mock out the navigation check (to pass) but maybe this is not needed.

src/middleware/company.authentication.middleware.ts
jkayyala: can do that change, think it can be only tested in unit test and not locally since it will be hard to replicate the scenario

src/routes/index.ts
jkayyala: @AndreKramer7654 , have added this everywhere as its easier to understand from the code and avoiding the risk of navigation checks being wrong, this will result in calls to web-security-node from all the pages.

test/middleware/company.authentication.middleware.spec.ts
jkayyala: have mocked the navigation check as this as its now added to all the pages

src/routes/index.ts
AndreKramer7654: I'm hoping it calls the web-security SDK but then finds that we are already authenticated without calling a remote API or link so that any overhead is minimal.

src/middleware/company.authentication.middleware.ts
AndreKramer7654: Suggestion only. It really should not happen so maybe it's better to error.

src/routes/index.ts
AndreKramer7654: Done now.

AndreKramer7654 (approved):

sjahir (approved):

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+      throw new Error(`OE number not found in session "${ entityNumber }"`);
```
mwestacott, line 26: Is there any point including `${ entityNumber }` in the output message here as we've already established that none was found?

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+import { logger } from '../utils/logger';
+import { authMiddleware, AuthOptions } from "@companieshouse/web-security-node";
+import { CHS_URL, OVERSEAS_ENTITY_PRESENTER_URL } from '../config';
+import { getApplicationData } from "../utils/application.data";
+import { ApplicationData } from "../model";
+import { EntityNumberKey } from "../model/data.types.model";
+
+export const companyAuthentication = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `COMPANY AUTHENTICATION REQUEST`);
```
mwestacott, line 11: Any reason this output message is in capitals? It will make it stand out in the logs, but does it need to?

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+    const entityNumber: string = appData?.[EntityNumberKey] as string;
+
+    if (entityNumber) {
+      const authMiddlewareConfig: AuthOptions = {
+        chsWebUrl: CHS_URL,
+        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+        companyNumber: entityNumber
+      };
+      logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+      return authMiddleware(authMiddlewareConfig)(req, res, next);
```
mouhajer-ch, line 23: By using the `web-security-node` you will end-up having duplicating checks between our authentication middleware and the new one.
I would have created locally a separate middleware with a simple logic, similar to [here](https://github.com/companieshouse/web-security-node/blob/5acbe195eea859a0021e2d911f35637f2f533bb7/src/index.ts#L59)

```
function isAuthorisedForCompany(companyNumber: string, signInInfo: ISignInInfo): boolean {
  const authorisedCompany = signInInfo[SignInInfoKeys.CompanyNumber]
  if (!authorisedCompany) {
    return false
  }

  return authorisedCompany.localeCompare(companyNumber) === 0
}
```

src/routes/index.ts
```diff
@@ -286,24 +287,26 @@ router.route(config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL)
...
 
 router.route(config.OVERSEAS_ENTITY_PRESENTER_URL)
   .all(
     authentication,
+    companyAuthentication,
```
mouhajer-ch, line 295: I do not know if we need to add `companyAuthentication` in each endpoint if we have a navigation check in place, probably check it the first time when the company is posted and use the navigation logic already in place

mouhajer-ch (commented):
It looks good to me, but I have added two suggestion/improvement

src/middleware/company.authentication.middleware.ts
jkayyala: @mwestacott , no reason to and can surely make it follow the existing format

src/routes/index.ts
jkayyala: @mouhajer-ch , had followed the approach of adding it only on the first endpoint but was suggested adding it to all as an extra security measure in case navigation checks are missed.

src/middleware/company.authentication.middleware.ts
jkayyala: @mwestacott , its an edge case as this should not generally occur , so can remove it

src/middleware/company.authentication.middleware.ts
jkayyala: @mouhajer-ch ,  but won't this expose the checking logic to overseas-web which currently delegates the checking logic to web-security-node

mouhajer-ch: What do you mean?

jkayyala: @mouhajer-ch This logic currently resides in web-security-node , so was not sure if we need to add this into overseas-web

mouhajer-ch: Nope

bantao-ch: FYI we are using the current pattern used in other CH AuthCode integrations such as confirmation-statement

bantao-ch: @jkayyala I think @mouhajer-ch suggestion on review is a good one --
we also do the authenticated check in our middleware and REDIRECT only if not authenticated. This does save trips/redirects to just do the authenticated checks !

    if (entityNumber && !isAuthorisedForCompany ){
      const authMiddlewareConfig: AuthOptions = {
        chsWebUrl: CHS_URL,
        returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,

mwestacott: I agree with @mouhajer-ch that the current implementation within the `web-security-node` `authMiddleware` does not seem ideal, as it bundles user _authentication_ checks together with company _authorisation_ checks (and we are already handling user authentication in our app).

Appreciate that you are just following the same pattern as used in other CH web apps but I think it would worth be looking at the solution that implements a local company auth middleware check. If it doesn't work, fine, we can fallback to this approach but worthy of consideration I think.

jkayyala: @mwestacott , implemented the changes @mouhajer-ch had suggested, can you please review

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+      logger.errorRequest(req, "No entity number to authenticate against");
+      throw new Error("OE number not found in session");
+    }
+  } catch (err) {
+    logger.errorRequest(req, err);
+    next(err);
+  }
+};
+
+function isAuthorisedForCompany(companyNumber: string, signInInfo: ISignInInfo): boolean {
```
mouhajer-ch, line 41: Create an util function on `session.ts` file and use `getSignInInfo`, it will make the code much readable

src/middleware/company.authentication.middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (entityNumber) {
+      if (!isAuthorisedForCompany(entityNumber, signInInfo)) {
+        const authMiddlewareConfig: AuthOptions = {
+          chsWebUrl: CHS_URL,
+          returnUrl: OVERSEAS_ENTITY_PRESENTER_URL,
+          companyNumber: entityNumber
+        };
+        logger.infoRequest(req, `Invoking company authentication with (${ entityNumber }) present in session`);
+        return authMiddleware(authMiddlewareConfig)(req, res, next);
```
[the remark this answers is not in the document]
mwestacott: Yes, also confused by this. If it can work this way then the dependency on `web-security-node` can be removed I think.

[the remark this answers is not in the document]
AndreKramer7654: We need to supply a scope for update to be granted the right permission in a follow on PR. That is best done in the security component I would think.

src/routes/index.ts
```diff
@@ -328,24 +333,32 @@ router.route(config.UPDATE_DUE_DILIGENCE_URL)
...
     navigation.hasWhoIsMakingUpdate
   )
   .get(updateDueDiligenceOverseasEntity.get)
   .post(...validator.overseasEntityDueDiligence, checkValidations, updateDueDiligenceOverseasEntity.post);
 
 router.route(config.OVERSEAS_ENTITY_REVIEW_URL)
-  .all(authentication)
+  .all(
```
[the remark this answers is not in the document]
AndreKramer7654: out of cope for this PR

src/middleware/company.authentication.middleware.ts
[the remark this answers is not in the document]
jkayyala: Also the web-security-node does lot of checks , checks chsWebUrl, !req.session are we saying that we need to do the checks here, if so won't this make the web-security-node redundant, think we should stick to the pattern of delegating the checks to web-security-node

src/routes/index.ts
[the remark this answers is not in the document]
mouhajer-ch: Not really, it is the reason that you need to add company check for all Update journey when you really need it once, but if you like duplicating things

src/middleware/company.authentication.middleware.ts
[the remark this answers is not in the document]
mouhajer-ch: Probably @AndreKramer7654 you are right, create a new middleware in the `web-security-node` without duplicating the Authentication, just the company authorisation logic

src/middleware/company.authentication.middleware.ts
mwestacott: Resolving this comment as covered by newer comments below.

src/routes/index.ts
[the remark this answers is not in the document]
mwestacott: > out of scope for this PR

Is there a separate JIRA story or task to add these in later @AndreKramer7654 ? Normally we've implemented these 'safety checks' at the point of introducing the routing.

src/middleware/company.authentication.middleware.ts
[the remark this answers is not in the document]
mwestacott: Resolving this conversation as way forwards agreed and documented in the comment/conversation below.

src/middleware/company.authentication.middleware.ts
mwestacott: 👍 Thanks @bantao-ch . Please could you send me a link to the JIRA ticket once created.

src/routes/index.ts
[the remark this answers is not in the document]
AndreKramer7654: https://companieshouse.atlassian.net/browse/UAR-347 created to re view this

mwestacott (approved):

PR #697, opened 2023-02-24
Feature/roe 1566 add another trust page

JoshFen95:
### JIRA link

[ROE-1566](https://companieshouse.atlassian.net/browse/ROE-1566)

Video of changes can be found on JIRA ticket

### Change description
- create add trust page
- Modify routing to handle new add trust page and having multiple trusts in the journey(including back links)
- create and modify relevant tests

**Changes maybe considered out of scope**
- Modify `saveTrustInApp`. This method was causing the display order of trusts on the check-your-answers page to change when a trust was updated after navigating back to a previously filled out page. This was because every time the method is called if the trust was present it would the trust from the array, update the data and append it on the end of the array which would result in the display order change. This has been altered so the trust is updated in place on the array. 

### Work checklist

### Merge instructions

```
```

[ROE-1566]: https://companieshouse.atlassian.net/browse/ROE-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/controllers/add.trust.controller.ts
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    logger.debugRequest(req, `POST ${config.ADD_TRUST_PAGE}`);
+    const addNewTrust = req.body["addTrust"];
+
+    if (addNewTrust === '1') {
+      const appData = getApplicationData(req.session);
+      const newTrustId = generateTrustId(appData);
+
+      return res.redirect(`${config.TRUST_DETAILS_URL}/${newTrustId}`);
+    } else if (addNewTrust === '0') {
```
SteveHicksCH, line 66: Should we just gave an else clause since there are no more options

SteveHicksCH (commented):
Just one minor comment

Ignored the navigation since that will be the subject of a new Jira

SteveHicksCH (approved):

PR #694, opened 2023-02-22
Roe 1987 due diligence date error focus

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1987

### Change description

Need to use individual field validation to get the the required click field focus. This is followed by setting further validation to check for multiple empty fields and proceed only if they have not been found. If they have been found we just pass over the validation and return true, as no further validation will be required.

### Work checklist

- [x] Tests added where applicable
- [x] UI changes meet accessibility criteria

### Merge instructions

We are committed to keeping commit history clean, consistent and linear. To achieve this, this commit should be structured as follows:

```
<type>[optional scope]: <description>
```

and contain the following structural elements:

- fix: a commit that patches a bug in your codebase (this correlates with PATCH in semantic versioning),
- feat: a commit that introduces a new feature to the codebase (this correlates with MINOR in semantic versioning),
- BREAKING CHANGE: a commit that has a footer `BREAKING CHANGE:` introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type,
- types other than `fix:` and `feat:` are allowed, for example `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`, and others,
- footers other than `BREAKING CHANGE: <description>` may be provided.

PR #705, opened 2023-02-28
ROE-1938 - TL - Gov BO et al - Focus required for empty date error

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-1938

### Change description

Mainly refactoring: 

Moved empty date check to the day field validation to provide the click and focus required.
Replaced existing check in the test for the whole date with a simple true or false check and appended it to the other checks.

Also noticed here that a condition was being repeated for start dates for the work done on ROE-1987: https://github.com/companieshouse/overseas-entities-web/pull/694/files

 if (checkMoreThanOneDateFieldIsNotMissing(dayStr, monthStr, yearStr) && isYearEitherMissingOrCorrectLength(yearStr)) {
    if (isYearEitherMissingOrCorrectLength(yearStr)) {

so removed the unnecessary extra check.

As the changes affected the optional date and date of birth so have had to update these as well to be consistent with the start date.

We now have separate day field checks for date of birth, start date and optional/identity date. Moreover, the month and year checks need another parameter as dob has a different message.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -29,13 +29,6 @@ export const checkInvalidCharactersIfRadioButtonSelected = (selected: boolean, e
   return true;
 };
 
-export const checkDateIsNotCompletelyEmptyWithErrorThrown = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
-  if ( !day.trim() && !month.trim() && !year.trim() ) {
-    throw new Error(errMsg);
-  }
-  return true;
-};
-
 export const checkDateIsNotCompletelyEmpty = (day: string = "", month: string = "", year: string = "") => {
   if ( !day.trim() && !month.trim() && !year.trim() ) {
     return false;
@@ -161,7 +154,9 @@ export const checkDateFieldDay = (dayStr: string = "", monthStr: string = "", ye
     } else if (dayStr === "" && monthStr !== "" && yearStr === "") {
       throw new Error(ErrorMessages.DAY_AND_YEAR);
     } else {
-      checkDateIsNotCompletelyEmptyWithErrorThrown(ErrorMessages.ENTER_DATE, dayStr, monthStr, yearStr);
+      if (!checkDateIsNotCompletelyEmpty(dayStr, monthStr, yearStr)) {
+        throw new Error(ErrorMessages.ENTER_DATE);
+      }
     }
   }
   return true;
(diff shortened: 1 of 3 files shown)


SVarallo (approved):
Did some testing on the branch and seemingly fixes https://companieshouse.atlassian.net/browse/ROE-1936 as well. Looks all good to me but would recommend getting a second opinion

src/validation/custom.validation.ts
```diff
@@ -29,13 +29,20 @@ export const checkInvalidCharactersIfRadioButtonSelected = (selected: boolean, e
   return true;
 };
 
-export const checkDateIsNotCompletelyEmpty = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
+export const checkDateIsNotCompletelyEmptyWithErrorThrown = (errMsg: string, day: string = "", month: string = "", year: string = "") => {
```
mwestacott, line 32: I think now having two functions with the same logic in apart from either throwing an error or returning false is not great. Better to have just the one method (keep the one below) which returns true or false and in those three (I think) cases where it's currently called and needs to throw the error, react to the 'false' value returned and do that there.

src/validation/custom.validation.ts
```diff
@@ -151,25 +152,55 @@ const checkStartDateFields = (dayStr: string = "", monthStr: string = "", yearSt
...
+export const checkDateFieldDayOfBirth = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
+  if (isYearEitherMissingOrCorrectLength(yearStr)) {
+    if (dayStr === "" && monthStr !== "" && yearStr !== "") {
+      throw new Error(ErrorMessages.DAY_OF_BIRTH);
+    } else if (dayStr === "" && monthStr === "" && yearStr !== "") {
+      throw new Error(ErrorMessages.DAY_AND_MONTH_OF_BIRTH);
+    } else if (dayStr === "" && monthStr !== "" && yearStr === "") {
+      throw new Error(ErrorMessages.DAY_AND_YEAR_OF_BIRTH);
+    } else {
+      checkDateIsNotCompletelyEmptyWithErrorThrown(ErrorMessages.ENTER_DATE_OF_BIRTH, dayStr, monthStr, yearStr);
```
mwestacott, line 179: See comment above. So here you'd have to check for 'false' being returned and if so throw the error...

src/validation/custom.validation.ts
```diff
@@ -151,25 +152,55 @@ const checkStartDateFields = (dayStr: string = "", monthStr: string = "", yearSt
...
+    } else if (dayStr === "" && monthStr === "" && yearStr !== "") {
+      throw new Error(ErrorMessages.DAY_AND_MONTH);
+    } else if (dayStr === "" && monthStr !== "" && yearStr === "") {
+      throw new Error(ErrorMessages.DAY_AND_YEAR);
+    }
+  }
+  return true;
+};
+
+export const checkDateFieldMonth = (monthMissingMessage: string, monthYearMsg: string, dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
```
mwestacott, line 198: Would make the parameter naming consistent here - so `monthMissingMessage` and `monthYearMissingMessage`.

src/validation/custom.validation.ts
mattch1: Removed second function

src/validation/custom.validation.ts
mattch1: Removed second function

src/validation/custom.validation.ts
mattch1: Renamed

mwestacott (approved):

PR #717, opened 2023-03-03
Bug/roe 2018 roe 1507 fix duplication of trust

mouhajer-ch:
### JIRA link

Resolves [ROE-1507](https://companieshouse.atlassian.net/browse/ROE-1507)

### Change description

- Fix duplication of trust

### Work checklist

[ROE-1507]: https://companieshouse.atlassian.net/browse/ROE-1507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/views/trust-information.html
+++ b/views/trust-information.html
@@ -100,7 +100,7 @@
           }
         }) }}
 
-        {% if trustsAddedSoFar > 0 or trusts.length > 0 %}
+        {% if trustsAddedSoFar > 0 or (not errors and trusts.length) %}
           <div class="govuk-button-group">
              {{ govukButton({
               html: noMoreToAddHtml,


mattch1:
Committed a fix for:

1) validation error message for bad json

2) count of trusts seems to be the length of string when there is an error.

views/trust-information.html
```diff
@@ -16,7 +16,11 @@
 
 {% set trustAdded %}
   {% if errors and trustsAddedSoFar %}
-    ({{trustsAddedSoFar.length}} Trust information added so far)
+    {% if trustsAddedSoFar.length > 0 %}
+        ({{trustsAddedSoFar.length}} Trust information added so far)
```
mwestacott, line 20: And this output is the other part that needs checking with UX.

src/controllers/trust.information.controller.ts
```diff
@@ -33,6 +33,10 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
   try {
     logger.debugRequest(req, `POST ${config.TRUST_INFO_PAGE}`);
 
+    if (req.body.submit) {
```
markpit, line 36: looks like moving the submit redirect here means that it will ignore json posted in the json field on teh page and won't add it to the model. This is different behaviour from what was there previously, it might be ok, we'll have to check with the PO tomorrow in the office. It also allows us to move to the check your answers screen with 0 trusts added, so perhaps we should only allow the user to move forward if there is at least one trust in the model/session data. Can discuss in office tomorrow.

src/validation/custom.validation.ts
```diff
@@ -452,22 +454,55 @@ export const checkAtLeastOneFieldHasValue = (errMsg: string, ...fields: any[]) =
   throw new Error(errMsg);
 };
 
-export const checkTrustFields = (trustsJson: string) => {
+const isSubmitTrustSelected = (session: Session, action: string = ""): boolean => {
+  const appData: ApplicationData = getApplicationData(session);
+  return action === "submit" && (appData.trusts || [] ).length > 0;
+};
+
+export const checkTrustBO = (req) => {
```
mwestacott, line 462: I think method name should be `checkTrustBOs`. It's called from the validator and is linked to the "beneficialOwners" field.

mattch1:
Updated to hide submit button when 0 trusts are added.

PR #745, opened 2023-03-10
Feature/roe 2032 individual trustee change links

JoshFen95:
### JIRA link
[ROE-2032](https://companieshouse.atlassian.net/browse/ROE-2032)

### Change description

- update change links to they link to correct data fields on web page
- add map to page mapper method and use it to map formdata in the get method of controller. This was required because change links were linking to empty page without previous data being displayed
- add/update relevant unit tests

### Work checklist

### Merge instructions

```
```

[ROE-2032]: https://companieshouse.atlassian.net/browse/ROE-2032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -187,7 +187,7 @@ export const UPDATE_MANAGING_OFFICER_CORPORATE_URL = UPDATE_AN_OVERSEAS_ENTITY_U
 // URL PARAMS
 export const ROUTE_PARAM_TRUST_ID = "trustId";
 export const ROUTE_PARAM_BENEFICIAL_OWNER_ID = "boId";
-export const ROUTE_PARAM_TRUSTEE_ID = "trusteeID";
+export const ROUTE_PARAM_TRUSTEE_ID = "trusteeId";
 export const ID = "/:id"; // Same param name as the ID in the data model
 export const BO_ID = `/:${ROUTE_PARAM_BENEFICIAL_OWNER_ID}`;
 export const TRUST_ID = `/:${ROUTE_PARAM_TRUST_ID}`;
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -187,9 +187,11 @@ export const UPDATE_MANAGING_OFFICER_CORPORATE_URL = UPDATE_AN_OVERSEAS_ENTITY_U
 // URL PARAMS
 export const ROUTE_PARAM_TRUST_ID = "trustId";
 export const ROUTE_PARAM_BENEFICIAL_OWNER_ID = "boId";
+export const ROUTE_PARAM_TRUSTEE_ID = "trusteeID";
 export const ID = "/:id"; // Same param name as the ID in the data model
 export const BO_ID = `/:${ROUTE_PARAM_BENEFICIAL_OWNER_ID}`;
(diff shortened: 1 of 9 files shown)


src/config/index.ts
```diff
@@ -239,7 +239,11 @@ export const CONTACT_EMAIL = "#contact_email";
 export const CREATED_DATE = "#createdDate-createdDateDay";
 export const INCORPORATION_COUNTRY = "#incorporation_country";
 export const ENTITY_NAME = "#entity_name";
-export const ROLE_WITHIN_TRUST = "#role_within_trust";
```
SteveHicksCH, line 242: Is there a reason for this change, the former assignment seems more descriptive

src/routes/index.ts
```diff
@@ -252,7 +252,7 @@ router
   .post(trustHistoricalbeneficialOwner.post);
 
 router
-  .route(config.TRUST_ENTRY_URL + config.TRUST_ID + config.TRUST_INDIVIDUAL_BENEFICIAL_OWNER_URL + config.ID + '?')
+  .route(config.TRUST_ENTRY_URL + config.TRUST_ID + config.TRUST_INDIVIDUAL_BENEFICIAL_OWNER_URL + config.BO_ID + '?')
```
SteveHicksCH, line 255: BO_ID suggests beneficial owner rather than trustee. Maybe TRUSTEE_ID is better if you want to use a different value from ID

src/config/index.ts
JoshFen95: This var wasn't being used anywhere currently and the value for `role_with_trust `on the web page is still `type` in the html so the change link needs to look for type. 
I agree with you though and didn't really like the change to much myself but thought it best for another story where we can maybe refactor the entire use of `type` for `role_within_trust`.

What do you think?

src/routes/index.ts
JoshFen95: Yeah good point, I'll create one

src/config/index.ts
SteveHicksCH: Fair point. We probably want to send Sarah these refactorings in groups for stories.

src/routes/index.ts
SteveHicksCH: Thanks

JoshFen95: done

src/config/index.ts
```diff
@@ -187,9 +187,11 @@ export const UPDATE_MANAGING_OFFICER_CORPORATE_URL = UPDATE_AN_OVERSEAS_ENTITY_U
 // URL PARAMS
 export const ROUTE_PARAM_TRUST_ID = "trustId";
 export const ROUTE_PARAM_BENEFICIAL_OWNER_ID = "boId";
+export const ROUTE_PARAM_TRUSTEE_ID = "trusteeID";
```
SteveHicksCH, line 190: Other Id's end in "Id" not "ID"

JoshFen95: my bad, should be sorted now

SteveHicksCH (approved):

PR #759, opened 2023-03-15
uar-127 fix bomo fetch

AndreKramer7654:
### JIRA link https://companieshouse.atlassian.net/browse/UAR-127

Bug fix for setting [] in session 

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/beneficial.owner.type.controller.ts
+++ b/src/controllers/update/beneficial.owner.type.controller.ts
@@ -1,5 +1,4 @@
 import { NextFunction, Request, Response } from "express";
-import { Session } from "@companieshouse/node-session-handler";
 import { logger } from "../../utils/logger";
 
 import * as config from "../../config";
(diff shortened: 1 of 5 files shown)


src/controllers/update/beneficial.owner.type.controller.ts
```diff
@@ -25,8 +22,9 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
     const appData: ApplicationData = getApplicationData(req.session);
 
     if (!hasFetchedBoAndMoData(appData)) {
-      await retrieveBeneficialOwners(req, appData);
-      await retrieveManagingOfficers(req, appData);
+      await getCompanyOfficers(req, appData[EntityNumberKey] as string);
```
olaych, line 25: This will still need to be replaced by map methods rather than await the service call itself?

corybudd-ch: yeah I agree with @olaych, I think keeping the methods retrieveManagingOfficers and retrieveBeneficialOwners is good as mapping stuff will need to add them back in anyways.

src/model/update.type.model.ts
```diff
@@ -3,6 +3,6 @@ export const UpdateKey = "update";
 export const UpdateKeys: string[] = [ "date_of_creation" ];
 
 export interface Update {
-    date_of_creation?: string
+    date_of_creation?: string,
+    fetchedBoMoData?: true
```
corybudd-ch, line 7: does it matter that keys are mixture of snake and camel case?

corybudd-ch (approved):
Approved as functionality all good just couple of comments

src/controllers/update/beneficial.owner.type.controller.ts
AndreKramer7654: Let's leave that to the mappings. I did not want two one line functions.

src/model/update.type.model.ts
AndreKramer7654: Not in this case as just for session but will align it.

olaych (approved):
Approved with existing issue to be carried forward

PR #774, opened 2023-03-21
feature/UAR-323-update-you-have-signed-out-page

azafarCH:
### Work checklist

### Merge instructions

```
```


test/controllers/update/update.signed.out.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+import { logger } from "../../../src/utils/logger";
+
+const mockAuthenticationMiddleware = authentication as jest.Mock;
+mockAuthenticationMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => next() );
+const mockServiceAvailabilityMiddleware = serviceAvailabilityMiddleware as jest.Mock;
+mockServiceAvailabilityMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => next() );
+
+const mockLoggerDebugRequest = logger.debugRequest as jest.Mock;
+
+describe("SIGNED OUT PAGE controller", () => {
```
sjahir, line 31: remove the word PAGE

sjahir (approved):

src/controllers/update/update.signed.out.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { logger } from "../../utils/logger";
+import * as config from "../../config";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `${req.method} ${req.route.path}`);
+
+    return res.render(config.UPDATE_SIGNED_OUT_PAGE, {
+      templateName: config.UPDATE_SIGNED_OUT_PAGE,
+      signedOut: true
```
AndreKramer7654, line 12: Is signedOut the right name here? showBanners seem more like the effect that you are after?

AndreKramer7654:
I guess you want to redirect to the sign in page. The authentication.middleware.ts redirects to "/signin". It also allows you to return to the journey start after sign in but maybe we should be 'returning' to the "your-filings" page after sign in?

src/controllers/update/update.signed.out.controller.ts
azafarCH: would it be more appropriate to have hideBanners, since it will be set to true in this controller?

src/controllers/update/update.signed.out.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+import { logger } from "../../utils/logger";
+import * as config from "../../config";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `${req.method} ${req.route.path}`);
+
+    return res.render(config.UPDATE_SIGNED_OUT_PAGE, {
+      templateName: config.UPDATE_SIGNED_OUT_PAGE,
+      hideBanners: true
```
AndreKramer7654, line 12: Are we the only renderer of this page so do we need to pass in hideBanners or just do it in view?

AndreKramer7654: might result in most of the template copied.

src/controllers/update/update.sign.out.controller.ts
```diff
@@ -26,7 +26,7 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
       throw createAndLogErrorRequest(req, `${previousPage} page is not part of the journey!`);
     }
     if (req.body["sign_out"] === 'yes') {
-      return res.redirect(config.UPDATE_ACCOUNTS_SIGN_OUT_URL);
+      return res.redirect(config.UPDATE_SIGNED_OUT_URL);
```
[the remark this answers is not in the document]
AndreKramer7654: The only way I can see this working is if we can say something like:
return res.redirect(config.UPDATE_ACCOUNTS_SIGN_OUT_URL + "?return_to=" + config.UPDATE_SIGNED_OUT_URL);

PR #804, opened 2023-03-24
PA roe-1599 fixing name bugs in view

okoye-companieshouse:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-1599

### Change description

Made name and id attribute for elements legal form end governing law same

### DIFF
--- a/views/trust-legal-entity-beneficial-owner.html
+++ b/views/trust-legal-entity-beneficial-owner.html
@@ -106,22 +106,20 @@
         {% include "includes/inputs/address/principal-address-input.html" %}
 
         {% set fieldParam = {
-          name: 'legal_form',
-          id: 'legal_form',
+          name: 'legalForm',
           label: 'What is its legal form?',
           value: formData.legalForm,
           hint: 'For example, limited company',
-          error: errors.legal_form if errors
+          error: errors.legalForm if errors
         } %}
         {% include "includes/inputs/text-input.html" %}
 
         {% set fieldParam = {
-          name: 'law_governed',
-          id: 'law_governed',
+          name: 'governingLaw',
           label: 'What is the governing law?',
           value: formData.governingLaw,
           hint: 'This is the law that the entity operates under.',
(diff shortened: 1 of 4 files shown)


views/trust-legal-entity-beneficial-owner.html
```diff
@@ -107,7 +107,7 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ pageParams.title }}</h1>
 
         {% set fieldParam = {
           name: 'legalForm',
-          id: 'legal_form',
+          id: 'legalForm',
```
JoshFen95, line 110: I think these renamings will cause issues with the legal entity change links on the check-answers-page. They will stop the change links linking the correct fields. May need to update them if you need to do these renamings. 

Also, if you are renaming the `id` to the same value as the `name` then you don't need to specify the `id`. If the `id` field is not set then it is set to the value of the `name` field. See inside `includes/inputs/text-input.html`

views/trust-legal-entity-beneficial-owner.html
```diff
@@ -117,7 +117,7 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ pageParams.title }}</h1>
 
         {% set fieldParam = {
           name: 'governingLaw',
-          id: 'law_governed',
```
JoshFen95, line 120: I think these renamings will cause issues with the legal entity change links on the check-answers-page. They will stop the change links linking the correct fields. May need to update them if you need to do these renamings. 

Also, if you are renaming the `id` to the same value as the `name` then you don't need to specify the `id`. If the `id` field is not set then it is set to the value of the `name` field. See inside `includes/inputs/text-input.html`

okoye-companieshouse: I also fixed the check-answers-page as well. I think it should be fine now

JoshFen95 (approved):

PR #809, opened 2023-03-27
Correct Trust mappings on a resume

SteveHicksCH:
### JIRA link
[ROE-2086](https://companieshouse.atlassian.net/browse/ROE-2086)

### Change description
The API returns data in a different type of resource as is sent up to it! This is hard coded in the API with the Node SDK just doing some date mapping.

 Rather than re-writing the Node SDK and the input part of the API I coded it in a function that is only called when the Trust Screens feature flag is true.

### Work checklist

- x ] Tests added where applicable

### Merge instructions

```
```

[ROE-2086]: https://companieshouse.atlassian.net/browse/ROE-2086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/controllers/resume.submission.controller.ts
+++ b/src/controllers/resume.submission.controller.ts
@@ -20,7 +20,7 @@ import { ManagingOfficerCorporate, ManagingOfficerCorporateKey } from "../model/
 import { ManagingOfficerIndividual, ManagingOfficerKey } from "../model/managing.officer.model";
 import { startPaymentsSession } from "../service/payment.service";
 import { getTransaction } from "../service/transaction.service";
-import { generateTrusteeIds } from "../utils/trusts";
+import { generateTrusteeIds, mapTrusteeAddressesFromApiToWebModel } from "../utils/trusts";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
@@ -98,7 +98,13 @@ const setWebApplicationData = (session: Session, appData: ApplicationData, trans
     appData[WhoIsRegisteringKey] = WhoIsRegisteringType.AGENT;
   }
 
-  generateTrusteeIds(appData);
+  if (isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUSTS_WEB)) {
+
+    generateTrusteeIds(appData);
+    mapTrusteeAddressesFromApiToWebModel(appData);
+  }
 
   setExtraData(session, appData);
+
+  logger.debug(`resume app data ${JSON.stringify(appData)}`);
 };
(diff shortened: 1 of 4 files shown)


src/controllers/resume.submission.controller.ts
```diff
@@ -98,7 +98,10 @@ const setWebApplicationData = (session: Session, appData: ApplicationData, trans
     appData[WhoIsRegisteringKey] = WhoIsRegisteringType.AGENT;
   }
 
-  generateTrusteeIds(appData);
+  if (isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUSTS_WEB)) {
```
markpit, line 101: we'd need to keep the generateTrustIds so it gets called if we switch the feature flag off, so have on/off = new/old

SteveHicksCH: We only need the trustee ids when we have the CHS Screens since they are used to support this. The spreadsheet solution does NOT need them (and they are not in the API Model)

markpit: ah ok, i thought it was something added for the spreadsheets, i can see it was added recently by looking at the git blame.

src/utils/trusts.ts
```diff
@@ -318,25 +318,134 @@ const saveIndividualTrusteeInTrust = (trust: Trust, trusteeData: IndividualTrust
...
+        return {
+          id: uuidv4(),
+          type: getRoleWithinTrustType(apiData.type) as RoleWithinTrustType,
+          name: apiData.name,
+          date_became_interested_person_day: apiData?.date_became_interested_person_day,
+          date_became_interested_person_month: apiData?.date_became_interested_person_month,
+          date_became_interested_person_year: apiData?.date_became_interested_person_year,
+          ro_address_premises: apiData.registered_office_address.property_name_number,
+          ro_address_line_1: apiData.registered_office_address.line_1,
+          ro_address_line_2: apiData?.registered_office_address.line_2,
```
markpit, line 358: noticed that some of the apiData have the null check '?', just wondering if they should all be the same

SteveHicksCH: I have copied the null check '?' from the web models in src/model/trust.model.ts (TrustCorporate etc)

okoye-companieshouse (approved):

src/utils/trusts.ts
markpit: i think it would be safer to use the ? on all the fields during the mapping, to prevent null pointer errors

PR #833, opened 2023-04-05
*PA*roe-2132: Add function to convert api corporate_Indicator

JoshFen95:
### JIRA link

[ROE-2132](https://companieshouse.atlassian.net/browse/ROE-2132)

### Change description

- The web uses a yesNoResponse which reflects a yes or no choice using 1 and 0. The api is returing the corporate_indicator as true or false which makes checking it more difficult.
- This private function is used to convert the api corporate_indicator into a 1 or 0 so that the web app understands it. This function is called when data is being populated to the session from a save and resumed application

### Work checklist

### Merge instructions

```
```

[ROE-2132]: https://companieshouse.atlassian.net/browse/ROE-2132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/trusts.ts
+++ b/src/utils/trusts.ts
@@ -13,6 +13,7 @@ import {
   TrustKey,
   TrustCorporate,
 } from "../model/trust.model";
+import { yesNoResponse } from "../model/data.types.model";
 
 /**
  * Checks whether any beneficial owners requires trust data due to at least one of them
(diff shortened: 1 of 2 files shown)


src/utils/trusts.ts
```diff
@@ -446,6 +445,16 @@ function getRoleWithinTrustType(type: any): RoleWithinTrustType | undefined {
...
 }
 
+function convertBooleanToYesNoResponse(apiYesNoResponse: yesNoResponse): yesNoResponse {
+  // used to convert boolean recieved from api to number for yesNoResponse used in web.
+  // although true/false is sent by the api, typescript still detects it as a yesNoResponse hence why yesNoResponse is passed to this function
+
+  if (apiYesNoResponse === undefined){
+    return yesNoResponse.No;
+  }
+  return Number(apiYesNoResponse);
```
okoye-companieshouse, line 455: parseInt may be better here as you are expecting an int. Nonetheless will still work.

JoshFen95: vscode complains when using parseInt in this case

PR #893, opened 2023-05-17
UAR-473: read nationalities from comma separated list

mmcpolin-ch:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-473

### Change description
Nationality information for an MOs ‘nationality’ and an MOs ‘additional nationality’ is stored as a comma separated list in CHIPs. 
We need to a solution to ensure that the second nationality listed in the comma separated list, appears separately on the page as a second item. This solution will assist the ROE Agent in entering, checking and amending the MO ‘nationality’ and ‘additional nationality’ data.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/update/managing.officer.mapper.ts
+++ b/src/utils/update/managing.officer.mapper.ts
@@ -9,6 +9,7 @@ export const mapToManagingOfficer = (officer: CompanyOfficer): ManagingOfficerIn
   const service_address = mapBOMOAddress(officer.address);
   const address = undefined;
   const names = splitNames(officer.name);
+  const nationalities = splitNationalities(officer.nationality);
   const formernames = getFormerNames(officer.formerNames);
 
   return {
@@ -18,7 +19,8 @@ export const mapToManagingOfficer = (officer: CompanyOfficer): ManagingOfficerIn
     has_former_names: officer.formerNames ? yesNoResponse.Yes : yesNoResponse.No,
     former_names: formernames,
     date_of_birth: mapDateOfBirth(officer.dateOfBirth),
-    nationality: officer.nationality,
+    nationality: nationalities[0],
+    second_nationality: nationalities[1],
     usual_residential_address: address,
     is_service_address_same_as_usual_residential_address: isSameAddress(service_address, address) ? yesNoResponse.Yes : yesNoResponse.No,
     service_address: service_address,
@@ -67,6 +69,13 @@ export const splitNames = (officerName: string): string[] => {
       return names;
     }
   }
+};
+
+export const splitNationalities = (officerNationalities: string | undefined): string[] => {
+  if (!officerNationalities) {
+    return [""];
+  }
+  return officerNationalities.split(/\s*,\s*/).slice(0, 2);
 };
 
 export const getFormerNames = (formerNames?: FormerNameResource[]): string => {
--- a/src/utils/update/managing.officer.mapper.ts
+++ b/src/utils/update/managing.officer.mapper.ts
@@ -64,7 +64,7 @@ export const splitNames = (officerName: string): string[] => {
           firstNames += " ";
         }
       }
-      return [firstNames, names[names.length - 1]];
+      return [firstNames.trim(), names[names.length - 1].trim()];
     } else {
       return names;
     }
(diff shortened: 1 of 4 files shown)


jkayyala (approved):

src/utils/update/managing.officer.mapper.ts
```diff
@@ -18,7 +19,8 @@ export const mapToManagingOfficer = (officer: CompanyOfficer): ManagingOfficerIn
     has_former_names: officer.formerNames ? yesNoResponse.Yes : yesNoResponse.No,
     former_names: formernames,
     date_of_birth: mapDateOfBirth(officer.dateOfBirth),
-    nationality: officer.nationality,
+    nationality: nationalities[0],
+    second_nationality: nationalities[1],
```
AndreKramer7654, line 23: Maybe be explicit about the case when there is only one element in list
second_nationality: nationalities.length < 2? undefined: nationalities[2]

src/utils/update/managing.officer.mapper.ts
```diff
@@ -69,6 +71,26 @@ export const splitNames = (officerName: string): string[] => {
   }
 };
 
+export const splitNationalities = (officerNationalities: string | undefined): string[] => {
+  if (officerNationalities === undefined) {
+    return ["", ""];
+  } else {
+    const nationalities = officerNationalities.split(",");
```
AndreKramer7654, line 78: just 
if (!nationalities) return [""];
return nationalities.length > 1?  nationalities.slice(0, 2) else nationalities;

src/utils/update/managing.officer.mapper.ts
mmcpolin-ch: if nationalities[1] doesn't exist then it will return undefined anyway, so this isn't necessary

src/utils/update/managing.officer.mapper.ts
mmcpolin-ch: This won't work as it returns a string as opposed to a list of strings

src/utils/update/managing.officer.mapper.ts
AndreKramer7654: Yes, but might help to be explicit. Can also be ["",""] so we have undefined or "" being no second nationality.

src/utils/update/managing.officer.mapper.ts
AndreKramer7654: Sorry? what returns a string?

azafarCH (approved):

src/utils/update/managing.officer.mapper.ts
AndreKramer7654: The task is to split a "," separated string and use the first item for "nationality" and second item for "second nationaily" and ignore any others. So just:

const  firstAndSecondN = "american, french".split(",").slice(0, 2);
const firstN = firstAndSecondN[0]=== undefined? "": firstAndSecondN[0].trim();
const secondN = firstAndSecondN[0]  === undefined ? "": firstAndSecondN[1].trim();

src/utils/update/managing.officer.mapper.ts
```diff
@@ -69,6 +71,13 @@ export const splitNames = (officerName: string): string[] => {
   }
 };
 
+export const splitNationalities = (officerNationalities: string | undefined): string[] => {
+  if (!officerNationalities) {
+    return [ "" ];
+  }
+  return officerNationalities.length > 1 ? officerNationalities.split(",") : [officerNationalities];
```
AndreKramer7654, line 78: Not sure why you would special case if the officersNationalities is longer than 1 character? Do you have a test for that not being the case? Or just drop that test?

AndreKramer7654: Do you need to trim values if it "country1, country2"?

test/utils/update/managing.officer.mapper.spec.ts
```diff
@@ -34,7 +34,7 @@ describe("Test mapping to managing officer", () => {
         month: managingOfficerMock.dateOfBirth?.month,
         year: managingOfficerMock.dateOfBirth?.year
       },
-      nationality: managingOfficerMock.nationality,
+      nationality: splitNationalities(managingOfficerMock.nationality)[0],
```
AndreKramer7654, line 37: Don't use the function you are testing in the test mocks or expects.

AndreKramer7654: just use the mock county1 and country2 as strings.

src/utils/update/managing.officer.mapper.ts
```diff
@@ -62,13 +64,20 @@ export const splitNames = (officerName: string): string[] => {
           firstNames += " ";
         }
       }
-      return [firstNames, names[names.length - 1]];
+      return [firstNames.trim(), names[names.length - 1].trim()];
```
AndreKramer7654, line 67: good catch

AndreKramer7654 (approved):

PR #904, opened 2023-05-26
UAR-286: Add update journey to "page not found" page

stephanie-ch:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-268

### Work checklist

- [ ] Tests added where applicable
- [ ] UI changes meet accessibility criteria

### Merge instructions

We are committed to keeping commit history clean, consistent and linear. To achieve this, this commit should be structured as follows:

```
<type>[optional scope]: <description>
```

and contain the following structural elements:

- fix: a commit that patches a bug in your codebase (this correlates with PATCH in semantic versioning),
- feat: a commit that introduces a new feature to the codebase (this correlates with MINOR in semantic versioning),
- BREAKING CHANGE: a commit that has a footer `BREAKING CHANGE:` introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type,
- types other than `fix:` and `feat:` are allowed, for example `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`, and others,
- footers other than `BREAKING CHANGE: <description>` may be provided.

PR #932, opened 2023-06-26
UAR-642: Change header content

jkayyala:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-642

### Change description
Changes for changing the header text and link and the feedback url for the survey

### Work checklist

### Merge instructions

```
```


src/controllers/error.controller.ts
```diff
@@ -3,7 +3,13 @@ import { logger } from "../utils/logger";
 import * as config from "../config";
 
 const pageNotFound = (req: Request, res: Response) => {
-  return res.status(404).render(config.NOT_FOUND_PAGE);
+  let isRegistration = true;
```
AndreKramer7654, line 6: const isRegistration : boolean = req.path.startsWith(config.UPDATE_LANDING_URL);

src/controllers/error.controller.ts
```diff
@@ -13,8 +19,13 @@ const pageNotFound = (req: Request, res: Response) => {
  */
 const errorHandler = (err: Error, req: Request, res: Response, _next: NextFunction) => {
   logger.errorRequest(req, `An error has occurred. Re-routing to the error screen - ${err.stack}`);
+  let isRegistration = true;
```
AndreKramer7654, line 22: as previous

src/utils/beneficial.owner.statements.ts
```diff
@@ -30,7 +30,8 @@ export const getBeneficialOwnerStatements = (req: Request, res: Response, next:
       backLinkUrl: BACK_LINK,
       templateName: templateName,
       [BeneficialOwnerStatementKey]: appData[BeneficialOwnerStatementKey],
-      noChangeFlag
+      noChangeFlag,
+      ...appData
```
AndreKramer7654, line 34: if we can avoid passing all the appData in that would be good

views/page-not-found.html
```diff
@@ -1,4 +1,14 @@
-{% extends "layout.html" %}
+{% if isRegistration %}
```
AndreKramer7654, line 1: There is a blocked ticket in K1 to page not found: https://github.com/companieshouse/overseas-entities-web/pull/904 I'm happy for this change here.

src/controllers/error.controller.ts
jkayyala: @AndreKramer7654 , can change it `let isRegistration : boolean = req.path.startsWith(config.UPDATE_LANDING_URL);` as I change the value later and with a const its not possible

views/update/update-layout.html
```diff
@@ -0,0 +1,95 @@
+{% extends "govuk/template.njk" %}
```
AndreKramer7654, line 1: duplication with "layout.html" here but I think it's ok to duplicate in order to keep it simpler to diverge.

jkayyala: yeah , tried avoiding it but its harder to do without it , else it might impact the Reg journey

test/__mocks__/session.mock.ts
```diff
@@ -1494,6 +1494,7 @@ export const APPLICATION_DATA_MOCK: ApplicationData = {
 
 export const APPLICATION_DATA_MOCK_WITHOUT_UPDATE: ApplicationData = {
   ...APPLICATION_DATA_MOCK,
+  [EntityNumberKey]: undefined,
```
corybudd-ch, line 1497: waas there a reason you added this? Is there a scenario where without the key something breaks?
Because I don't think in register journey this key will be in the AppData with undefined like this

corybudd-ch: Hmm but if without this it displays the wrong template would that not mean wrong template would be displayed in register journey? During register journey `EntityNumberKey` will not be in AppData.

corybudd-ch: Have you tested register journey?

jkayyala: yes tested the Reg journey , EntityNumberKey maps to entity_number and it will always be present as its defined in ApplicationData in /model/application.model as we share the same model in Registration and Update

corybudd-ch: entity_number key is optional and doesn't get set or exist in ApplicationData during a register journey so I would make your test data here reflect that and not set entity_number = undefined. Then fix whatever potential issue there was around that.

AndreKramer7654: Isn't that the same as setting to undefined though? Depends on what you think is clearer.

corybudd-ch: Not exactly from an object point of view, Object.keys would contain `EntityNumberKey` if undefined. It wouldn't if we don't define/add the key at all. 
Which was just all my original comment was, adding `[EntityNumberKey]: undefined` to a mock shouldn't be necessary here and should just use a mock without EntityNumberKey.

corybudd-ch: Which looks like @jkayyala has done now by adding APPLICATION_DATA_REGISTRATION_MOCK. Do you still need to add this line to APPLICATION_DATA_MOCK_WITHOUT_UPDATE then @jkayyala or can it be removed?

jkayyala: @corybudd-ch , added it as it was clear both(undefined and not having the key at all) works in the same way and hence added the test , but it can be removed if needed

corybudd-ch (approved):

AndreKramer7654 (approved):

views/includes/update_phase_banner.html
```diff
@@ -0,0 +1,6 @@
+{{ govukPhaseBanner({
+  tag: {
+    text: "beta"
+  },
+  html: 'This is a new service - your <a class="govuk-link" href="https://www.smartsurvey.co.uk/s/oe-update-fdbk/">feedback</a> will help us to improve it.'
```
stephanie-ch, line 5: Need a data event id for matamo?

stephanie-ch:
@jkayyala @AndreKramer7654 are we saying we're happy with the duplication for now - and we'll add it to be address in the Sonar ticket coming up?

stephanie-ch (approved):

PR #946, opened 2023-06-29
Fix/uar 775 address displaying incorrectly

cdrainCH:
### JIRA link

https://companieshouse.atlassian.net/jira/software/c/projects/UAR/boards/531?modal=detail&selectedIssue=UAR-775&sprint=2901

### Change description

Deselects radio button for 'Is correspondence address same as principle office address' and does not populate the fields for correspondence address. As principle address cannot be pulled in at this time, this is a temporary workaround.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/update.review.managing.officer.corporate.controller.ts
+++ b/src/controllers/update/update.review.managing.officer.corporate.controller.ts
@@ -27,7 +27,13 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     if (appData?.managing_officers_corporate){
       dataToReview = appData?.managing_officers_corporate[Number(index)];
       principalAddress = (dataToReview) ? mapDataObjectToFields(dataToReview[PrincipalAddressKey], PrincipalAddressKeys, AddressKeys) : {};
-      serviceAddress = (dataToReview) ? mapDataObjectToFields(dataToReview[ServiceAddressKey], ServiceAddressKeys, AddressKeys) : {};
+      let undefinedPrincipalAddressProperties;
+      if (principalAddress !== undefined) {
+        undefinedPrincipalAddressProperties = Object.values(principalAddress).every(value => value === undefined);
+      }
+      if (!undefinedPrincipalAddressProperties) {
+        serviceAddress = (dataToReview) ? mapDataObjectToFields(dataToReview[ServiceAddressKey], ServiceAddressKeys, AddressKeys) : {};
+      }
     }
 
     const templateOptions = {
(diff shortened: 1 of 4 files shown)


src/controllers/update/update.review.managing.officer.corporate.controller.ts
```diff
@@ -27,7 +27,13 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     if (appData?.managing_officers_corporate){
       dataToReview = appData?.managing_officers_corporate[Number(index)];
       principalAddress = (dataToReview) ? mapDataObjectToFields(dataToReview[PrincipalAddressKey], PrincipalAddressKeys, AddressKeys) : {};
-      serviceAddress = (dataToReview) ? mapDataObjectToFields(dataToReview[ServiceAddressKey], ServiceAddressKeys, AddressKeys) : {};
+      let undefinedPrincipalAddressProperties;
```
corybudd-ch, line 30: is the code across the 3 files exactly the same? Might be better extracting this out into a method to call from the 3 places?

cdrainCH: Yeah it is the same, will do

corybudd-ch (changes requested):
Think we need to try doing this at mapper level and avoid controller and HTML changes. Will give it a go with you

src/utils/update/managing.officer.mapper.ts
```diff
@@ -31,15 +31,15 @@ export const mapToManagingOfficer = (officer: CompanyOfficer): ManagingOfficerIn
 };
 
 export const mapToManagingOfficerCorporate = (officer: CompanyOfficer): ManagingOfficerCorporate => {
-  const service_address = mapBOMOAddress(officer.address);
+  const service_address = mapBOMOAddress(undefined);
```
AndreKramer7654, line 34: We'll have no addresses displayed for corporate managing officer if we set both service_address and principle_address to undefined?

corybudd-ch: Yeah I also didn't understand we'd do this and asked @cdrainCH to clarify with Matthew Poland what the expectation is here ..

cdrainCH: So early on Friday Matthew thought that this was the right path to take, but after a chat with Stephanie he realised that this wasn't necessary so this will be reverted and the address will be mapped to the service address again

stephanie-ch: Yeah, Matthew and I ran through this Friday evening. The only change required is not selecting the yes/no radio button. Let's please remember to do 3 amigos when picking up a ticket as it would have saved some time and confusion here :)

stephanie-ch (approved):

joshuahigginsonCH (approved):

src/utils/update/managing.officer.mapper.ts
```diff
@@ -39,7 +39,7 @@ export const mapToManagingOfficerCorporate = (officer: CompanyOfficer): Managing
     ch_reference: mapSelfLink(officer.links?.self),
     name: officer.name,
     principal_address: address,
-    is_service_address_same_as_principal_address: isSameAddress(service_address, address) ? yesNoResponse.Yes : yesNoResponse.No,
+    is_service_address_same_as_principal_address: undefined,
```
AndreKramer7654, line 42: if we wanted to future proof it we could do:
 isSameAddress(service_address, address) ? yesNoResponse.Yes : undefined
But ok for now.

AndreKramer7654 (approved):

PR #948, opened 2023-06-29
[UAR-803] update mapping for is_on_register_in_country_formed_in

corybudd-ch:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-803

### Change description
After QA completeing a registration of an OE and then using this OE in the update journey, the `is_on_register_in_country_formed_in` field isn't populated into the company_profiles collection in the MongoDB even though it is part of the ApplicationData model and I can see it being added to it in the registration journey. 

The field is also defined against the CompanyProfile model in the CH api sdk, so the `getCompanyProfile` service in the SDK will automatically set `is_on_register_in_country_formed_in = false` if it does not get it back in the API response.

```
const isOnRegisterInCountryFormedIn = (body.is_on_register_in_country_formed_in)
            ? body.is_on_register_in_country_formed_in === "true"
            : false;
```

I think CHIPS is dropping that information off whenever populating the MongoDB for whatever reason so as a solution here I've went with setting `is_on_register_in_country_formed_in` based on the existence of a `registration_number` instead of the field itself, as it seems to be possible at the minute for public_register information to be present without the `is_on_register_in_country_formed_in`.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/update/mapper.utils.ts
+++ b/src/utils/update/mapper.utils.ts
@@ -1,5 +1,5 @@
 import { Address, InputDate } from "../../model/data.types.model";
-import { OfficeAddress, ServiceAddress } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
+import { CompanyProfile, OfficeAddress, ServiceAddress } from "@companieshouse/api-sdk-node/dist/services/company-profile/types";
 import { Address as PSCAddress, DateOfBirth as PSCDateOfBirth } from "@companieshouse/api-sdk-node/dist/services/company-psc/types";
 import { Address as OfficerAddress, DateOfBirth as OfficerDateOfBirth } from "@companieshouse/api-sdk-node/dist/services/company-officers/types";
 import { padWithZero } from "../../controllers/update/update.review.beneficial.owner.individual";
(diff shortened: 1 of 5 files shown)


joshuahigginsonCH (approved):

src/utils/update/mapper.utils.ts
```diff
@@ -121,3 +121,13 @@ export const lowerCaseAllWordsExceptFirstLetters = (country: string | undefined)
   }
   );
 };
+
+export const hasPublicRegisterDetails = (cp: CompanyProfile): boolean => {
```
stephanie-ch, line 125: can you call it company profile please

src/utils/update/mapper.utils.ts
```diff
@@ -121,3 +121,13 @@ export const lowerCaseAllWordsExceptFirstLetters = (country: string | undefined)
...
+  if (cp.isOnRegisterInCountryFormedIn !== undefined){
+    return cp.isOnRegisterInCountryFormedIn;
+  } else if (cp.foreignCompanyDetails?.originatingRegistry?.name && cp.foreignCompanyDetails?.registrationNumber) {
```
stephanie-ch, line 128: Minor but as both fields are mandatory on registration if yes selected, you could probably just check for one of them.

stephanie-ch (approved):

AndreKramer7654 (approved):

stephanie-ch (approved):

PR #953, opened 2023-07-03
Bugfix/uar 585 add validation to continue button

bmcgreevyCH:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-585

### Change description
Added validation to radio buttons / continue button.

### DIFF
--- a/src/validation/error.messages.ts
+++ b/src/validation/error.messages.ts
@@ -182,7 +182,7 @@ export enum ErrorMessages {
   SELECT_IF_STILL_MANAGING_OFFICER = "Select yes if they are still a managing officer",
   SELECT_IF_ANY_TRUSTS_INVOLVED = "Select yes if there are any trusts involved in this overseas entity",
   SELECT_DO_YOU_WANT_TO_MAKE_OE_CHANGE = "Select yes if you need to make any changes to this overseas entity",
-  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if this information is correct",
+  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if all the information in this update statement is correct",
 
   // MAX Lengths
   MAX_FIRST_NAME_LENGTH = "First name must be 50 characters or less",
--- a/src/validation/error.messages.ts
+++ b/src/validation/error.messages.ts
@@ -182,6 +182,7 @@ export enum ErrorMessages {
   SELECT_IF_STILL_MANAGING_OFFICER = "Select yes if they are still a managing officer",
   SELECT_IF_ANY_TRUSTS_INVOLVED = "Select yes if there are any trusts involved in this overseas entity",
   SELECT_DO_YOU_WANT_TO_MAKE_OE_CHANGE = "Select yes if you need to make any changes to this overseas entity",
+  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if this information is correct",
 
   // MAX Lengths
   MAX_FIRST_NAME_LENGTH = "First name must be 50 characters or less",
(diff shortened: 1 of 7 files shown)


src/validation/error.messages.ts
```diff
@@ -182,6 +182,7 @@ export enum ErrorMessages {
   SELECT_IF_STILL_MANAGING_OFFICER = "Select yes if they are still a managing officer",
   SELECT_IF_ANY_TRUSTS_INVOLVED = "Select yes if there are any trusts involved in this overseas entity",
   SELECT_DO_YOU_WANT_TO_MAKE_OE_CHANGE = "Select yes if you need to make any changes to this overseas entity",
+  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if this information is correct",
```
n-r-b, line 185:
```suggestion
  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if all the information in this update statement is correct",
```

AndreKramer7654: Did the content come from the designers? Quick check with Sheena?

n-r-b: I've just taken it from the ticket, I can get this double checked though

bmcgreevyCH: I didn't even notice his comment. I've changed it to match.

test/controllers/update/update.review.statement.controller.spec.ts
```diff
@@ -156,7 +157,9 @@ describe("Update review overseas entity information controller tests", () => {
       mockIsActiveFeature.mockReturnValueOnce(true);
       mockGetApplicationData.mockReturnValueOnce(APPLICATION_DATA_UPDATE_BO_MOCK);
       mockPaymentsSession.mockReturnValueOnce(PAYMENT_LINK_JOURNEY);
-      const resp = await request(app).post(UPDATE_REVIEW_STATEMENT_URL);
+      const resp = await request(app)
+        .post(UPDATE_REVIEW_STATEMENT_URL)
+        .send({ no_change_review_statement: '1' });
```
n-r-b, line 162: The test name mentions 'no_change_review_statement' being undefined, so we either need to keep the test as is and potentially update the logic in the controller or update the name of the test?

bbell-companieshouse (approved):
Just one comment from Natalie left to address but other than that LGTM

AndreKramer7654 (approved):

views/update/review-update-statement.html
```diff
@@ -49,7 +49,22 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds-from-desktop">
 
-      <span class="govuk-caption-xl govuk-!-padding-bottom-1">{{ appData.entity_name }} - {{ appData.entity_number }}</span>
+      {% include "includes/list/errors.html" %}
+      
+      <span class="govuk-caption-xl govuk-!-padding-bottom-1">
+        {% if appData.entity_name %}
```
AndreKramer7654, line 55: Can we avoid these conditionals by change to always include entityName and entityNumber in the session so nver using appData.entity_name/number?

bmcgreevyCH: The first time the page loads it uses appData, then when an error is triggered the validation middleware spreads appData out and entity name and number can only be accessed through the variables directly. Because our current implementation iterates through BOs and MOs, it means that code then breaks as the data is spread. I'd need more time with it as it involves more investigation, but not worth holding the ticket back for given how close we are to the wire. This patch fix I made will ensure the ticket gets out in this sprint. By the looks of it a previous ticket has had to do the same thing a number of months back. I don't like it either but Its a weird one which I'll address with a bug fix after we get this ticket over the line.

AndreKramer7654: Ok for a bug fix but something to revisit.

test/controllers/update/update.review.statement.controller.spec.ts
bmcgreevyCH: Changed name (its outdated now), cheers!

n-r-b (approved):

PR #958, opened 2023-07-06
UAR 811 show trust NOC boxes

cdrainCH:
### JIRA link

https://companieshouse.atlassian.net/jira/software/c/projects/UAR/boards/531?modal=detail&selectedIssue=UAR-811&sprint=2913

### Change description

Showing the nature of control boxes for BO individual and BO other

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/views/includes/page-templates/beneficial-owner-nature-of-control-types.html
+++ b/views/includes/page-templates/beneficial-owner-nature-of-control-types.html
@@ -9,7 +9,7 @@
 {% set errorMessage = errors.beneficial_owner_nature_of_control_types if errors %}
 {% include "includes/inputs/nature-of-control-checkboxes/beneficial-owner-nature-of-control-types.html" %}
 
-{% if not isBeneficialOwnerGov and not entity_number %}
+{% if not isBeneficialOwnerGov and (not entity_number or OE_CONFIGS.FEATURE_FLAG_ENABLE_UPDATE_TRUSTS === "true") %}
   {% set errorMessage = {} if errors.beneficial_owner_nature_of_control_types %}
   {% include "includes/inputs/nature-of-control-checkboxes/trustees-nature-of-control-types.html" %}
 {% endif %}


views/includes/page-templates/beneficial-owner-nature-of-control-types.html
```diff
@@ -9,9 +9,16 @@
 {% set errorMessage = errors.beneficial_owner_nature_of_control_types if errors %}
 {% include "includes/inputs/nature-of-control-checkboxes/beneficial-owner-nature-of-control-types.html" %}
 
-{% if not isBeneficialOwnerGov and not entity_number %}
-  {% set errorMessage = {} if errors.beneficial_owner_nature_of_control_types %}
-  {% include "includes/inputs/nature-of-control-checkboxes/trustees-nature-of-control-types.html" %}
+{% if entity_number %}
```
conuirchgov, line 12: Tell me if I'm wrong.. but we want to see Trustee NOC if:
- they're not a BO gov 
AND
- in registration OR the flag is enabled?

So could this block/condition be simplified into...?

```
{% if not isBeneficialOwnerGov and (pageParams.isRegistration or pageParams.isUpdateTrustsFeatureEnabled) %}
  {% set errorMessage = {} if errors.beneficial_owner_nature_of_control_types %}
  {% include "includes/inputs/nature-of-control-checkboxes/trustees-nature-of-control-types.html" %}
{% endif %}
```

Also a couple of small other comments on naming in that suggestion:
- I'd recommend instead of just using the existence of the entity_number in the condition to signify being in registration or update, abstract that 'logic' away in a variable, i.e. `isRegistration` for example, to make reading the condition at a glance a little easier
- There is already an `isTrustFeatureEnabled` parameter in other pages/controllers using the feature flag introduced months ago. Even though it's not a direct conflict, distinguishing between the two by mentioning `Update` in the name could again make it clearer what flag is referenced here

views/includes/page-templates/beneficial-owner-nature-of-control-types.html
```diff
@@ -9,9 +9,16 @@
...
+  {% if not isBeneficialOwnerGov and pageParams.isTrustFeatureEnabled %}
```
stephanie-ch, line 13: You have access to the flag through nunjucks global variables, defined we have: `nunjucksEnv.addGlobal("OE_CONFIGS", config);` so in the view you can just do `OE_CONFIGS.<flagname>` and this avoids changing the controllers

views/includes/page-templates/beneficial-owner-nature-of-control-types.html
stephanie-ch: @conuirchgov there is a known "quirk" where if we submit a page in update journey and it triggers validation, when the page re-renders - you won't have access to the page params you passed in on the first load. This makes it tricky to use `isRegistration` flags in all the views. Happy to go into more detail if you want - but @cdrainCH let's use entity_number for now.

conuirchgov: Right... ok. That's interesting - never thought of that but I suppose that makes sense given the validation setup. 😬 So it's only `appData` we keep access to?

Tbh I didn't even mean to use pageParams, I'd just seen that as an example elsewhere, but even as a 

`{% set isRegistration = !entity_number %}`

in this template.

conuirchgov: Dismiss both naming points then 😂

Still think it would be worth simplifying the condition down to just to one expression.

views/update/review-beneficial-owner-individual.html
```diff
@@ -65,6 +65,7 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">Review the individual beneficial o
       {% set isBeneficialOwnerGov = false %}
       {% set natureOfControlText = "What is their nature of control?" %}
       {% set beneficial_owner_nature_of_control_types = beneficial_owner_nature_of_control_types %}
+      {% set trustees_nature_of_control_types = trustees_nature_of_control_types %}
```
conuirchgov, line 68: Can I ask what this is doing? From what I can see, nothing? 🙈 Same with the other view?

cdrainCH: It was removed as part of 678, but I suppose if we can't do a digital update while the OE has trusts then...yeah this is useless :D

test/controllers/update/update.beneficial.owner.other.controller.spec.ts
```diff
@@ -113,7 +112,6 @@ describe("UPDATE BENEFICIAL OWNER OTHER controller", () => {
   describe("GET tests", () => {
     test(`Renders the ${UPDATE_BENEFICIAL_OWNER_OTHER_PAGE} page`, async () => {
       mockGetApplicationData.mockReturnValueOnce({ ...APPLICATION_DATA_UPDATE_BO_MOCK });
-
       const resp = await request(app).get(UPDATE_BENEFICIAL_OWNER_OTHER_URL);
 
       expect(resp.status).toEqual(200);
```
corybudd-ch, line 117: do you want to undo your changes in this file?

cdrainCH: Done

AndreKramer7654 (approved):

corybudd-ch (approved):

conuirchgov (approved):

stephanie-ch (approved):

conuirchgov (approved):

PR #962, opened 2023-07-07
ROE-2282 Fix Sonar issues related to sign out redirection

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2282

### Change description

Addresses the security vulnerability within the Sign Out controller code:

* A relative URL is now used and checked when redirecting back to the previous page
* New test added to check that relative URL is correctly calculated

### Work checklist

### Merge instructions

```
```


src/controllers/sign.out.controller.ts
```diff
@@ -29,17 +28,34 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
...
+function getPreviousPageUrl(req: Request) {
+  const headers = req.rawHeaders;
+  const absolutePreviousPageUrl = headers.filter(item => item.includes(config.REGISTER_AN_OVERSEAS_ENTITY_URL))[0];
+
+  // Don't attempt to determine a relative previous page URL if no absolute URL is found
+  if (!absolutePreviousPageUrl) {
+    return absolutePreviousPageUrl;
+  }
+
+  const startingIndexOfRelativePath = absolutePreviousPageUrl.indexOf(config.REGISTER_AN_OVERSEAS_ENTITY_URL);
```
markpit, line 55: what happens if the absolutePreviousPageUrl doesn't contain the REGISTER_AN_OVERSEAS_ENTITY_URL, i think it will return '-1', which will be passed into substring - will that cause an error?

mwestacott: I have just tested this. It is fine because in the case that `absolutePreviousPageUrl` doesn't contain the REGISTER_AN_OVERSEAS_ENTITY_URL, the function will return at line 52. It will never get to the `indexOf` or `substring` calls.

I've added a new unit test to prove this though.

src/controllers/sign.out.controller.ts
```diff
@@ -29,17 +28,34 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
...
+    return safeRedirect(res, previousPage);
   } catch (error) {
     logger.errorRequest(req, error);
     next(error);
   }
 };
+
+function getPreviousPageUrl(req: Request) {
+  const headers = req.rawHeaders;
+  const absolutePreviousPageUrl = headers.filter(item => item.includes(config.REGISTER_AN_OVERSEAS_ENTITY_URL))[0];
```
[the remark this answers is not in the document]
mwestacott: The Update journey has its own sign out controller, so no issues here I think.

src/controllers/sign.out.controller.ts
```diff
@@ -29,17 +28,34 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
...
+
+  // Don't attempt to determine a relative previous page URL if no absolute URL is found
+  if (!absolutePreviousPageUrl) {
+    return absolutePreviousPageUrl;
+  }
+
+  const startingIndexOfRelativePath = absolutePreviousPageUrl.indexOf(config.REGISTER_AN_OVERSEAS_ENTITY_URL);
+  const relativePreviousPageUrl = absolutePreviousPageUrl.substring(startingIndexOfRelativePath);
+
+  logger.debug(`Relative previous page URL is ${relativePreviousPageUrl}`);
```
markpit, line 58: can you use debugRequest here and use the req param

src/controllers/sign.out.controller.ts
AndreKramer7654, line 55: Would it be safer to use:
const relativePreviousPageUrl = new URL(absolutePreviousPageUrl).pathname;

mwestacott: Not sure about safer, but definitely simpler. However, if the `absolutePreviousPageUrl` contains a sub-heading reference at the end, this isn't returned by the `pathname` function.

E.g. If `absolutePreviousPageUrl` is **http://hostname/register-an-overseas-entity/new-page#sub-heading** then only **/register-an-overseas-entity/new-page** is returned.

What do we think (@AndreKramer7654 @markpit ) is this ok?

AndreKramer7654: ok

AndreKramer7654 (approved):

src/controllers/sign.out.controller.ts
mwestacott: 👍

src/controllers/sign.out.controller.ts
mwestacott: Done.

markpit (approved):

PR #1003, opened 2023-08-03
UAR-794: Add trusts section to review update statement page

conuirchgov:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-794

### Change description

Added trusts summary section to Review Update Statement page (behind Update Trusts feature flag)

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/check.your.answers.ts
+++ b/src/utils/check.your.answers.ts
@@ -10,6 +10,7 @@ import { createOverseasEntity } from "../service/overseas.entities.service";
 import { OverseasEntityKey, Transactionkey } from "../model/data.types.model";
 import { closeTransaction, postTransaction } from "../service/transaction.service";
 import { startPaymentsSession } from "../service/payment.service";
+import { checkHasAnyBosWithTrusteeNocs } from "./update/trusts";
 
 import {
   OVERSEAS_ENTITY_UPDATE_DETAILS_URL,
@@ -19,11 +20,21 @@ import {
   UPDATE_AN_OVERSEAS_ENTITY_URL,
   CHS_URL,
   UPDATE_DO_YOU_WANT_TO_MAKE_OE_CHANGE_URL,
-  FEATURE_FLAG_ENABLE_TRUSTS_WEB
+  FEATURE_FLAG_ENABLE_TRUSTS_WEB,
+  FEATURE_FLAG_ENABLE_UPDATE_TRUSTS,
+  UPDATE_CHECK_YOUR_ANSWERS_PAGE,
+  UPDATE_REVIEW_STATEMENT_PAGE,
+  UPDATE_TRUSTS_ASSOCIATED_WITH_THE_OVERSEAS_ENTITY_URL,
+  UPDATE_BENEFICIAL_OWNER_TYPE_URL,
 } from "../config";
 import { RoleWithinTrustType } from "../model/role.within.trust.type.model";
 
-export const getDataForReview = (req: Request, res: Response, next: NextFunction, templateName: string, backLinkUrl: string, noChangeFlag?: boolean) => {
+export const getDataForReview = (req: Request, res: Response, next: NextFunction, isNoChangeJourney: boolean) => {
+  const hasAnyBosWithTrusteeNocs = checkHasAnyBosWithTrusteeNocs(getApplicationData(req.session));
+
+  const backLinkUrl = getBackLinkUrl(isNoChangeJourney, hasAnyBosWithTrusteeNocs);
+  const templateName = getTemplateName(isNoChangeJourney);
+
   try {
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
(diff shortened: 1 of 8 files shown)


joshuahigginsonCH (approved):

cdrainCH (approved):

src/utils/check.your.answers.ts
```diff
@@ -88,3 +100,21 @@ export const postDataForReview = async (req: Request, res: Response, next: NextF
...
+const getBackLinkUrl = (isNoChangeJourney: boolean, hasAnyBosWithTrusteeNocs: boolean) => {
+  if (isNoChangeJourney) {
+    return UPDATE_DO_YOU_WANT_TO_MAKE_OE_CHANGE_URL;
+  } else {
+    const updateTrustsEnabled = isActiveFeature(FEATURE_FLAG_ENABLE_UPDATE_TRUSTS);
+
+    return updateTrustsEnabled && hasAnyBosWithTrusteeNocs
+      ? UPDATE_TRUSTS_ASSOCIATED_WITH_THE_OVERSEAS_ENTITY_URL
+      : UPDATE_BENEFICIAL_OWNER_TYPE_URL;
+  }
```
AndreKramer7654, line 113: what about registration?

src/utils/update/trusts.ts
```diff
@@ -0,0 +1,12 @@
+import { ApplicationData } from "model";
+
+export const checkHasAnyBosWithTrusteeNocs = (appData: ApplicationData) => {
+  const bosThatCanHaveTrusteeNoc = [
+    ...(appData.beneficial_owners_individual || []),
+    ...(appData.beneficial_owners_corporate || []),
+    ...(appData.update?.review_beneficial_owners_individual || []),
```
AndreKramer7654, line 7: we should not be checking the review ones. They are temporary and moved into main list once reviewed!

src/utils/update/trusts.ts
```diff
@@ -0,0 +1,12 @@
+import { ApplicationData } from "model";
+
+export const checkHasAnyBosWithTrusteeNocs = (appData: ApplicationData) => {
+  const bosThatCanHaveTrusteeNoc = [
```
AndreKramer7654, line 4: quite a bit of work to join the lists into one big one just to check some. Could just use 
[appData.beneficial_owners_individual, appData.beneficial_owners_corporate].some(l => l.map(bo => bo.trustees_nature_of_control_types > 0));

src/utils/check.your.answers.ts
AndreKramer7654: ok - not used in registration

src/utils/update/trusts.ts
conuirchgov: Inclined to leave as is - does no harm in checking both.

And in the no change journey, where this is currently called, the beneficial owners are actually still in the review lists.

src/utils/update/trusts.ts
conuirchgov: I don't see the benefit of flatMap here either. Can you write up a suggestion? 

Tbh I think `[...(appData.array || []), ...]` is the simplest way of flattening those arrays.

src/utils/update/trusts.ts
AndreKramer7654: ok, it's important to know. We should really have separate checks for each set of lists. review and reviewed

src/utils/update/trusts.ts
AndreKramer7654: [ appData.array1, appData.array2].flat().some(a => a?.length > 0);

was hoping this handles undefines so no need for "|| []"

AndreKramer7654: typescript flat() does not seem to deal well with undefines. Some languages just filter those. So how about the some of some suggestion. Just to iterate over the list - not expand everything into a new list as [ ...array1, ...array2 ] does:

type X = {
   t? : Array<String>;
};

const l0 : Array<X> = [];
const l1 : Array<X> = [ {} ];
const l2 : Array<X> | undefined = undefined;
const l3 : Array<X> = [{ t: ["here "] }];

const result = [l0, l1, l2, l3].some(x => (x ?? []).some(y => ( y.t ?? []).length > 0));

console.log(result);

src/utils/update/trusts.ts
AndreKramer7654: Maybe just go over review lists then?

src/utils/update/trusts.ts
conuirchgov: I don't think this solution is as readable, and if there's an implied argument for potential performance improvements, I'd expect the benefit to be negligible.

Gonna make a call here and say I don't think this change would provide any real benefit.

src/utils/update/trusts.ts
conuirchgov: > ok, it's important to know. We should really have separate checks for each set of lists. review and reviewed

Should we? Currently no need for checking them independently here. 

> Maybe just go over review lists then?

As above, I'm inclined to leave it as is. Right now we just need to know if there are _any_ BOs with trustee NOC; where they are doesn't matter.

stephanie-ch (approved):

src/utils/update/trusts.ts
AndreKramer7654: There is duplication with scr/utils/trust.ts which checks for NoC too.

PR #1008, opened 2023-08-08
CC-458 Update node engine version

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/CC-458

### Change description

- Update Node engine version to 18 in package.json
- Ensure service fails in Tilt if Node version is incorrect

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/Tiltfile.dev
+++ b/Tiltfile.dev
@@ -1,6 +1,6 @@
 local_resource(
   name = 'dev:overseas-entities-web',
-  cmd = 'npm install && npm run build',
+  cmd = 'rm -rf ./dist && npm install && npm run build',
   deps = ['src', 'views']
 )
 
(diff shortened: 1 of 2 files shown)


Tiltfile.dev
```diff
@@ -1,6 +1,6 @@
 local_resource(
   name = 'dev:overseas-entities-web',
-  cmd = 'npm install && npm run build',
+  cmd = 'rm -rf ./dist && npm install && npm run build',
```
mwestacott, line 3: Deleting the `dist` folder here ensures that the OE web Tilt service fails to start up if in development mode and the service fails to build, e.g. if node or npm versions in the UNIX shell don't match the required version.

Without it, the Tilt dev:overseas-entities-web 'task' fails but the overseas-entities-web one doesn't (and seems to use a potentially older version of the compiled code, in the `dist` folder).

markpit (approved):

dlloyd1-ch (approved):

PR #1019, opened 2023-08-23
UAR-770: Error scenario - not added or ceased a bo

joshuahigginsonCH:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-770

### Work checklist

### Merge instructions

```
```


src/middleware/statement.validation.middleware.ts
```diff
@@ -35,6 +35,11 @@ const validateIdentifiedBOsStatement = (appData: ApplicationData, errorList: str
     return false;
   }
 
+  if (hasNotAddedOrCeasedBos(appData) && appData.update?.[RegistrableBeneficialOwnerKey] === 1) {
```
conuirchgov, line 38: Correct me if I'm wrong, but I think there are two issues here:

1. this method is for checking the model based on the statement stored in `appData[BeneficialOwnersStatementKey]` but this new check is checking against `appData.update[RegistrableBeneficialOwnerKey]`. 
2. the error message `ACTIVE_REGISTRABLE_BO` that's added on line 44 won't be considered if this condition is met, due to the early return statement on line 40.

I _think_ this should be pulled out into its own method and `&&` on line 52.

src/utils/application.data.ts
```diff
@@ -76,6 +76,16 @@ export const checkMOsDetailsEntered = (appData: ApplicationData): boolean => {
   return Boolean( appData[ManagingOfficerKey]?.length || appData[ManagingOfficerCorporateKey]?.length ) ;
 };
 
+export const hasNotAddedOrCeasedBos = (appData: ApplicationData): boolean => {
+  if (allBeneficialOwners(appData).some((bo) => !bo.ch_reference)) {
```
conuirchgov, line 80: I think this needs tweaked slightly.

If we've no ceased BOs this will return false prematurely, as we may still have added new BOs that have not ceased.

Could the two conditions be combined too? 

```javascript
const hasAddedOrCeasedABO = allBeneficialOwners(appData).some(bo => !bo.ch_reference || (bo.ceased_date && Object.keys(bo.ceased_date).length !== 0));

return !hasAddedOrCeasedABO;
```

(Separated the return statement to negate the result for readability, because the name of the method is the opposite of what the actual check is looking for. I'd almost argue to rename the method and negate the result where it's used but I'll not :joy:).

joshuahigginsonCH: So currently all we need to check for is that no bos have been added or ceased. So if either a BO has been added this should be false regardless.

I have taken the other comments from this in and added together and negated

src/middleware/statement.validation.middleware.ts
joshuahigginsonCH: Fixed to separate this out

src/utils/application.data.ts
conuirchgov: Ah, cool :thumbsup: The order of these checks has changed since I started writing that comment; I hope it was originally checking the ceased date first :joy:

joshuahigginsonCH: Yep 😄

conuirchgov (approved):

corybudd-ch (approved):

test/controllers/update/update.statement.validation.errors.controller.spec.ts
```diff
@@ -177,6 +179,41 @@ describe("Update statement validation errors controller", () => {
...
+      expect(resp.text).toContain(UPDATE_REGISTRABLE_BENEFICIAL_OWNER_URL);
+      expect(resp.text).toContain('Potato - OE991992');
+      expect(resp.text).toContain('There is at least one active registrable beneficial owner.');
+      expect(resp.text).toContain('The entity has reasonable cause to believe that at least one person has become or ceased to be a registrable beneficial owner during the update period.');
+
+      expect(resp.text).not.toContain('value="statement-resolution-change-information" checked');
+      expect(resp.text).not.toContain('value="statement-resolution-change-statement" checked');
+      expect(resp.text).not.toContain(PAGE_TITLE_ERROR);
+    });
+
```
corybudd-ch, line 216: left this comment on Christy's PR as well:

"

this test file keeps growing and it is a lot of repeated checks - are these tests really necessary compared to ones that are already there? I don't think we need to have separate tests in here for every scenario because your tests in test/middleware/statement.validation.middleware.spec.ts cover all the error scenarios.

You could just add your error message into req['statementErrorList'] of an existing test but even then not sure what value you're getting out of that.
"

stephanie-ch (approved):

PR #1044, opened 2023-09-07
uar-1037 always return value from checkErrorResponse (now called hasErrorResponse)

AndreKramer7654:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1037

### Change description
Make all code paths from the error check function return a value and use it to explicitly return undefined for 404

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/service/private.overseas.entity.details.ts
+++ b/src/service/private.overseas.entity.details.ts
@@ -22,7 +22,7 @@ export const getPrivateOeDetails = async (
     overseasEntityId,
   );
 
-  if (checkErrorResponse(req, response, overseasEntityId, transactionId, "overseas entity")) {
+  if (hasErrorResponse(req, response, overseasEntityId, transactionId, "overseas entity")) {
     return undefined;
   }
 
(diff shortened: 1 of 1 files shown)


src/service/private.overseas.entity.details.ts
```diff
@@ -22,7 +22,9 @@ export const getPrivateOeDetails = async (
     overseasEntityId,
   );
 
-  checkErrorResponse(req, response, overseasEntityId, transactionId, "overseas entity");
+  if (checkErrorResponse(req, response, overseasEntityId, transactionId, "overseas entity")) {
```
n-r-b, line 25: Just thinking whether there's a more appropriate name for this
```suggestion
  if (responseHasError(req, response, overseasEntityId, transactionId, "overseas entity")) {
```

```suggestion
  if (hasErrorResponse(req, response, overseasEntityId, transactionId, "overseas entity")) {
```

olaych (approved):

PR #1045, opened 2023-09-07
UAR-1105: Added mock data for Manage trusts

conuirchgov:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1105

### Change description

Added a service to return mock trust data to populate the model with.
- the call to this service is behind the Manage trusts feature flag
- existing tests in the two controllers cover what the data should look like before/after retrieval _without_ trusts
- don't think it's helpful at this stage to add a mapper (as specified in the ticket) as the mock data was taken directly from what's populated in the Add Trusts journey, so this has been left out

### Work checklist


src/utils/update/trusts_data_fetch.ts
```diff
@@ -0,0 +1,318 @@
...
+    appData.update = {};
+  }
+
+  appData.update.review_trusts = [];
+};
+
+const retrieveTrusts = (appData: ApplicationData) => {
+  if (!appData.update) {
+    appData.update = {};
+  }
```
cdrainCH, line 29: Both initialiseTrustsUpdateAppData and retrieveTrusts do the same thing. Do we need both?

cdrainCH, line 29: This is the same as initialiseTrustsUpdateAppData above. Will this change as the tickets progress?

PR #1071, opened 2023-09-18
Feature/uar 620 individual mo full dob displayed

bmcgreevyCH:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-620

### Change description
Individual MO full DOB displayed to the user


test/utils/update/managing.officer.mapper.spec.ts
```diff
@@ -178,16 +184,48 @@ describe("Test mapping to managing officer", () => {
...
+        [
+          {
+            ...MOCK_MANAGING_OFFICERS_PRIVATE_DATA[0],
+            residentialAddress: undefined,
+            principalAddress: undefined,
+            hashedId: 'hashedId1',
+          },
+        ];
+      const address = mapMoPrivateAddress(mockDataWithUndefinedAddresses, 'hashedId1', false);
+      expect(address).toEqual({});
```
bmcgreevyCH, line 196: I had to change this test as I realised it wasn't testing properly (wrong hashed Id). When I changed it I noticed that we don't return _undefined_ from mapBOMOAddress function, that function seems to return an empty object.

src/utils/update/managing.officer.mapper.ts
```diff
@@ -88,3 +88,11 @@ export const mapMoPrivateAddress = (moPrivateData: ManagingOfficerPrivateData[],
     }
   }
 };
+
+export const mapMoPrivateDOB = (moPrivateData: ManagingOfficerPrivateData[], ch_reference: string) => {
+  for (const managingOfficerData of moPrivateData) {
+    if (managingOfficerData.hashedId === ch_reference) {
+      return managingOfficerData.dateOfBirth ? mapInputDate(managingOfficerData.dateOfBirth) : undefined;
```
AndreKramer7654, line 95: mapInputDate handles undefined so no need for ?: conditional

src/utils/update/managing.officer.mapper.ts
```diff
@@ -88,3 +88,11 @@ export const mapMoPrivateAddress = (moPrivateData: ManagingOfficerPrivateData[],
     }
   }
 };
+
+export const mapMoPrivateDOB = (moPrivateData: ManagingOfficerPrivateData[], ch_reference: string) => {
+  for (const managingOfficerData of moPrivateData) {
```
n-r-b, line 93: We could combine both the `mapMoPrivateAddress` and `mapMoPrivateDOB` functions into one to avoid duplication, something like:

```
const mapManagingOfficerIndividualData = (managingOfficer: ManagingOfficerIndividual, moPrivateData) => {
  for (const managingOfficerData of moPrivateData) {
    if (managingOfficerData.hashedId === managingOfficer.ch_reference) {
      managingOfficer.usual_residential_address = mapBOMOAddress(managingOfficerData.residentialAddress);
      managingOfficer.date_of_birth = mapInputDate(managingOfficerData.dateOfBirth);
    }
  }
};
```

test/utils/update/beneficial_owners_managing_officers_data_fetch.spec.ts
```diff
@@ -26,12 +29,14 @@ describe("util beneficial owners managing officers data fetch", () => {
...
     mockGetCompanyPscService.mockReset();
     mockGetCompanyOfficers.mockReset();
+    mockGetManagingOfficerPrivateData.mockReset();
   });
 
   test("retrieveBoAndMoData sets BO & MO data in appData.update, sets appData.update.bo_mo_data_fetched", async () => {
     appData = {};
     mockGetCompanyPscService.mockReturnValue(MOCK_GET_COMPANY_PSC_ALL_BO_TYPES);
     mockGetCompanyOfficers.mockReturnValue(MOCK_GET_COMPANY_OFFICERS);
+    mockGetManagingOfficerPrivateData.mockReturnValue(MOCK_MO_PRIVATE_DATA);
```
n-r-b, line 39: Are these changes necessary? We don't call the private data service in `beneficial_owners_managing_officers_data_fetch.ts`

src/utils/update/managing.officer.mapper.ts
AndreKramer7654: Just one observation which may or may not help. We need to rely on having a day of birth so that the backend displays data correctly (2 open bugs). So may end up failing if no day found - which would make the day kind of special.

bmcgreevyCH: > We could combine both the `mapMoPrivateAddress` and `mapMoPrivateDOB` functions into one to avoid duplication, something like:
> 
> ```
> const mapManagingOfficerIndividualData = (managingOfficer: ManagingOfficerIndividual, moPrivateData) => {
>   for (const managingOfficerData of moPrivateData) {
>     if (managingOfficerData.hashedId === managingOfficer.ch_reference) {
>       managingOfficer.usual_residential_address = mapBOMOAddress(managingOfficerData.residentialAddress);
>       managingOfficer.date_of_birth = mapInputDate(managingOfficerData.dateOfBirth);
>     }
>   }
> };
> ```

Yeah I had considered this, I think I'd brought it up in one of our stand-ups, but I wasn't sure if there was more changes in the pipeline which might mean it changing again, so left it as is.

bmcgreevyCH: > Just one observation which may or may not help. We need to rely on having a day of birth so that the backend displays data correctly (2 open bugs). So may end up failing if no day found - which would make the day kind of special.

@AndreKramer7654 Can you elaborate on that a bit, sorry? Not sure where that leaves us regarding the above. :-)

bmcgreevyCH: > We could combine both the `mapMoPrivateAddress` and `mapMoPrivateDOB` functions into one to avoid duplication, something like:
> 
> ```
> const mapManagingOfficerIndividualData = (managingOfficer: ManagingOfficerIndividual, moPrivateData) => {
>   for (const managingOfficerData of moPrivateData) {
>     if (managingOfficerData.hashedId === managingOfficer.ch_reference) {
>       managingOfficer.usual_residential_address = mapBOMOAddress(managingOfficerData.residentialAddress);
>       managingOfficer.date_of_birth = mapInputDate(managingOfficerData.dateOfBirth);
>     }
>   }
> };
> ```

I did talk about this in one of the stand-ups then never did it. I think because the Corporate MO would need its own method then so it was give and take. However I'm going to make the change as it keeps code relative to MO type.

AndreKramer7654: Can discuss outside of PR but it may mean we should handle the DOB more separately. But can still do it in a combined method as suggested above.

bmcgreevyCH: > We could combine both the `mapMoPrivateAddress` and `mapMoPrivateDOB` functions into one to avoid duplication, something like:
> 
> ```
> const mapManagingOfficerIndividualData = (managingOfficer: ManagingOfficerIndividual, moPrivateData) => {
>   for (const managingOfficerData of moPrivateData) {
>     if (managingOfficerData.hashedId === managingOfficer.ch_reference) {
>       managingOfficer.usual_residential_address = mapBOMOAddress(managingOfficerData.residentialAddress);
>       managingOfficer.date_of_birth = mapInputDate(managingOfficerData.dateOfBirth);
>     }
>   }
> };
> ```

I mentioned doing this in Stand-up once but then didn't do it as there would still be a map address method for corporate. I coded it to keep the MO logic organised depending on MO type... bit of work reorganising test files and changing tests :-(

AndreKramer7654: This part all looks good to me now, thanks.

AndreKramer7654 (approved):

PR #1143, opened 2023-10-17
Roe-2413 bo other remove with url params

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-2413

### Change description

Add url params into the remove button for BO Other page

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/beneficial.owner.other.ts
+++ b/src/utils/beneficial.owner.other.ts
@@ -26,9 +26,8 @@ import {
   StartDateKeys
 } from "../model/date.model";
 import { v4 as uuidv4 } from 'uuid';
-import { isActiveFeature } from "./feature.flag";
 import * as config from "../config";
-import { getUrlWithParamsToPath } from "./url";
+import { addActiveSubmissionBasePathToTemplateData } from "./template.data";
 
 export const getBeneficialOwnerOther = (req: Request, res: Response, templateName: string, backLinkUrl: string) => {
   logger.debugRequest(req, `${req.method} ${req.route.path}`);
@@ -63,11 +62,10 @@ export const getBeneficialOwnerOtherById = (req: Request, res: Response, next: N
       [StartDateKey]: startDate
     };
 
-    // Redis removal work - Add extra template options if Redis Remove flag is true
+    // Redis removal work - Add extra template options if Redis Remove flag is true and on Registration journey
     const isRegistration: boolean = req.path.startsWith(config.LANDING_URL);
-    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL) && isRegistration) {
-      templateOptions.FEATURE_FLAG_ENABLE_REDIS_REMOVAL = true;
-      templateOptions.activeSubmissionBasePath = getUrlWithParamsToPath(config.ACTIVE_SUBMISSION_BASE_PATH, req);
+    if (isRegistration) {
+      addActiveSubmissionBasePathToTemplateData(templateOptions, req);
     }
 
     const appData = getApplicationData(req.session);
--- a/src/utils/beneficial.owner.other.ts
+++ b/src/utils/beneficial.owner.other.ts
@@ -26,6 +26,9 @@ import {
   StartDateKeys
 } from "../model/date.model";
 import { v4 as uuidv4 } from 'uuid';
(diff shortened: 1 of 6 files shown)


mwestacott (approved):

mattch1 (approved):

src/utils/beneficial.owner.other.ts
```diff
@@ -60,6 +63,13 @@ export const getBeneficialOwnerOtherById = (req: Request, res: Response, next: N
       [StartDateKey]: startDate
     };
 
+    // Redis removal work - Add extra template options if Redis Remove flag is true
+    const isRegistration: boolean = req.path.startsWith(config.LANDING_URL);
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL) && isRegistration) {
+      templateOptions.FEATURE_FLAG_ENABLE_REDIS_REMOVAL = true;
```
dlloyd1-ch, line 69: Could this be extracted to a method for consistency?  See addCeasedDateToTemplateOptions(). It also uses different syntax to add the filelds.

src/utils/beneficial.owner.other.ts
```diff
@@ -60,6 +63,13 @@ export const getBeneficialOwnerOtherById = (req: Request, res: Response, next: N
       [StartDateKey]: startDate
     };
 
+    // Redis removal work - Add extra template options if Redis Remove flag is true
```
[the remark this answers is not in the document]
mwestacott: Discussed. Makes it easier later on when we come to do update if this remains where it is.

src/utils/beneficial.owner.other.ts
markpit: yes can do, i've now extracted it into a new module/file utils/template.data.ts, a function called addActiveSubmissionBasePathToTemplateData.

src/utils/template.data.ts
```diff
@@ -0,0 +1,11 @@
+import { isActiveFeature } from "./feature.flag";
+import * as config from "../config";
+import { getUrlWithParamsToPath } from "./url";
+import { Request } from "express";
+
+export const addActiveSubmissionBasePathToTemplateData = (templateData: Object, req: Request) => {
+  if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL)) {
+    templateData["FEATURE_FLAG_ENABLE_REDIS_REMOVAL"] = true;
```
mwestacott, line 8: I preferred the old way of setting the field data - less verbose, without []s and ""s. Is there an advantage to doing it this way?

test/utils/template.data.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      const testObject = {
+        name: "Bob",
+        isOnRegister: false
+      };
+      const request = {} as Request;
+      addActiveSubmissionBasePathToTemplateData(testObject, request);
+
+      expect(testObject["FEATURE_FLAG_ENABLE_REDIS_REMOVAL"]).toEqual(true);
+      expect(testObject["activeSubmissionBasePath"]).toEqual(dummyUrl);
+    });
```
mwestacott, line 33: I'd also check that the existing fields set on the `testObject` are still there and unchanged. Here and in test below.

src/utils/template.data.ts
markpit: not sure, it might be safer with undefined's but i can ind out

test/utils/template.data.spec.ts
markpit: ok, i'll update the test

src/utils/template.data.ts
mwestacott: 👍 if we can avoid the extra characters would be good.

markpit: can't use the dot as it says property undefined on type Object, so the type has to have the field names in it. Not sure why it seemed to be ok before though.

test/utils/template.data.spec.ts
markpit: updated

src/utils/template.data.ts
mwestacott: Oh well.

mwestacott (approved):

PR #1158, opened 2023-10-20
[UAR-1066] populate reviewed trusts on trusts associated with oe page

corybudd-ch:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1066

### Change description

Screenshot 2023-10-18 at 16 29 09

2 tables look different because existing table (what you have added so far) has a bug where it creates a table with one row for each trust added, instead of adding a row to an existing table. So for example, if  you add 3 trusts the page will be rendered with 3 summaryLists each with 1 row. There is going to be a separate ticket raised to address this as it effects both Registration and Update.

"What you have reviewed" table creates one summaryList and adds rows.

### Work checklist

### Merge instructions

```
```


src/controllers/update/update.manage.trusts.orchestrator.controller.ts
```diff
@@ -60,6 +61,38 @@ const shouldGoToReviewTheTrust = (trustInReview: Trust) => {
...
+    return;
+  }
+};
+
+const putTrustInChangeScenario = async (req: Request, appData: ApplicationData, trustId: string) => {
+  const trustForChangeScenario = appData.trusts?.splice(appData.trusts.findIndex(trust => trust.trust_id === trustId), 1)[0];
+  appData.update?.review_trusts?.push(trustForChangeScenario as Trust);
+  const trust = getTrustInReview(appData) as Trust;
+
+  trust.review_status = {
```
AndreKramer7654, line 85: Instead of setting all to true possibly just change the last one or set trust.review_status == undefined?

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
```diff
@@ -60,6 +61,38 @@ const shouldGoToReviewTheTrust = (trustInReview: Trust) => {
...
+    res.redirect(UPDATE_MANAGE_TRUSTS_REVIEW_THE_TRUST_URL);
+  } catch (error) {
+    next(error);
+    return;
+  }
+};
+
+const putTrustInChangeScenario = async (req: Request, appData: ApplicationData, trustId: string) => {
+  const trustForChangeScenario = appData.trusts?.splice(appData.trusts.findIndex(trust => trust.trust_id === trustId), 1)[0];
+  appData.update?.review_trusts?.push(trustForChangeScenario as Trust);
```
AndreKramer7654, line 82: do we need these "as" casts?

src/routes/index.ts
```diff
@@ -519,6 +519,15 @@ router.route(config.UPDATE_MANAGE_TRUSTS_ORCHESTRATOR_URL)
   )
   .all(updateManageTrustsOrchestrator.handler);
 
+router.route(config.UPDATE_MANAGE_TRUSTS_ORCHESTRATOR_URL + config.TRUST_ID)
```
AndreKramer7654, line 522: could be more explicit about the path by adding a "/change_handler/" or something before the trust_id?

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
AndreKramer7654: and could even move the trust over to the main list. Then 1064 would not need to do the findIndex/splice.

AndreKramer7654: It's been "reviewed" so just changes from now on. Fine with keeping it "in review" but seems strange to have all the flags set to true already.

src/routes/index.ts
AndreKramer7654: I guess "reviewed-trust" is what it is recording?

corybudd-ch: Yeah happy to do that, will do

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
corybudd-ch: I have to set the others to true as well, otherwise the user after submitting will be forced to go through all the other pages as well and re-review the trustees etc. 
So by setting others to true and only reviewed_trust_details to false, the user's flow will be as required after clicking a change link - "review the trust" -> "individuals and entities involved"

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
corybudd-ch: potentially not, I cant remember but I usually do it to get around those compilation errors that'll comeplain you can't pass object of type 'Trust | undefined' into the push(), although we know (via middleware) that a user can't get to this route without having a valid trust to edit.

AndreKramer7654: It's because the splice is on appData.trusts? that it knows undefined is possible.

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
AndreKramer7654: I'm somewhat concerned that the user could skip through the review by going through this and having all (but one of) the flags set to true. If we are sure that won't happen then ok.

src/controllers/update/update.manage.trusts.orchestrator.controller.ts
AndreKramer7654: findIndex could return -1. So probably worth checking for. I saw a similar splice in moveTrustOutOfReview which also assumes something is found - that comes from a user input. All user input should be handled carefully.

corybudd-ch: It's not that it hasn't been thought of, that's the point of the middleware, it checks that the user hasn't passed in a garbage trustId that doesn't exist because the navigation check that runs is `navigation.hasTrustWithIdUpdate`.
So if a user tries to access 'change-handler-url-or-whatever/1234' they'll be redirected if the ID doesn't exist.

AndreKramer7654: Excellent!

joshuahigginsonCH (approved):

cdrainCH (approved):

AndreKramer7654 (approved):

PR #1162, opened 2023-10-23
UAR-1211: Statement date validations

jkayyala:
### JIRA link
Jira [UAR-1108](https://companieshouse.atlassian.net/browse/UAR-1108)

### Change description
Add validations so that start date of BO is before the filing date, as the filing date is not present in the BO page, filing date is retrieved from session.

Renamed checkCeasedDateOnOrAfterStartDate to checkFirstDateOnOrAfterSecondDate as its now used to check for dates other than ceasedDate and startDate

### Work checklist

### Merge instructions

```
```

[UAR-1108]: https://companieshouse.atlassian.net/browse/UAR-1108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -258,6 +258,8 @@ export const checkFirstDateOnOrAfterSecondDate = (
   return true;
 };
 
+export const checkCeasedDateOnOrAfterStartDate = checkFirstDateOnOrAfterSecondDate;
+
 export const checkDateOfBirth = (dayStr: string = "", monthStr: string = "", yearStr: string = "") => {
   // to prevent more than 1 error reported on the date fields we check if the year is correct length or missing before doing the date check as a whole.
   if (checkMoreThanOneDateFieldIsNotMissing(dayStr, monthStr, yearStr)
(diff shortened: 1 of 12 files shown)


src/validation/custom.validation.ts
```diff
@@ -505,6 +506,23 @@ export const checkBeneficialOwnersSubmission = (req) => {
   return true;
 };
 
+export const checkFilingPeriod = (req, secondDateDay: string, secondDateMonth: string, secondDateYear: string, errorMessage: string) => {
```
AndreKramer7654, line 509: Could we not use FilingPeriod? checkPreviousToFilingDate?

AndreKramer7654: also arg should probably be just date and not secondDate

src/validation/fields/date.validation.ts
```diff
@@ -17,7 +17,8 @@ import {
   MonthFieldErrors,
   YearFieldErrors,
   checkDateIPLegalEntityBO,
-  checkCeasedDateOnOrAfterStartDate
```
AndreKramer7654, line 20: we could just define checkCeasedDateOnOrAfterStartDate in terms of checkFirstDateOnOrAfterSecondDate and leave callers as is. Might be clearer too.

bantao-ch (approved):

src/validation/fields/date.validation.ts
```diff
@@ -67,19 +69,30 @@ const is_still_active_validations = (date_field_id: string, radio_button_id: str
 
 const is_trust_still_active_validation = (error_message: string) => [
   body("endDate")
-    .custom((value, { req }) => checkCeasedDateOnOrAfterStartDate(
```
AndreKramer7654, line 70: revert this change?

AndreKramer7654: ok new name is better!

AndreKramer7654 (approved):

PR #1164, opened 2023-10-24
ROE-2401 Trust details url parameters

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2401

### Change description

Updated trust interrupt page for ac1 
Updates to trust details for validation.

### Work checklist

### Merge instructions

```
```


src/controllers/trust.interrupt.controller.ts
```diff
@@ -40,10 +42,13 @@ const post = (
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
     const firstTrustId = "1";
-    return safeRedirect(res, `${config.TRUST_ENTRY_URL + "/" + firstTrustId}`);
+    let trustDetailsUrl = config.TRUST_ENTRY_URL;
```
mwestacott, line 45: Would consider calling this variable `trustEntryUrl`, to reflect the value that gets assigned.

test/controllers/trust.interrupt.controller.spec.ts
```diff
@@ -71,6 +87,29 @@ describe('Trust Interrupt controller', () => {
     });
   });
 
+  describe('POST with parms unit tests', () => {
```
mwestacott, line 90: with 'params'.

mattch1: changed parms to params

src/controllers/trust.interrupt.controller.ts
mattch1: Changed the name as suggested.

src/controllers/trust.interrupt.controller.ts
```diff
@@ -40,10 +42,13 @@ const post = (
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
     const firstTrustId = "1";
-    return safeRedirect(res, `${config.TRUST_ENTRY_URL + "/" + firstTrustId}`);
+    let trustEntryUrl = config.TRUST_ENTRY_URL;
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL) && req) {
```
markpit, line 46: you don't need the && req check here, it should always be present in this post function. We only added it into a previous one because the req was an optional parameter so we had to check it existed before using it.

mattch1: Removed.

mwestacott (approved):

PR #1206, opened 2023-11-15
UAR-1301 Add service unavailable screen for remove journey

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1301

### Change description
Adding service unavailable page for urls containing journey=remove when remove feature flag is off

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -23,6 +23,6 @@ export const transactionIdAndSubmissionIdExistInRequest = (req: Request): boolea
 const getTransactionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_TRANSACTION_ID];
 const getSubmissionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_SUBMISSION_ID];
 
-export const urlContainsRemoveJourneyQueryParam = (req: Request): boolean => {
+export const isRemoveJourney = (req: Request): boolean => {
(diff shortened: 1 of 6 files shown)


src/utils/url.ts
```diff
@@ -21,3 +22,7 @@ export const transactionIdAndSubmissionIdExistInRequest = (req: Request): boolea
 
 const getTransactionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_TRANSACTION_ID];
 const getSubmissionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_SUBMISSION_ID];
+
+export const urlContainsRemoveJourneyQueryParam = (req: Request): boolean => {
```
mwestacott, line 26: I don't think the function name is 100% correct - it's checking whether the URL contains the 'journey' query parameter and that it's _value_ is 'remove'.

How about just `isRemoveJourney`? How it's determining that from the request parameter passed in is a hidden implementation detail of this function (in the future, for a remove specific screen, it may also see if the page name prefix is something like 'remove', if parameter not passed in).

markpit: isRemoveJourney was my first thought but wasn't sure if some of the update pages we will re-use would have the journey param so we would have to also check the url is in some sort of 'remove' url list. But agree, isRemoveJourney makes better reading.

mwestacott (approved):

dlloyd1-ch (approved):
Looks good to me.

PR #1212, opened 2023-11-17
UAR-1072 Add new Disposed All Land screen

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1072

### Change description

* Page displays correctly after navigation from 'continue with saved filing' page
* Title banner and back link function as expected
* 'Yes' and 'No' options navigate to new URLs
* Unit tests added
* Refactored location of journey query parameter definition and associated constants
* Page validation not yet implemented

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -144,6 +144,8 @@ export const UPDATE_STATEMENT_VALIDATION_ERRORS_PAGE = "update-statement-validat
 
 // REMOVE TEMPLATES
 export const REMOVE_SOLD_ALL_LAND_FILTER_PAGE = "remove-sold-all-land-filter";
+export const REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE = "remove-is-the-overseas-entity-registered-as-the-owner";
+export const REMOVE_UNALLOWED = "remove-you-cannot-apply-to-remove-this-overseas-entity";
 
 // URL PARAMS
 export const ROUTE_PARAM_TRUST_ID = "trustId";
@@ -168,6 +170,16 @@ export const REVIEW_OWNER_INDEX_PARAM = "?index=";
 export const UPDATE_REVIEW_OWNERS_PARAMS = REVIEW_OWNER_INDEX_PARAM + ID;
 export const ROUTE_PARAM_TRANSACTION_ID = "transactionId";
 export const ROUTE_PARAM_SUBMISSION_ID = "submissionId";
+export const LANDING_PAGE_QUERY_PARAM = "start";
+export const JOURNEY_QUERY_PARAM = "journey";
+
+export enum JourneyType {
+  register = "register",
+  update = "update",
+  remove = "remove",
+}
+
+export const JOURNEY_REMOVE_QUERY_PARAM = `?${JOURNEY_QUERY_PARAM}=${JourneyType.remove}`;
 
 // ROUTING PATHs
 export const LANDING_URL = "/register-an-overseas-entity";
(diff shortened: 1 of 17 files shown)


src/controllers/update/update.continue.saved.filing.controller.ts
```diff
@@ -29,7 +28,8 @@ export const post = (req: Request, res: Response, _: NextFunction) => {
   }
 
   if (isRemoveJourney(req)){
-    return res.redirect(`${config.REMOVE_SOLD_ALL_LAND_FILTER_PAGE}?${JOURNEY_QUERY_PARAM}=${JourneyType.remove}`);
+    return res.redirect(`${config.REMOVE_SOLD_ALL_LAND_FILTER_PAGE}`);
```
mwestacott, line 31: Because where we're going is a remove specific page, there's no need to pass in the query parameter.

src/config/index.ts
```diff
@@ -168,6 +170,16 @@ export const REVIEW_OWNER_INDEX_PARAM = "?index=";
 export const UPDATE_REVIEW_OWNERS_PARAMS = REVIEW_OWNER_INDEX_PARAM + ID;
 export const ROUTE_PARAM_TRANSACTION_ID = "transactionId";
 export const ROUTE_PARAM_SUBMISSION_ID = "submissionId";
+export const LANDING_PAGE_QUERY_PARAM = "start";
```
mwestacott, line 173: Moved these constants here as they are URL related and nothing to do with the data model (at least not at present). Has lead to some more minor changes in this PR, due to changed imports etc. Discussed with @markpit and we thought it was better to do this sooner rather than later.

src/config/index.ts
```diff
@@ -168,6 +170,16 @@ export const REVIEW_OWNER_INDEX_PARAM = "?index=";
...
+export const JOURNEY_QUERY_PARAM = "journey";
+
+export enum JourneyType {
+  register = "register",
+  update = "update",
+  remove = "remove",
+}
+
+export const JOURNEY_REMOVE_QUERY_PARAM = `?${JOURNEY_QUERY_PARAM}=${JourneyType.remove}`;
```
mwestacott, line 182: Since we use this string in a couple of places at present and will use it in more in the future, thought we could have a common constant for it.

dlloyd1-ch (approved):
Looks good to me.

src/controllers/update/update.continue.saved.filing.controller.ts
markpit: do we need to add start=0, we will have to wipe the session data at the start of the journey similar to the secure update filter page

views/update/remove-sold-all-land-filter.html
```diff
@@ -0,0 +1,57 @@
+{% extends "update-layout.html" %}
```
mattch1, line 1: So this is going in under "views/update", there isn't going to be a "views/remove"?

mwestacott: It's a valid question, which I did consider. If we did do this then we would probably want to place all the other code assets (e.g. controllers) under a specific 'remove' folder also. I just thought, as remove journey is so closely entwined with the update journey and there are relatively few specific screens, they'd just live alongside.

What do others think?

mattch1: We could leave it for now but keep this open as the update directory is already large, so would make remove code a lot easier to find if it had its own location.

mattch1 (approved):

views/update/remove-sold-all-land-filter.html
mwestacott: 👍  We could leave as is and keep an eye on it. Identifying the assets is important, which is why they have the 'remove-' prefix in their name.

src/controllers/update/update.continue.saved.filing.controller.ts
markpit: ok, we'll have to work at out what point we wipe the session, on update it's the first screen after the continue question.

markpit (approved):

src/controllers/update/update.continue.saved.filing.controller.ts
mwestacott: 👍 potentially the same point as when update does that, but coming later...

PR #1219, opened 2023-11-20
Uar 1283 cannot apply to remove page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1283

### Change description

Updated the config page name of stop page and added the url
Routing
Controller
Template

### Work checklist

### Merge instructions

```
```


src/controllers/update/remove.cannot.use.controller.ts
```diff
@@ -0,0 +1,9 @@
+import { Request, Response } from "express";
+import * as config from "../../config";
+
+export const get = (req: Request, res: Response) => {
+  return res.render(config.REMOVE_CANNOT_USE_PAGE, {
+    backLinkUrl: config.SOLD_LAND_FILTER_URL,
```
markpit, line 6: back link will need the remove query parameter on it

mattch1: Updated backlink with query parameter and removed update sold land page and replaced it with the remove sold land version.

src/config/index.ts
```diff
@@ -145,7 +145,7 @@ export const UPDATE_STATEMENT_VALIDATION_ERRORS_PAGE = "update-statement-validat
 // REMOVE TEMPLATES
 export const REMOVE_SOLD_ALL_LAND_FILTER_PAGE = "remove-sold-all-land-filter";
 export const REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE = "remove-is-the-overseas-entity-registered-as-the-owner";
-export const REMOVE_UNALLOWED = "remove-you-cannot-apply-to-remove-this-overseas-entity";
+export const REMOVE_CANNOT_USE_PAGE = "remove-cannot-use";
```
mwestacott, line 148: Happy with the constant name change but wondering why value of the page name has changed. I think we normally keep these (close to) the name of the prototype page.

src/controllers/index.ts
```diff
@@ -35,6 +35,7 @@ export * as trustInterrupt from "./trust.interrupt.controller";
 export * as addTrust from "./add.trust.controller";
 export * as overseasName from "./overseas.name.controller";
 export * as paymentFailed from "./payment.failed.controller";
+export * as removeCannotUse from "./update/remove.cannot.use.controller";
```
mwestacott, line 38: Please move this down, to around line 103, grouped with the other remove controller definitions.

views/update/remove-cannot-use.html
```diff
@@ -0,0 +1,49 @@
+{% extends "update-layout.html" %}
+
+{% block pageTitle %}
+  You cannot apply to de-register this overseas entity - {{ serviceName }} - GOV.UK
```
mwestacott, line 4: Just wondering, does the tab title display correctly here, using `{{ serviceName }}`? It should reference the remove journey, not update.

mwestacott (commented):
I'm confused by some of the ACs on this story - they refer to a page other than 'You cannot apply to remove this overseas entity'. And I don't think we've done that page yet, so presumably the entire story can't be delivered?

mattch1:
ACs 1 and 2 should really be on other tickets.

src/config/index.ts
mattch1: I was following the naming of existing pages here rather than the prototype ones. e.g.
export const CANNOT_USE_PAGE = "cannot-use";

mwestacott:
> ACs 1 and 2 should really be on other tickets.

Yes, AC1 already on https://companieshouse.atlassian.net/browse/UAR-1072 (as AC2). AC2 _and_ AC3 should be on other tickets. Needs raising.

views/update/remove-cannot-use.html
mattch1: Copied what was done on https://github.com/companieshouse/overseas-entities-web/pull/1212/files, what do we do to control this I cannot see a service name  parameter anywhere.

mattch1: Maybe service name is not the right thing to put  and should be using what is in the "journey" parameter of the controller.

src/controllers/index.ts
mattch1: Moved this down

views/update/remove-cannot-use.html
mwestacott: I think `SERVICE_NAME` is what you need to reference. It's set by the code in the `update-layout.html` template.

mattch1: Just run this up - tab does have the correct title.

src/config/index.ts
mwestacott: Fair enough. What do others think?

mwestacott:
I ran this locally. Font here looks wrong:

Screenshot 2023-11-21 at 09 52 07

Do you see the same @mattch1 ?

src/config/index.ts
mattch1: If so, we should probably update the other cannot use page as well.

mattch1:
I did see the font but when I checked I have used the same styling  used on other pages.

views/update/remove-cannot-use.html
```diff
@@ -0,0 +1,49 @@
...
+            <a href="https://www.nidirect.gov.uk/articles/searching-land-registry" class="govuk-link">Land Registry for Northern
+            Ireland</a>
+          </li>
+        </ul>
+      </p>
+      <p class="govuk-body">
+        You may also need to <a href="https://www.gov.uk/guidance/file-an-overseas-entity-update-statement" class="govuk-link" data-event-id="update-guidance-link">file your overseas entity update statement</a> if it is due.
+      </p>
+      <p class="govuk-body">
+        <a href="javascript:void(0)" class="govuk-link">Contact Companies House</a> if you need to speak to someone.
```
mwestacott, line 45: Why is href here pointing at `javascript:void(0)`?

mattch1: Copied form template:

Acceptance Criteria 5
Given I am on the ‘You cannot apply to remove this overseas entity’ page

And I select the ‘Contact Companies House’ link

Then I am taken to <TBC>

src/config/index.ts
markpit: ok, if it's similar in registration then should be ok as is

views/update/remove-cannot-use.html
mwestacott: Right, so _"Then I am taken to <TBC>"_ means story is incomplete - need to check with PG what this should be. Also about those two ACs that look like they should be on another story. Happy to join a call on this, if required.

src/config/index.ts
mwestacott: Yes, happy with this too.

views/update/remove-cannot-use.html
mattch1: This AC should also be removed if this cannot be answered.

mwestacott: I agree.

src/controllers/update/remove.cannot.use.controller.ts
mwestacott: This now looks good to me.

views/update/remove-cannot-use.html
mwestacott: AC5 now removed from story, so this is now fine (for now).

mattch1:
https://companieshouse.atlassian.net/browse/UAR-1283
Removed :

Acceptance Criteria 5
Given I am on the ‘You cannot apply to remove this overseas entity’ page
And I select the ‘Contact Companies House’ link
Then I am taken to TBC

And two acceptance criteria concerning the land registry page.

We should be ready to merge now.

mwestacott (approved):
Story now updated (ACs moved) so this PR now ok, I think.

PR #1238, opened 2023-11-28
UAR-1328 OE Id screen for remove

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1328

### Change description

* Changes for backlink and title banner when screen initially displays
* Changes for backlink after validation has fired and failed
* Failing test skipped for now
    
**_Unit tests to follow!_**

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/navigation.ts
+++ b/src/utils/navigation.ts
@@ -3,7 +3,7 @@ import { Navigation } from "../model/navigation.model";
 import { ApplicationData } from "../model/application.model";
 import { WhoIsRegisteringType } from "../model/who.is.making.filing.model";
 import { isActiveFeature } from "./feature.flag";
-import { getUrlWithParamsToPath } from "./url";
+import { getUrlWithParamsToPath, isRemoveJourney } from "./url";
 import { Request } from "express";
 
 export const getEntityBackLink = (data: ApplicationData): string => {
@@ -19,6 +19,15 @@ export const getSoldLandFilterBackLink = (): string => {
     return config.LANDING_PAGE_URL;
   }
 };
+
+export const getRemoveBackLink = (req: Request, backLinkUrl: string): string => {
+  if (isRemoveJourney(req)) {
+    return `${backLinkUrl}${config.JOURNEY_REMOVE_QUERY_PARAM}`;
+  } else {
+    return backLinkUrl;
+  }
+};
+
 export const NAVIGATION: Navigation = {
   [config.STARTING_NEW_URL]: {
     currentPage: config.STARTING_NEW_PAGE,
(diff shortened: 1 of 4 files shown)


src/utils/navigation.ts
```diff
@@ -19,6 +19,15 @@ export const getSoldLandFilterBackLink = (): string => {
     return config.LANDING_PAGE_URL;
   }
 };
+
+export const getRemoveBackLink = (req: Request, backLinkUrl: string): string => {
```
mwestacott, line 23: So this function would need to be called from the 'previousPage' definition for the Update screens below, at least whilst the param URL is still required for journey identification (until data persisted to Mongo).

src/utils/navigation.ts
```diff
@@ -42,7 +51,7 @@ export const NAVIGATION: Navigation = {
   },
   [config.OVERSEAS_ENTITY_QUERY_URL]: {
     currentPage: config.OVERSEAS_ENTITY_QUERY_PAGE,
-    previousPage: () => config.UPDATE_INTERRUPT_CARD_URL,
+    previousPage: (appData: ApplicationData, req: Request) => getRemoveBackLink(req, config.UPDATE_INTERRUPT_CARD_URL),
```
markpit, line 54: i don't think it should be called getRemoveBackLink as it is not just for remove, so how about just getBackLink(...)

test/utils/navigation.spec.ts
```diff
@@ -172,7 +172,7 @@ describe("NAVIGATION utils", () => {
     expect(navigation).toEqual(config.OVERSEAS_ENTITY_QUERY_URL);
   });
 
-  test(`NAVIGATION returns ${config.UPDATE_INTERRUPT_CARD_URL} when calling previousPage on ${config.OVERSEAS_ENTITY_QUERY_URL} object`, () => {
+  test.skip(`NAVIGATION returns ${config.UPDATE_INTERRUPT_CARD_URL} when calling previousPage on ${config.OVERSEAS_ENTITY_QUERY_URL} object`, () => {
```
markpit, line 175: skip will need removing

src/utils/navigation.ts
markpit: also, do you need to include appData? it doesn't get passed to getRemoveBackLink

test/utils/navigation.spec.ts
mwestacott: Yep, as per comment in the commit message. I've got a commit ready to go now, with this test fixed.

src/utils/navigation.ts
mwestacott: True. We wouldn't call this for a 'ROE registration' screen though. Should it maybe be called `getUpdateRemoveBackLink`?

markpit: sounds good, could have an 'or' in there - getUpdateOrRemoveBackLink

mwestacott: > also, do you need to include appData? it doesn't get passed to getRemoveBackLink

Yes, because it's an optional param. This question has cropped up before, but yes, needed.

markpit (approved):

PR #1242, opened 2023-11-29
UAR-1329 Remove journey navigation to confirm correct entity page

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1329

### Change description

Updated entity query post method function to include journey query parameter for the remove journey.
Updated get function on entity details to pass in remove journey to the template when the query parameter for the remove journey is present to display the remove banner.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/test/utils/navigation.spec.ts
+++ b/test/utils/navigation.spec.ts
@@ -13,6 +13,9 @@ import { getApplicationData } from "../../src/utils/application.data";
 const mockIsActiveFeature = isActiveFeature as jest.Mock;
 const mockGetApplicationData = getApplicationData as jest.Mock;
 const mockRemoveRequest = { } as Request;
+mockRemoveRequest["query"] = {
+  "journey": "remove"
+};
 
 describe("NAVIGATION utils", () => {
 
@@ -38,9 +41,6 @@ describe("NAVIGATION utils", () => {
   });
 
   test(`getRemoveBackLink returns a URL with the 'journey' query parameter present when on the Remove journey`, () => {
-    mockRemoveRequest["query"] = {
-      "journey": "remove"
-    };
     const removeBackLink = getUpdateOrRemoveBackLink(mockRemoveRequest, config.UPDATE_LANDING_PAGE_URL);
     expect(removeBackLink).toEqual(`${config.UPDATE_LANDING_PAGE_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`);
   });
@@ -189,17 +189,11 @@ describe("NAVIGATION utils", () => {
   });
 
   test(`NAVIGATION returns ${config.UPDATE_INTERRUPT_CARD_URL} with 'journey' param set when calling previousPage on ${config.OVERSEAS_ENTITY_QUERY_URL} object for Remove journey`, () => {
-    mockRemoveRequest["query"] = {
-      "journey": "remove"
-    };
     const navigation = NAVIGATION[config.OVERSEAS_ENTITY_QUERY_URL].previousPage(undefined, mockRemoveRequest);
     expect(navigation).toEqual(`${config.UPDATE_INTERRUPT_CARD_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`);
   });
 
   test(`NAVIGATION returns ${config.OVERSEAS_ENTITY_QUERY_URL} with 'journey' param set when calling previousPage on ${config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL} object for Remove journey`, () => {
-    mockRemoveRequest["query"] = {
-      "journey": "remove"
-    };
     const navigation = NAVIGATION[config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL].previousPage(undefined, mockRemoveRequest);
     expect(navigation).toEqual(`${config.OVERSEAS_ENTITY_QUERY_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`);
   });


src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,12 +16,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     const appData: ApplicationData = getApplicationData(req.session as Session);
     const update = appData.update as Update;
 
+    let journeyQuery = "";
```
mwestacott, line 19: Please follow the pattern used on other screens - so if a remove journey, have a new `res.render` command in the conditional block and leave the existing one unchanged.

src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,12 +16,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
+    let journeyType = config.JourneyType.update;
+    if (isRemoveJourney(req)) {
+      journeyType = config.JourneyType.remove;
+      journeyQuery = "?joureny=remove";
+    }
+
     return res.render(config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE, {
-      backLinkUrl: config.OVERSEAS_ENTITY_QUERY_URL,
+      backLinkUrl: `${config.OVERSEAS_ENTITY_QUERY_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`,
```
markpit, line 27: i would duplicate the res.render so we leave the original one untouched, but have another res.render if isRemoveJourney(req) is true

src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,12 +16,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     const appData: ApplicationData = getApplicationData(req.session as Session);
     const update = appData.update as Update;
 
+    let journeyQuery = "";
+    let journeyType = config.JourneyType.update;
+    if (isRemoveJourney(req)) {
+      journeyType = config.JourneyType.remove;
+      journeyQuery = "?joureny=remove";
```
mwestacott, line 23: Has this code been unit tested and tested in Tilt/Docker. There's a spelling mistake here which would mean things shouldn't work correctly...

This shouldn't be needed anyway, if you copy other example code in earlier controllers that have been adapted for Remove.

markpit, line 23: typo in the query , but it is also available in the config, it's the same one you've used on line 27

src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,12 +16,21 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
     return res.render(config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE, {
-      backLinkUrl: config.OVERSEAS_ENTITY_QUERY_URL,
+      backLinkUrl: `${config.OVERSEAS_ENTITY_QUERY_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`,
       updateUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE,
       appData,
-      registrationDate: update.date_of_creation
+      registrationDate: update.date_of_creation,
+      journey: journeyType,
+      journeyQuery
```
markpit, line 33: shouldn't need to pass the query into the page, if we pass journey then we can add the query from the config value based on the value of the journey field

src/controllers/update/confirm.overseas.entity.details.controller.ts
mattch1: Removed this and rewritten it elsewhere.

src/controllers/update/confirm.overseas.entity.details.controller.ts
mattch1: Updated this.

src/controllers/update/confirm.overseas.entity.details.controller.ts
mattch1: Oddly enough it seemed to work when I tested it but have updated and retested it now; reminds me of the mix of redirects some to pages  and some to urls  that might just be a naming mixup otherwise a lot seems to work when it shouldn't.

src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,13 +16,25 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
+      });
+    }
+
     return res.render(config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE, {
       backLinkUrl: config.OVERSEAS_ENTITY_QUERY_URL,
       updateUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE,
       appData,
-      registrationDate: update.date_of_creation
+      registrationDate: update.date_of_creation,
```
markpit, line 35: i'd remove the comma that was added here so it is the same as before and won't show up as a change

views/update/confirm-overseas-entity-details.html
```diff
@@ -13,6 +13,12 @@
 
 {% set displayRegistrationDate = dateMacros.formatDate(registrationDate["day"], registrationDate["month"], registrationDate["year"]) %}
 
+{% if (journey == "remove") %}
+  {% set chooseLink = "/update-an-overseas-entity/overseas-entity-query?journey=remove" %}
```
markpit, line 17: you should be able to use OE_CONFIGS.JOURNEY_REMOVE_QUERY_PARAM instead of the ?journey=remove
as it is already inside a {% %} block i think you could use 
{% set chooseLink = "/update-an-overseas-entity/overseas-entity-query" = OE_CONFIGS.JOURNEY_REMOVE_QUERY_PARAM %}

views/update/confirm-overseas-entity-details.html
```diff
@@ -13,6 +13,12 @@
 
 {% set displayRegistrationDate = dateMacros.formatDate(registrationDate["day"], registrationDate["month"], registrationDate["year"]) %}
 
+{% if (journey == "remove") %}
```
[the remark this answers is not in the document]
mattch1: use triple =

views/update/confirm-overseas-entity-details.html
mattch1: using var

src/controllers/update/confirm.overseas.entity.details.controller.ts
```diff
@@ -15,13 +16,25 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     const appData: ApplicationData = getApplicationData(req.session as Session);
     const update = appData.update as Update;
 
+    if (isRemoveJourney(req)) {
+      return res.render(config.CONFIRM_OVERSEAS_ENTITY_DETAILS_PAGE, {
+        journey: config.JourneyType.remove,
+        backLinkUrl: `${config.OVERSEAS_ENTITY_QUERY_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`,
```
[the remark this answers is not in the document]
mwestacott: I've just merged my PR into main @mattch1 . Give me a shout if any questions.

test/__mocks__/text.mock.ts
```diff
@@ -256,3 +256,6 @@ export const UPDATE_MANAGE_TRUSTS_REVIEW_FORMER_BO_TITLE = "Review former benefi
 export const UPDATE_MANAGE_TRUSTS_TELL_US_ABOUT_FORMER_BO_TITLE = "Tell us about the former beneficial owner";
 export const UPDATE_MANAGE_TRUSTS_REVIEW_FORMER_BO_TABLE_HEADING = "<th scope=\"col\" class=\"govuk-table__header\">Name</th>";
 export const UPDATE_MANAGE_TRUSTS_TELL_US_ABOUT_THE_LEGAL_ENTITY_TITLE = "Tell us about the legal entity";
+export const UPDATE_BANNER = "File an overseas entity update statement";
```
mwestacott, line 259: I don't think these definitions are required - you can just reference the UPDATE_SERVICE_NAME and REMOVE_SERVICE_NAME constants in the main `config/index.ts` file. That's what's been done in the other tests.

test/controllers/update/confirm.overseas.entity.details.controller.spec.ts
```diff
@@ -69,6 +75,20 @@ describe("Confirm company data", () => {
...
       expect(resp.text).toContain(testIncorporationCountry);
+      expect(resp.text).toContain(UPDATE_BANNER);
+    });
+
+    test(`renders the ${config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL} page for the remove journey`, async () => {
+
+      mockGetApplicationData.mockReturnValueOnce(entityProfileModelMock).mockReturnValueOnce(entityProfileModelMock);
+      const resp = await request(app).get(`${config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`);
+      expect(resp.status).toEqual(200);
+      expect(resp.text).toContain(BACK_LINK_FOR_UPDATE_OE_CONFIRM);
```
mwestacott, line 86: The assertion on this line is passing but it needs to be more specific. Something like:
   `expect(resp.text).toContain(`${BACK_LINK_FOR_UPDATE_OE_CONFIRM}${config.JOURNEY_REMOVE_QUERY_PARAM}`);`

As query param should be on end of the back link. In fact, test may fail after changing this, until you incorporate the fix that @markpit highlighted above.

src/controllers/update/confirm.overseas.entity.details.controller.ts
mwestacott: Thanks.

src/controllers/update/confirm.overseas.entity.details.controller.ts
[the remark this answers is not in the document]
mattch1: Updated the navigation.

test/controllers/update/confirm.overseas.entity.details.controller.spec.ts
mattch1: Add specifics.

test/utils/navigation.spec.ts
```diff
@@ -196,6 +197,15 @@ describe("NAVIGATION utils", () => {
     expect(navigation).toEqual(`${config.UPDATE_INTERRUPT_CARD_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`);
   });
 
+  test(`NAVIGATION returns ${config.OVERSEAS_ENTITY_QUERY_URL} with 'journey' param set when calling previousPage on ${config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL} object for Remove journey`, () => {
+    const mockRequest = { } as Request;
```
mwestacott, line 201: Perhaps worth defining this as a `mockRemoveRequest` at top of file as it's also used in the previous test I defined. We can then remove the definition here and in my test method and reduce LOC. It's going to be needed in more test methods going forwards too (I expect).

test/utils/navigation.spec.ts
```diff
@@ -37,11 +38,10 @@ describe("NAVIGATION utils", () => {
   });
 
   test(`getRemoveBackLink returns a URL with the 'journey' query parameter present when on the Remove journey`, () => {
-    const mockRequest = { } as Request;
-    mockRequest["query"] = {
+    mockRemoveRequest["query"] = {
```
mwestacott, line 41: Can these three lines also move up and be removed here and below?

 ```
   mockRequest["query"] = {	
      "journey": "remove"	
    };	
```

mattch1: Moved these up as well.

markpit (approved):

PR #1250, opened 2023-12-06
Uat 1332 remove store radio selections

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1332

### Change description
Updated 
- start new
- remove sold land filter
- remove entity is registered
Click back link previous input is displayed in the radio buttons.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/shared/common.resume.submission.controller.ts
+++ b/src/controllers/shared/common.resume.submission.controller.ts
@@ -16,7 +16,6 @@ import {
   IsSecureRegisterKey,
   OverseasEntityKey,
   Transactionkey,
-  ContinueSavedKey,
   HasDisposedAllLandKey
 } from "../../model/data.types.model";
 import { WhoIsRegisteringKey, WhoIsRegisteringType } from "../../model/who.is.making.filing.model";
@@ -105,7 +104,6 @@ const setWebApplicationData = (session: Session, appData: ApplicationData, trans
   appData[IsSecureRegisterKey] = '0';
   appData[Transactionkey] = transactionId;
   appData[OverseasEntityKey] = overseaEntityId;
-  appData[ContinueSavedKey] = '0';
   appData[HasDisposedAllLandKey] = '0';
 
   if (Object.keys(appData[OverseasEntityDueDiligenceKey] as OverseasEntityDueDiligence).length) {
--- a/src/controllers/shared/common.resume.submission.controller.ts
+++ b/src/controllers/shared/common.resume.submission.controller.ts
@@ -15,8 +15,7 @@ import {
   ID,
   IsSecureRegisterKey,
   OverseasEntityKey,
-  Transactionkey,
-  IsListedAsPropertyOwnerKey
+  Transactionkey
 } from "../../model/data.types.model";
 import { WhoIsRegisteringKey, WhoIsRegisteringType } from "../../model/who.is.making.filing.model";
 import { OverseasEntityDueDiligence, OverseasEntityDueDiligenceKey } from "../../model/overseas.entity.due.diligence.model";
@@ -104,7 +103,6 @@ const setWebApplicationData = (session: Session, appData: ApplicationData, trans
   appData[IsSecureRegisterKey] = '0';
   appData[Transactionkey] = transactionId;
   appData[OverseasEntityKey] = overseaEntityId;
-  appData[IsListedAsPropertyOwnerKey] = '0';
 
   if (Object.keys(appData[OverseasEntityDueDiligenceKey] as OverseasEntityDueDiligence).length) {
     appData[WhoIsRegisteringKey] = WhoIsRegisteringType.SOMEONE_ELSE;
--- a/src/controllers/shared/common.resume.submission.controller.ts
+++ b/src/controllers/shared/common.resume.submission.controller.ts
@@ -10,7 +10,15 @@ import { setExtraData } from "../../utils/application.data";
 import { isActiveFeature } from "../../utils/feature.flag";
 import { getOverseasEntity } from "../../service/overseas.entities.service";
 
-import { HasSoldLandKey, ID, IsSecureRegisterKey, OverseasEntityKey, Transactionkey } from "../../model/data.types.model";
+import {
+  HasSoldLandKey,
+  ID,
+  IsSecureRegisterKey,
+  OverseasEntityKey,
+  Transactionkey,
+  ContinueSavedKey,
+  HasDisposedAllLandKey
+} from "../../model/data.types.model";
 import { WhoIsRegisteringKey, WhoIsRegisteringType } from "../../model/who.is.making.filing.model";
 import { OverseasEntityDueDiligence, OverseasEntityDueDiligenceKey } from "../../model/overseas.entity.due.diligence.model";
 import { DueDiligence, DueDiligenceKey } from "../../model/due.diligence.model";
@@ -97,6 +105,8 @@ const setWebApplicationData = (session: Session, appData: ApplicationData, trans
   appData[IsSecureRegisterKey] = '0';
   appData[Transactionkey] = transactionId;
   appData[OverseasEntityKey] = overseaEntityId;
+  appData[ContinueSavedKey] = '0';
+  appData[HasDisposedAllLandKey] = '0';
 
   if (Object.keys(appData[OverseasEntityDueDiligenceKey] as OverseasEntityDueDiligence).length) {
     appData[WhoIsRegisteringKey] = WhoIsRegisteringType.SOMEONE_ELSE;
--- a/src/controllers/shared/common.resume.submission.controller.ts
+++ b/src/controllers/shared/common.resume.submission.controller.ts
@@ -16,7 +16,7 @@ import {
   IsSecureRegisterKey,
   OverseasEntityKey,
   Transactionkey,
-  HasDisposedAllLandKey
+  IsListedAsPropertyOwnerKey
 } from "../../model/data.types.model";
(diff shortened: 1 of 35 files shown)


src/controllers/update/remove.is.entity.registered.owner.controller.ts
```diff
@@ -1,15 +1,20 @@
...
   try {
     logger.debugRequest(req, `GET ${config.REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE}`);
+    const appData: ApplicationData = getApplicationData(req.session);
 
     return res.render(config.REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE, {
       journey: config.JourneyType.remove,
       backLinkUrl: `${config.REMOVE_SOLD_ALL_LAND_FILTER_URL}${config.JOURNEY_REMOVE_QUERY_PARAM}`,
-      templateName: config.REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE
+      templateName: config.REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE,
+      [HasDisposedAllLandKey]: appData?.[HasDisposedAllLandKey]
```
markpit, line 17: the sold-land-filter page asks 'Has the overseas entity disposed of all its property or land in the UK?' so the sold-land-filter controller and HasSoldLandKey will handle the 'disposed all land' value.  This controller is for 'Is the overseas entity currently listed on any land registry records as the registered owner of property or land in the UK?' so we'll need the key and model field to represent that. something like is_listed_as_land_or_property_owner

mattch1: A good point, but I would advise against any name changes, as soon as you attempt to change the name the code reacts badly, you enter a yes or a no , the error is displayed as though no input has been entered at and the user will not be able to go any further. Let's say I change all instances "owner_disposed" to "listed_as_property_owner" such that they correspond as before this is the result: 
Screenshot 2023-12-07 at 08 43 37 (2)
There seems to be a very strong bond with the existing name.

mattch1: Updated the name but to something a bit shorter than the suggestion above.

views/update/remove-is-entity-registered-owner.html
```diff
@@ -40,17 +40,19 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">Is the overseas entity curr
           },
           items: [
             {
-              value: "yes",
+              value: 1,
```
dlloyd1-ch, line 43: If changing from a string to boolean can we use true/false values for readability?

markpit: i think 1/0 might be the nunjucks component defaults, i know we used 1/0 in registration so they are consistent

views/update/remove-is-entity-registered-owner.html
```diff
@@ -27,10 +27,10 @@ <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">Is the overseas entity curr
         }) }}
 
         {{ govukRadios({
-          errorMessage: errors.owner_disposed if errors,
+          errorMessage: errors.listed_as_property_owner if errors,
           classes: "govuk-radios",
-          idPrefix: "owner_disposed",
-          name: "owner_disposed",
+          idPrefix: "listed_as_property_owner",
```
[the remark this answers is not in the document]
mattch1: Owner disposed sounded too much like the previous sold land filter page this would cause serious confusion so we need a name that is clearly different here.

src/model/data.types.model.ts
```diff
@@ -83,3 +83,6 @@ export const EntityNumberKey = "entity_number";
 export const UpdateModelKey = "update";
 export const DoYouWantToRemoveKey = "do_you_want_to_remove";
 export const AnyTrustsInvolvedKey = "any_trusts_involved";
+
+// Remove Journey
+export const IsListedAsPropertyOwnerKey = "listed_as_property_owner";
```
markpit, line 88: can you add is_ to the "listed_as_property_owner", so it matches the IsListedAsPropertyOwnerKey and then it will be consistent with 
export const IsSecureRegisterKey = "is_secure_register";

markpit: the "disposed_all_land" is the name of the html element, not the field in our model. I'd prefer to have 
export const IsListedAsPropertyOwnerKey = "is_listed_as_property_owner";
instead of 
export const IsListedAsPropertyOwnerKey = "listed_as_property_owner";
on the model.
Any preference @dlloyd1-ch @mwestacott ?

mattch1: Made this further update

mattch1:
Looked at @markpit 's updates - LGTM - will let others update with any comments.

src/model/application.model.ts
```diff
@@ -38,7 +39,7 @@ export interface ApplicationData {
     is_secure_register?: string;
     who_is_registering?: string;
     update?: updateType.Update;
-    is_listed_as_property_owner?: string;
+    remove?: removeType.Remove;
```
mwestacott, line 42: Probably don't need to do this right now (and perhaps after [this](https://github.com/companieshouse/overseas-entities-api/pull/369) PR has been merged) but need to remember to persist the new `remove` block data in Mongo at some point. So changes in the SDK and API to support this.

I don't think the `is_listed_as_property_owner` field would be saved in Mongo anyway (as a filter page response) but we need to do this for the statement field.

src/utils/application.data.ts
```diff
@@ -161,3 +162,7 @@ const getIndexInApplicationData = (req: Request, appData: ApplicationData, key:
     throw createAndLogErrorRequest(req, `application.data getIndexInApplicationData - unable to find object in session data for key ${key} and ID ${id}`);
   }
 };
+
+export const getRemove = (appData: ApplicationData): Remove => {
```
mwestacott, line 166: Fine, but just wondering why this was added as we don't have a similar function for Update.

markpit: I was using `remove = appData.remove || {} ` in a few places so thought it would be good to have it one place with the `|| {}` to return an empty object instead of undefined. Happy to change it back if needed.

src/controllers/update/remove.is.entity.registered.owner.controller.ts
```diff
@@ -19,7 +26,15 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
-    if (req.body["owner_disposed"] === 'no') {
+    const isListedAsPropertyOwner = req.body[IsListedAsPropertyOwnerKey];
+
+    const appData: ApplicationData = getApplicationData(req.session);
+    const remove = getRemove(appData);
+    remove[IsListedAsPropertyOwnerKey] = isListedAsPropertyOwner;
+
+    setApplicationData(req.session, remove, RemoveKey);
+
+    if (isListedAsPropertyOwner === '0') {
```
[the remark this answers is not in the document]
markpit: yes, could introduce new constants and start using them.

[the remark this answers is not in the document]
dlloyd1-ch: > Would be nice to have a couple of string constants defined for these '0' and '1' values (they appear in a lot of places and not that readable), e.g. `OPTION_NO` and `OPTION_YES`. Too late?

I agree. It would make it more readable and was something I raised in an earlier review.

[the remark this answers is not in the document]
mwestacott: 👍

src/utils/application.data.ts
mwestacott: No. All good I think.

src/controllers/update/remove.is.entity.registered.owner.controller.ts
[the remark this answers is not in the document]
markpit: update to use config values for '1' and '0', went with OPTION_YES , OPTION_NO

src/controllers/shared/common.resume.submission.controller.ts
```diff
@@ -10,7 +10,13 @@ import { setExtraData } from "../../utils/application.data";
...
 import { getOverseasEntity } from "../../service/overseas.entities.service";
 
-import { HasSoldLandKey, ID, IsSecureRegisterKey, OverseasEntityKey, Transactionkey } from "../../model/data.types.model";
+import {
+  HasSoldLandKey,
+  ID,
+  IsSecureRegisterKey,
+  OverseasEntityKey,
+  Transactionkey
+} from "../../model/data.types.model";
```
markpit, line 19: I undid the changes to the resume controller, which were setting default values for is_listed_as_property_owner, I assumed we'd work on the resuming in a different ticket.

src/controllers/update/remove.is.entity.registered.owner.controller.ts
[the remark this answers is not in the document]
markpit: missed a unit test, i'll update it

src/controllers/shared/common.resume.submission.controller.ts
mwestacott: Yes, a number of things to do to get resume working.

src/controllers/update/remove.is.entity.registered.owner.controller.ts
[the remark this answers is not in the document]
markpit: updated

mwestacott (approved):
Thanks for making those constant name changes. Looks good.

mwestacott:
I'm still happy with this PR.

PR #1263, opened 2023-12-15
Uar 1285 investigate new approach

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1285

### Change description

Changed implementation to have a single 'isRemoveJourney()' function that checks session data first and query param second. Also fixed the resulting, failing tests.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -31,7 +31,7 @@ export const isRemoveJourney = (req: Request): boolean => {
   const appData: ApplicationData = getApplicationData(session);
 
   if (appData) {
-    if (appData.is_remove === true) {
+    if (appData['is_remove'] === true) {
       return true;
     }
     // TODO Decide if return false here if appData.is_remove = false...
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -31,10 +31,9 @@ export const isRemoveJourney = (req: Request): boolean => {
   const appData: ApplicationData = getApplicationData(session);
 
   if (appData) {
-    if (appData['is_remove'] === true) {
-      return true;
+    if (appData['is_remove'] !== undefined && appData['is_remove'] !== null) {
+      return appData['is_remove'];
     }
-    // TODO Decide whether to return false here, if appData['is_remove'] is present and value is 'false'
   }
 
   return req.query[config.JOURNEY_QUERY_PARAM] === config.JourneyType.remove;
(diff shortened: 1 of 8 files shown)


src/utils/url.ts
```diff
@@ -24,14 +26,16 @@ const getTransactionIdFromRequestParams = (req: Request): string => req.params[c
...
+
+  const appData: ApplicationData = getApplicationData(session);
 
-export const isRemoveJourneyInSession = (appData: ApplicationData ): boolean => {
   if (appData) {
-    if (appData.is_remove === true) {
+    if (appData['is_remove'] === true) {
       return true;
     }
+    // TODO Decide whether to return false here, if appData['is_remove'] is present and value is 'false'
```
mwestacott, line 37: Need to discuss this.

src/utils/url.ts
```diff
@@ -24,14 +27,15 @@ const getTransactionIdFromRequestParams = (req: Request): string => req.params[c
 const getSubmissionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_SUBMISSION_ID];
 
 export const isRemoveJourney = (req: Request): boolean => {
-  return req.query[config.JOURNEY_QUERY_PARAM] === config.JourneyType.remove;
-};
+  const session = req.session as Session;
```
mwestacott, line 30: Need to add some tests around this new implementation. Either now or after merging into @mattch1 ' s PR.

src/utils/url.ts
mwestacott: Agreed to leave as is for now.

src/utils/url.ts
mwestacott: Tests added.

test/utils/url.spec.ts
```diff
@@ -72,15 +78,91 @@ describe("Url utils tests", () => {
...
+          is_remove: true
+        }
+      );
+
+      const result = urlUtils.isRemoveJourney(req);
+
+      expect(result).toBeTruthy();
+    });
+
+    test("returns false if is_remove is false in session data and query param journey=remove", () => {
```
dlloyd1-ch, line 150: Shouldn't isRemoveJourney() return true it the remove query param exists?

mwestacott: So this situation shouldn't arise - a conflict between what the `is_remove` value says and the `journey` query param value. But when discussed with @markpit and @mattch1 we thought that the `is_remove` value should take precedence.

makhtar-ch (approved):

test/utils/url.spec.ts
dlloyd1-ch: ok, makes sense. It would return true if the is_remove value is not present but the query param exists. I can see that's covered in another test.

dlloyd1-ch (approved):

PR #1278, opened 2024-01-03
UAR-1352 Exclusive remove pages must be inaccessible for feature flag

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1352

### Change description

Added extra check in isRemove to seek the term "remove" in the url so as to prevent access to remove pages when the FEATURE_FLAG_ENABLE_ROE_REMOVE_24112022 is false.

If we cannot access the sold land filter page and the remove is entity registered pages then we cannot access the secure update filter from remove and this would enforce the order specified on the ticket as only the update version of this page could be accessed.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -37,7 +37,7 @@ export const isRemoveJourney = (req: Request): boolean => {
     }
   }
 
-  // UAT-1352 Only required for remove feature flag - can be removed with it
+  // UAR-1352 Only required for remove feature flag - can be removed with it
   if (!req.query[config.JOURNEY_QUERY_PARAM] && req.url.includes("remove")) {
     return true;
   }
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -37,5 +37,10 @@ export const isRemoveJourney = (req: Request): boolean => {
     }
   }
 
+  // UAT-1352 Only required for remove feature flag - can be removed with it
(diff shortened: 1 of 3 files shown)


src/utils/url.ts
```diff
@@ -37,5 +37,10 @@ export const isRemoveJourney = (req: Request): boolean => {
     }
   }
 
+  // UAR-1352 Only required for remove feature flag - can be removed with it
+  if (!req.query[config.JOURNEY_QUERY_PARAM] && req.url.includes("remove")) {
```
mwestacott, line 41: I think looking for a string 'remove' in the URL is too risky - it could be present if on the update or registration journeys (e.g. if there was a page to remove a trust/BO etc).

The idea was to explore storing the 'isRemove=true' field in the session on the very first screen of the remove journey and then check for that in this method. You might want to try that.

mattch1: Would this mean getting rid of the query parameter altogether?

mwestacott: Potentially. I think in the first instance it would be good to just see if the approach would work and how much effort it is (perhaps ignoring unit-test changes to start with). Maybe on a new branch.

mattch1:
We'll need to start a new branch and investigate from there.

PR #1279, opened 2024-01-03
uar-853-remove-country-change-link-on-cya-page

makhtar-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-853

### Change description

Journey-specific modification on entity.html
When on update, remove the change link for 'country it was formed in' in both inNoChangeJourney and inChangeJourney
When on registration, keep it there.
Edited previous test not.toContain country change link

### Work checklist

### Merge instructions

```
```


mattch1 (approved):
Looks good. Just needs a second approval now.

test/controllers/update/check.your.answers.controller.spec.ts
```diff
@@ -270,7 +270,7 @@ describe("CHECK YOUR ANSWERS controller", () => {
       expect(resp.text).toContain(UPDATE_CHECK_YOUR_ANSWERS_PAGE_BENEFICIAL_OWNER_STATEMENTS_CEASED_TITLE);
       expect(resp.text).toContain(CHANGE_LINK);
       expect(resp.text).toContain(CHANGE_LINK_ENTITY_NAME);
-      expect(resp.text).toContain(CHANGE_LINK_ENTITY_INCORPORATION_COUNTRY);
+      expect(resp.text).not.toContain(CHANGE_LINK_ENTITY_INCORPORATION_COUNTRY);
```
mwestacott, line 273: Rather than checking it's not present, I think the unit test change would be to just delete this line and also the definition of the CHANGE_LINK_ENTITY_INCORPORATION_COUNTRY constant.

I know we've changed the implementation here but our tests don't exhaustively check that all the links that shouldn't be there aren't, so looks a bit odd now including this check but not others. Do you agree?

views/includes/entity.html
```diff
@@ -71,6 +71,42 @@
   %}
 {% endif %}
 
+{% if pageParams.isRegistration %}
```
mwestacott, line 74: Wondering why all this new template code was added for the registration journey as defect only talks about removing the change link for the update journey?

mattch1: That is actually the existing code, we have just replaced it with a variable so we can cover the difference with update as well.

mwestacott: 👍 thanks.

mwestacott (approved):

PR #1281, opened 2024-01-08
Uar 1321 confirm removal statement

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1321

### Change description

Template
Routing
Navigation
Controller
Validation

Post function incomplete as pages to navigate to do not yet exist.
Also navigation to the page is not yet possible as this is dependent on previous page sin the journey being completed.
### Work checklist

### Merge instructions

```
```


test/controllers/update/remove.confirm.statement.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
+jest.mock("../../../src/utils/logger");
```
markpit, line 1: can you add jest.mock("ioredis");
at the top?
not adding it causes the 'test was forced to quit after timeout' message when you run the tests

test/controllers/update/remove.confirm.statement.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+const mockGetRemove = getRemove as jest.Mock;
+
+describe("Remove confirmation statement controller", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockGetRemove.mockReturnValue({});
+  });
+
+  describe ("GET tests", () => {
+    test(`renders the ${config.REMOVE_CONFIRM_STATEMENT_PAGE}`, async () => {
```
markpit, line 38: the appData is passed into the template and the entity name and number are rendered on the screen (taken from appData). Could you put dummy values into the mock appData and check those dummy values are in the resp.text

mattch1: Created dummy values and checks fors them.

src/controllers/update/remove.confirm.statement.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+    });
+  } catch (error) {
+    next(error);
+  }
+};
+
+export const post = (req: Request, res: Response, next: NextFunction) => {
+  try {
+    logger.debugRequest(req, `POST ${config.REMOVE_CONFIRM_STATEMENT_PAGE}`);
+    return res.redirect(`${config.REMOVE_CANNOT_USE_URL}`);
```
mwestacott, line 23: Is this redirection just a temporary implementation? Until the next screen/story is done? If yes, might be worth making this clear with a TODO comment.

src/utils/registrable.beneficial.owner.ts
```diff
@@ -47,13 +48,17 @@ export const postRegistrableBeneficialOwner = async (req: Request, res: Response
     setExtraData(req.session, appData);
     await saveAndContinue(req, session, false);
 
-    if (noChangeFlag) {
-      noChangeHandler(req, res);
+    if (isRemoveJourney(req)) {
+      return res.redirect(config.REMOVE_CONFIRM_STATEMENT_URL);
     } else {
```
mwestacott, line 53: I think that the `if (noChangeFlag)` check can go on this line, with no need for an extra conditional block.

test/__mocks__/text.mock.ts
```diff
@@ -254,6 +254,7 @@ export const REMOVE_SOLD_ALL_LAND_FILTER_PAGE_TITLE = "Has the overseas entity d
 export const REMOVE_IS_ENTITY_REGISTERED_OWNER_TITLE = "Is the overseas entity currently listed on any land registry records as the registered owner of property or land in the UK?";
 export const REMOVE_INTERRUPT_CARD_TEXT = "Once an overseas entity is removed";
 export const REMOVE_USE_INFORMATION_NEED_MORE = "We’ll use this if we need more information about the application.";
+export const REMOVE_CONFIRMATION_STATEMENT_TITLE = "Confirm the removal statement - Apply to remove and overseas entity";
```
mwestacott, line 257: Text content here doesn't look right - 'and overseas entity'. Surely it should be 'an overseas entity'?

But if this is wrong and tests are passing, either it's not actually being used by a test OR it's also wrong somewhere else.

views/update/remove-confirm-statement.html
```diff
@@ -0,0 +1,49 @@
+{% extends "update-layout.html" %}
+
+{% set title = "Confirm the removal statement - Apply to remove and overseas entity" %}
```
mwestacott, line 3: Ah yes, so text looks wrong here too. Does it come from the prototype? Might need to clarify with analyst and UI team.

src/controllers/update/remove.confirm.statement.controller.ts
mattch1: Added TODO

src/utils/registrable.beneficial.owner.ts
mattch1: Rearranged this and added validation flag for not matching screen  - AC4

views/update/remove-confirm-statement.html
mattch1: As above

mattch1:
Added code for AC4 "the statement you have chosen does not match the information provided in this update"

src/utils/registrable.beneficial.owner.ts
mwestacott: Nice.

mwestacott (approved):

src/utils/registrable.beneficial.owner.ts
```diff
@@ -25,7 +26,7 @@ export const getRegistrableBeneficialOwner = (req: Request, res: Response, next:
     return res.render(templateName, {
       backLinkUrl: backLinkUrl,
       templateName: templateName,
-      appData,
+      ...appData,
```
markpit, line 29: the ...appData change here will also affect the update screen, I'm guessing the update version of this screen had been tested previously and working ok so this change might break it?

mattch1: We need this because the company name and number do not display when there are errors on the page.

markpit (approved):

PR #1289, opened 2024-01-10
UAR-1366 Fix mappings for trustees

dlloyd1-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1366

### Change description

Fix Part 1 -  Set mapping for trustee with trustee type id 5001 to be undefined rather than incorrectly map as Beneficiary. Ultimately a data fix will be required to prevent trustees with type 5001 mapped in the first place. 
Fix Part 2 - correct mapping for trustee types read from Mongo (for save and resume)

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/model/trust.model.ts
+++ b/src/model/trust.model.ts
@@ -48,7 +48,7 @@ export interface TrustReviewStatus {
 export interface TrustIndividual {
   id?: string;
   ch_references?: string;
-  type: RoleWithinTrustType;
+  type: RoleWithinTrustType | undefined;
   forename: string;
   other_forenames: string;
   surname: string;
@@ -120,7 +120,7 @@ export type TrustHistoricalBeneficialOwner =
 export type TrustCorporate = {
   id?: string
   ch_references?: string;
-  type: string;
+  type: string | undefined;
   name: string;
   date_became_interested_person_day?: string;
   date_became_interested_person_month?: string;
--- a/src/utils/update/trust.model.fetch.ts
+++ b/src/utils/update/trust.model.fetch.ts
@@ -308,7 +308,7 @@ const linkBoToTrust = (beneficialOwner: BeneficialOwnerIndividual | BeneficialOw
   beneficialOwner.trust_ids.push(trust.trust_id);
 };
 
-const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
+export const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
   switch (trusteeTypeId) {
       case "5005":
         return RoleWithinTrustType.INTERESTED_PERSON;
@@ -317,8 +317,9 @@ const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
       case "5003":
         return RoleWithinTrustType.SETTLOR;
       case "5002":
-      case "5001":
         return RoleWithinTrustType.BENEFICIARY;
+      case "5001":
+        return RoleWithinTrustType.BENEFICIAL_OWNER;
       default:
         throw new Error(`Trustee Type ${trusteeTypeId} not recognised`);
   }
--- a/src/utils/update/trust.model.fetch.ts
+++ b/src/utils/update/trust.model.fetch.ts
@@ -320,10 +320,10 @@ const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType | undefined 
         return RoleWithinTrustType.BENEFICIARY;
       case "5001":
         /*
-          Type 5001 is not a valid type for an individuual or corporate trustee. Instances of individual or corporate
-          trustees that have this rusteeTypeId are are result of incorrect data originally submitted through the
-          trust spreadsheet (which has now been superceded by web screens). In these instances we do not display the type.
-          Data fixes shoud prevent these instances from being retrieved.
+          Type 5001 is not a valid type for an individual or corporate trustee. Instances of individual or corporate
+          trustees that have this trusteeTypeId are the result of incorrect data originally submitted through the
+          trust spreadsheet (which has now been superseded by web screens). In these instances we do not display the type.
+          Data fixes should prevent these instances from being retrieved.
         */
         logger.info(`Warning - invalid data. Trustee type ${trusteeTypeId} found when mapping trustee data`);
         return undefined;
--- a/src/utils/update/trust.model.fetch.ts
+++ b/src/utils/update/trust.model.fetch.ts
@@ -319,7 +319,12 @@ const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType | undefined 
       case "5002":
         return RoleWithinTrustType.BENEFICIARY;
       case "5001":
-        // Type 5001 is not a valid type for an individuual or corporate trustee
+        /*
+          Type 5001 is not a valid type for an individuual or corporate trustee. Instances of individual or corporate
+          trustees that have this rusteeTypeId are are result of incorrect data originally submitted through the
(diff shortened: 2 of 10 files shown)


src/utils/update/trust.model.fetch.ts
```diff
@@ -308,7 +308,7 @@ const linkBoToTrust = (beneficialOwner: BeneficialOwnerIndividual | BeneficialOw
   beneficialOwner.trust_ids.push(trust.trust_id);
 };
 
-const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
+export const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
```
mwestacott, line 311: Wondering why this function needed the `export` keyword added? Can't see it being newly called from anywhere else in this PR.

test/utils/update/mocks.ts
```diff
@@ -304,7 +304,7 @@ export const FETCH_INDIVIDUAL_TRUSTEE_DATA_MOCK: IndividualTrusteeData[] = [
     dateOfBirth: "1988-12-01",
     nationality: "German",
     corporateIndicator: "N",
-    trusteeTypeId: "5001",
+    trusteeTypeId: "5002",
```
mwestacott, line 307: Wondering why the 5001 value had to change to 5002 here? There seems to already be some example data with the 5002 code (see line 265).

Also, now no example with the 5001 code in this test file - is that (or could that be) a problem?

dlloyd1-ch: I changed it to 5002 as the expected value is Beneficiary. The code was incorrectly mapping both 5001 and 5002 to Beneficiary before the code fix. The alternative is to keep it at 5001 and change the expected value to Beneficial Owner - although we are testing this mapping in other tests.

src/utils/update/trust.model.fetch.ts
dlloyd1-ch: It doesn't. I exposed the function for rudimentary testing. I thought I had reverted it.

dlloyd1-ch: export keyword removed.

views/includes/trustee-macros.html
```diff
@@ -1,13 +1,13 @@
 {% macro readableTrusteeRole(role) %}
-  {% if role === 'Beneficial_Owner' %}
+  {% if role === 'BENEFICIAL_OWNER' %}
```
mwestacott, line 2: So this addressing issue 2 in in the defect. I think PR description now needs updating.

test/utils/update/mocks.ts
dlloyd1-ch: The code was incorrectly mapping 5001 to type Beneficiary. 5002 is the correct code for Beneficiary so no need to revert.

src/utils/update/trust.model.fetch.ts
```diff
@@ -319,7 +319,9 @@ export const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType => {
       case "5002":
         return RoleWithinTrustType.BENEFICIARY;
       case "5001":
-        return RoleWithinTrustType.BENEFICIAL_OWNER;
+        // Type 5001 is not a valid type for an individuual or corporate trustee
+        logger.info(`Warning - invalid data. Trustee type ${trusteeTypeId} found when mapping trustee data`);
```
mwestacott, line 323: Guessing 'info' is the right log level. We don't seem to have a 'warning' level and 'error' might result in lots of log entries due to dodgy spreadsheet data.

Perhaps worth updating the comment to say why this might occur? Dodgy spreadsheet data...

src/model/trust.model.ts
```diff
@@ -48,7 +48,7 @@ export interface TrustReviewStatus {
 export interface TrustIndividual {
   id?: string;
   ch_references?: string;
-  type: RoleWithinTrustType;
+  type: RoleWithinTrustType | undefined;
```
markpit, line 51: Just wondering if the OE-API accepts the type if it is undefined, have we tried saving the data to mongo if the type gets set to undefined for the 5001 case.

test/utils/update/mocks.ts
mwestacott: Resolving this as 5001 no longer a valid code and test for such an invalid code is covered below.

views/includes/trustee-macros.html
```diff
@@ -1,11 +1,11 @@
 {% macro readableTrusteeRole(role) %}
-  {% if role === 'Beneficiary' %}
+  {% if role === 'BENEFICIARY' %}
```
markpit, line 2: we'll need some tests around the html changes so I can look at adding those myself

src/utils/update/trust.model.fetch.ts
markpit: i'd say info is ok, as it's an expected scenario?

mwestacott: 👍 still think a comment explaining why we're having to handle the invalid type would be useful.

test/controllers/update/update.manage.trusts.review.individuals.controller.spec.ts
```diff
@@ -132,7 +153,7 @@ describe('Update - Manage Trusts - Review individuals', () => {
 
   describe('POST tests', () => {
     test('when req body contains addIndividual (add button has been clicked), redirects to tell us about the individual with no id param', async () => {
-      mockIsActiveFeature.mockReturnValue(true);
+      mockIsActiveFeature.mockReturnValueOnce(true);
```
markpit, line 156: updated the existing mockReturnValue to the ..Once versions so as not to interfere with the mock returns in other tests in this file. There weren't any problems but have seen some in other files in the past.

views/includes/trustee-macros.html
markpit: added 
test/controllers/update/update.manage.trusts.review.legal.entities.controller.spec.ts
test/controllers/update/update.manage.trusts.review.individuals.controller.spec.ts

src/utils/update/trust.model.fetch.ts
dlloyd1-ch: Comment updated

src/model/trust.model.ts
markpit: i'll try and replicate this in docker, just wondering what will happen if we submit type as undefined to the SDK/API

dlloyd1-ch: Have tested with a full e2e web journey. The type is omitted from the record in Mongo if it is undefined. This works fine for subsequent pages (e.g Check Your Answers). 
Would also need to see if there are no issues when the data is sent to CHIPS.

src/utils/update/trust.model.fetch.ts
```diff
@@ -319,7 +319,12 @@ const mapTrusteeType = (trusteeTypeId: string): RoleWithinTrustType | undefined
       case "5002":
         return RoleWithinTrustType.BENEFICIARY;
       case "5001":
-        // Type 5001 is not a valid type for an individuual or corporate trustee
+        /*
+          Type 5001 is not a valid type for an individuual or corporate trustee. Instances of individual or corporate
```
mwestacott, line 323: Minor, but some spelling mistakes - 'individuual', 'rusteeTypeId', 'shoud'.

Thanks for adding the comment!

dlloyd1-ch: Thanks. Have installed a spell checker into my IDE on the back of this.

markpit (approved):

mwestacott (approved):

PR #1297, opened 2024-01-17
Uar 1352 make remove specific pages unreachable

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1352

### Change description

For four pages:
REMOVE_SOLD_ALL_LAND_FILTER 
REMOVE_CANNOT_USE 
REMOVE_IS_ENTITY_REGISTERED_OWNER
REMOVE_CONFIRM_STATEMENT

Moved template pages into a new remove subdirectory located under update so update/remove and added the update/remove directory to app ts.
Added a “/remove/” into the page urls after update.
Updated the isRemoveJourney function - introduce a check against the url to find the “/remove/” after the appData check and before the query parameter.

Had to update redirect from the "update continue saved filing" page to use a URL rather than a page, given that the "sold land filter" page follows and has changed its url.

### Work checklist

### Merge instructions

```
```


src/utils/url.ts
```diff
@@ -37,5 +37,9 @@ export const isRemoveJourney = (req: Request): boolean => {
     }
   }
 
+  if (req.url !== undefined && req.url.includes("/remove/")) {
```
markpit, line 40: I would make /remove/ a constant in the constants file. As we already have the REMOVE_SECTION for the url building, you could have a REMOVE_URL_IDENTIFIER = "/" + REMOVE_SECTION and use that in the if check.

mwestacott (commented):
Looks good to me. Just the comment from @markpit to address.

test/middleware/service.availability.middleware.spec.ts
```diff
@@ -109,4 +121,63 @@ describe("service availability middleware tests", () => {
...
+
+    test(`Does not reach the ${REMOVE_CANNOT_USE_PAGE} when the remove flag is false`, async () => {
+      mockIsActiveFeature
+        .mockReturnValueOnce(false) // SHOW_SERVICE_OFFLINE_PAGE
+        .mockReturnValueOnce(true) // FEATURE_FLAG_ENABLE_ROE_UPDATE
+        .mockReturnValueOnce(false); // FEATURE_FLAG_ENABLE_ROE_REMOVE
+      const resp = await request(app).get(`${REMOVE_CANNOT_USE_URL}?${PREVIOUS_PAGE_QUERY_PARAM}=${REMOVE_SOLD_ALL_LAND_FILTER_PAGE}`);
+      expect(resp.status).toEqual(200);
+      expect(resp.text).not.toContain("You cannot apply to remove this overseas entity");
+      expect(resp.text).not.toContain("Service unavailable");
```
markpit, line 147: we should also check that the correct page is displayed, so we would need a .toContain("Sorry, the service is unavailable")

test/middleware/service.availability.middleware.spec.ts
```diff
@@ -109,4 +121,63 @@ describe("service availability middleware tests", () => {
...
+      jest.resetAllMocks();
+    });
+
+    test(`Redirects to the next page whenever the remove flag is true`, async () => {
+      mockIsActiveFeature
+        .mockReturnValueOnce(false) // SHOW_SERVICE_OFFLINE_PAGE
+        .mockReturnValueOnce(true) // FEATURE_FLAG_ENABLE_ROE_UPDATE
+        .mockReturnValueOnce(true); // FEATURE_FLAG_ENABLE_ROE_REMOVE
+      const resp = await request(app).get(`${REMOVE_CANNOT_USE_URL}?${PREVIOUS_PAGE_QUERY_PARAM}=${REMOVE_SOLD_ALL_LAND_FILTER_PAGE}`);
+      expect(resp.status).toEqual(302);
```
markpit, line 136: I don't think it's enough just to check for any redirect, we should also check it is actually going to the next page, so would need a header.location check (or whatever other tests do to check it redirects to correct page)

test/middleware/service.availability.middleware.spec.ts
markpit: same comment applies to the rest of the 'Does not reach' tests

src/utils/url.ts
mattch1: Updated

test/middleware/service.availability.middleware.spec.ts
mattch1: Added a further check.

test/middleware/service.availability.middleware.spec.ts
mattch1: That works with the controller tests but this is a middleware test so I'm limited as to what I can check for, I have added what I can here.

markpit: it's redirecting to the sign-in page, and sending user to the start of journey as they aren't logged in so just check header location is
/signin?return_to=/update-an-overseas-entity/secure-update-filter

markpit: ignore that, the check is already there, i missed it

test/middleware/service.availability.middleware.spec.ts
```diff
@@ -109,4 +122,64 @@ describe("service availability middleware tests", () => {
...
+    });
+
+    test(`Does not reach the ${REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE} when the remove flag is false`, async () => {
+      mockIsActiveFeature
+        .mockReturnValueOnce(false) // SHOW_SERVICE_OFFLINE_PAGE
+        .mockReturnValueOnce(true) // FEATURE_FLAG_ENABLE_ROE_UPDATE
+        .mockReturnValueOnce(false); // FEATURE_FLAG_ENABLE_ROE_REMOVE
+      const resp = await request(app).get(`${REMOVE_IS_ENTITY_REGISTERED_OWNER_URL}?${PREVIOUS_PAGE_QUERY_PARAM}=${REMOVE_SOLD_ALL_LAND_FILTER_PAGE}`);
+      expect(resp.status).toEqual(200);
+      expect(resp.text).not.toContain("You cannot apply to remove this overseas entity");
```
markpit, line 170: the text "You cannot apply to remove this overseas entity" should only be on the REMOVE_CANNOT_USE_PAGE. Each of the 'Does not reach" tests would need .not.contain. text from the page you are requesting, eg REMOVE_IS_ENTITY_REGISTERED_OWNER_PAGE in this test.

test/middleware/service.availability.middleware.spec.ts
mattch1: Not sure what happed there must have backed them out. They should be present now.

markpit (approved):

PR #1307, opened 2024-01-26
New remove journey middleware to set the page banner

mwestacott:
### JIRA link

None

### Change description

A PoC just to see if this will work. Looking good (remove banner displays, also after validation failures), although question around the changes required for unit tests.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/middleware/navigation/remove/remove.journey.middleware.ts
+++ b/src/middleware/navigation/remove/remove.journey.middleware.ts
@@ -1,0 +1,21 @@
+/**
+ * Middleware to check if this request is for a Remove journey and if so set this as the
+ * journey type (which can then be used by page templates to set the banner etc.).
+ */
+import { NextFunction, Request, Response } from "express";
+
+import {
+  JourneyType
+} from "../../../config";
+import { logger } from "../../../utils/logger";
+import { isRemoveJourney } from "../../../utils/url";
+
+export const removeJourneyMiddleware = (req: Request, res: Response, next: NextFunction) => {
+
+  if (isRemoveJourney(req)) {
+    logger.infoRequest(req, "Marking this request/response as a Remove Journey");
+    res.locals.journey = JourneyType.remove;
+  }
+
+  return next();
+};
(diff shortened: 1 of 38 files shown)


test/controllers/secure.register.filter.controller.spec.ts
```diff
@@ -43,6 +45,9 @@ const mockSetExtraData = setExtraData as jest.Mock;
 const mockServiceAvailabilityMiddleware = serviceAvailabilityMiddleware as jest.Mock;
 mockServiceAvailabilityMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => next() );
 
+const mockRemoveJourneyMiddleware = removeJourneyMiddleware as jest.Mock;
```
mwestacott, line 48: Tested locally. Seems to work, but quite a few tests fail.

Would need to add this to each failing test or is there some way to define a general mock implementation across tests @markpit ?

markpit: yes, you can create the mock in a separate file and import it

markpit: eg https://github.com/companieshouse/confirmation-statement-web/blob/main/test/controllers/create.transaction.controller.unit.ts#L5

mwestacott: Thanks. Looks simple enough. But we would then need to add the import statement to every (failing) test file, yes?

markpit: yes, you'd still need to import it

src/middleware/navigation/remove/remove.journey.middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import {
+  JourneyType
+} from "../../../config";
+import { logger } from "../../../utils/logger";
+import { isRemoveJourney } from "../../../utils/url";
+
+export const removeJourneyMiddleware = (req: Request, res: Response, next: NextFunction) => {
+
+  if (isRemoveJourney(req)) {
```
mattch1, line 15: Good idea this makes the remove test more explicit, are there plans to delete the existing isRemoveJourney from url.ts?

mwestacott: No, that function will still be needed as called in a number of places (mainly controller code).

PR #1314, opened 2024-01-30
Uar 1286 review-update-statement page for remove

makhtar-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1286

### Change description

As a ROE Remove user I would like to be given the option to review my overseas entity’s information before I submit my application for removal.

- Set title for browser tab/window
- Set 'remove' heading
- Set h1 page title to 'Review the information before submitting' if on remove.
- Set 'Who can we contact about this application?' if on remove.
- Hide all changeLinks included from presenter.html and review-update-statement page if on remove.
- Set hasSoldAllLand and IsListedAsPropertyOwner statement values from appData.
- Add "Has the overseas entity disposed of all its property or land in the UK?" statement if on remove
- Add "Is the overseas entity currently listed on any land registry records as the registered owner of property or land in the UK?" for remove.
- Set "Is all the information in this application correct?" if on remove.
- "Do you need to make any changes to this overseas entity?" in update is in a different location to remove. We hid the current one for remove and added a new section further down as per prototype.
- Add "All information submitted must be correct on" if on remove.
- Created date function (date.ts) using luxon.
- Error validation.
- print button.
- backlink and navigation from "No, I need to make changes"
- Bunch of mocking and tests for the above.
- Tested both journeys
- Remove journey gets to payment screen with correct amount displayed.

### Work checklist


src/utils/date.ts
```diff
@@ -0,0 +1,16 @@
+import { DateTime } from "luxon";
+
+export interface DateAsStrings {
```
mwestacott, line 3: Rather than define a new structure here, could you maybe (re-) use the `InputDate` definition in data.types.model.ts?

src/utils/date.ts
```diff
@@ -0,0 +1,16 @@
...
+    day: string,
+    month: string,
+    year: string
+}
+
+export const getTodaysDate = (): DateAsStrings => {
+  const now = DateTime.now().toUTC();
+  return {
+    day: "" + now.day,
```
mwestacott, line 12: With regard to that other structure (InputDate), I've seen other functions that pad the values assigned - e.g. "01" instead of "1" for the first day of the month. Is that relevant here?

src/validation/error.messages.ts
```diff
@@ -197,6 +197,7 @@ export enum ErrorMessages {
   SELECT_IF_ANY_TRUSTS_INVOLVED = "Select yes if there are any trusts involved in this overseas entity",
   SELECT_DO_YOU_WANT_TO_MAKE_OE_CHANGE = "Select yes if you need to make any changes to this overseas entity",
   SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_UPDATE_STATEMENT = "Select yes if all the information in this update statement is correct",
+  SELECT_DO_YOU_WANT_TO_MAKE_CHANGES_REMOVE_STATEMENT = "Select yes if all the information in this application correct",
```
mwestacott, line 200: Guessing this text relates to AC3 but I can't see this actual text on [this](https://companieshouse.atlassian.net/wiki/spaces/ROE2/pages/4334977098/Error+messages) page and if it is correct, it looks like there's an "is" missing from the near the end.

test/utils/navigation.spec.ts
```diff
@@ -76,6 +77,17 @@ describe("NAVIGATION utils", () => {
     expect(backLink).toEqual(config.UPDATE_FILING_DATE_URL);
   });
 
+  test(`getUpdateReviewStatementBackLink returns the correct URL when not on the Remove journey`, () => {
```
mwestacott, line 80: Very minor, but I would capitalise the word NOT in the description. Just makes it stand out from the test below.

views/update/review-update-statement.html
```diff
@@ -127,10 +175,34 @@ <h1 class="govuk-heading-xl">Review the information before submitting this updat
...
+              "data-event-id": "nochange-changelink-entity-review-statement-button"
+            }
+          }
+        ]
+      }) }}
+      <br>
+    {% endif %}
+
+    {% if isRemoveJourney %}
+      {% set hidePresenterChangeLink = true %}
```
mwestacott, line 204: Just wondering why you did this, rather than let presenter.html test against the `isRemoveJourney` flag directly?

views/update/review-update-statement.html
```diff
@@ -27,6 +51,10 @@
 
 {% set updateStatementDate = dateMacros.formatDate(appData.update.filing_date["day"], appData.update.filing_date["month"], appData.update.filing_date["year"]) %}
 
+{% if isRemoveJourney %}
+  {% set removeStatementDate = dateMacros.formatDate(pageParams.today["day"], pageParams.today["month"], pageParams.today["year"]) %}
```
mwestacott, line 55: Maybe this does the padding and my previous question is not relevant?

src/utils/date.ts
markpit: yes, didn't spot that, looks the same so could re-use it

src/utils/date.ts
markpit: the template uses a macro which pads and converts month to eg "January' so we just needed to get the numeric values for todays date into the template

test/utils/navigation.spec.ts
makhtar-ch: We kept it lowercase to maintain consistency with the other 3 lowercase "not" words in the same file.

src/validation/error.messages.ts
markpit: one of the messages was missing from the guide, so we changed it based on the other text changes, eg, converting 'update statement' to 'application'.

The 'is' should be added so we can fix that.

views/update/review-update-statement.html
markpit: yes, the macro converts the numerics into a padded sentence - eg 05 January 2024

views/update/review-update-statement.html
markpit: There might be other situations where change links might not want to be shown so figured it more relevant to have a hidePresenterChangeLink value. As the presenter.html is a re-usable 'chunk' , by having a hidePresenterChangeLink, we can let whatever uses it decide if change links to be shown or not.

src/utils/date.ts
mwestacott: Ok. Nice.

src/validation/error.messages.ts
mwestacott: Ok. I think we need to get AC updated though (otherwise this might come back as a defect/tester query).

mwestacott (approved):

dlloyd1-ch (approved):

PR #1345, opened 2024-03-13
UAR-1430 add trust status to trusts associated page

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1430

### Change description
Adding the word "Active" or "Removed" against each trust on the trust summary page, but they will only show when the feature flag for the cease trusts is enabled.
"Removed" will be shown if all parts of the trust ceased date are present (day, month, year)

### Work checklist

### Merge instructions

```
```


test/utils/test.utils.ts
```diff
@@ -0,0 +1,6 @@
+export function wordCount(wordToCount: string, inText: string): number {
+  const regex = new RegExp(`\\b${wordToCount}\\b`, 'gi');
+  const matches = inText.match(regex);
+  const wordCount = matches ? matches.length : 0;
+  return wordCount;
+}
```
markpit, line 6: I wanted to be able to count occurrences of a word in text in the tests, so created this test.utils.ts file for this and other 'helper' test functions

views/includes/page-templates/add-trust.html
```diff
@@ -1,3 +1,11 @@
+{% macro trustStatus(trust) %}
+  {% if trust.ceased_date_day and trust.ceased_date_month and trust.ceased_date_year %}
+    Removed
+  {% else %}
+    Active
+  {% endif %}
+{% endmacro %}
+
```
markpit, line 8: the macro is essentially a nunjucks function that can be called from within the template

src/utils/add.trust.ts
```diff
@@ -53,6 +54,7 @@ const getPageProperties = (
       subtitle: ADD_TRUST_TEXTS.subtitle,
     },
     isUpdate,
+    isFeatureFlagCeaseTrustsEnabled: isActiveFeature(config.FEATURE_FLAG_ENABLE_CEASE_TRUSTS),
```
mattch1, line 57: Should this be "FEATURE_FLAG_ENABLE_CEASED_TRUSTS"?

markpit: I think it used to be called that but has now been renamed to FEATURE_FLAG_ENABLE_CEASE_TRUSTS after previous comments

mwestacott: That sounds wrong. Was that one of my comments?

mwestacott: I think it was something like FEATURE_FLAG_ENABLE_TRUST_CEASED_DATE but I suggested a new name that was more related to the action of 'ceasing' a trust.

mwestacott, line 57: Do we need this? The template could in theory get hold of the feature flag value by doing something like this:

`{% if FEATURE_FLAG_ENABLE_CEASE_TRUSTS %}`

And then set a local template variable, if required.

src/utils/add.trust.ts
```diff
@@ -31,6 +31,7 @@ type TrustInvolvedPageProperties = {
   formData?: PageModel.AddTrust,
   errors?: FormattedValidationErrors,
   url: string,
+  isFeatureFlagCeaseTrustsEnabled: boolean
```
mwestacott, line 34: Not sure why this is in the data structure, but question might be answered by response to [this](https://github.com/companieshouse/overseas-entities-web/pull/1345/files#r1523493973) comment.

markpit: it's because i'm passing the feature flag to the template and that is the object type that is passed so needs to be added there or you get errors

mwestacott: Ok. All clear.

mwestacott (approved):

dlloyd1-ch (approved):

PR #1347, opened 2024-03-20
UAR-1425 filter out ceased bos on trust details screen

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1425

### Change description
Needed to exclude ceased BOs from the list of BOs that are assignable to trusts, so they don't appear in the checkbox list on the trust details screen

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/utils/trusts.ts
+++ b/src/utils/trusts.ts
@@ -158,14 +158,16 @@ const getBoIndividualAssignableToTrust = (
   appData: ApplicationData,
 ): BeneficialOwnerIndividual[] => {
   return (appData[BeneficialOwnerIndividualKey] ?? [])
-    .filter((bo: BeneficialOwnerOther) => bo.trustees_nature_of_control_types?.length);
+    .filter((bo: BeneficialOwnerIndividual) => bo.trustees_nature_of_control_types?.length)
+    .filter((bo: BeneficialOwnerIndividual) => Object.keys(bo.ceased_date || {}).length === 0);
 };
 
 const getBoOtherAssignableToTrust = (
   appData: ApplicationData,
 ): BeneficialOwnerOther[] => {
   return (appData[BeneficialOwnerOtherKey] ?? [])
-    .filter((bo: BeneficialOwnerOther) => bo.trustees_nature_of_control_types?.length);
+    .filter((bo: BeneficialOwnerOther) => bo.trustees_nature_of_control_types?.length)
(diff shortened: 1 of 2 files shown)


src/utils/trusts.ts
```diff
@@ -158,14 +158,16 @@ const getBoIndividualAssignableToTrust = (
   appData: ApplicationData,
 ): BeneficialOwnerIndividual[] => {
   return (appData[BeneficialOwnerIndividualKey] ?? [])
-    .filter((bo: BeneficialOwnerOther) => bo.trustees_nature_of_control_types?.length);
+    .filter((bo: BeneficialOwnerIndividual) => bo.trustees_nature_of_control_types?.length)
+    .filter((bo: BeneficialOwnerIndividual) => Object.keys(bo.ceased_date || {}).length === 0);
```
mattch1, line 162: Can you clarify what the **"|| {}"** adds? Is it just covering all non null fields even if they are empty?

markpit: it's to prevent an error if ceased_date is undefined. If we call .length on an undefined object we'll get the equivalent of a null pointer exception/error. So the || will mean that {} (empty object) gets returned/used if ceased_date is undefined. it's like saying use bo.ceased date if not undefined else use {}.

markpit: i basically took it from the kainos code that decides whether to show the ceased date on the screen or not for BOs

test/utils/trust.spec.ts
```diff
@@ -73,6 +73,19 @@ describe('Trust Utils method tests', () => {
     id: '9002',
   } as BeneficialOwnerIndividual;
 
+  const mockBoIndividualCeasedWithTrusteeNoc = {
```
mwestacott, line 76: `mockBoCeasedIndividualWithTrusteeNoc` perhaps? (if that's what it is)

test/utils/trust.spec.ts
```diff
@@ -85,6 +98,19 @@ describe('Trust Utils method tests', () => {
     ],
   } as BeneficialOwnerOther;
 
+  const mockBoOleCeasedWithTrusteeNoc = {
```
mwestacott, line 101: I see it's already present in the file, but what's an 'Ole'?

markpit: other legal entitiy

mwestacott (approved):

PR #1350, opened 2024-03-26
UAR-1354 sign out validation and back link fix fir remove journey

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1354

### Change description
- Fix to make the sign-out 'save' button appear for remove journey
- Fix the sign-out link to make it work for remove journey urls that have more slashes eg, .../remove/somepage
- Use the Registration approach to get the previous page from the request headers and use that for the back link / No button on the sign-out screen

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/update.sign.out.controller.ts
+++ b/src/controllers/update/update.sign.out.controller.ts
@@ -10,6 +10,8 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
 
   if (isRemoveJourney(req)) {
     journey = config.JourneyType.remove;
+    // some sign-out back links don't work if just using the template name, some need other path structures and query params
+    // so use the same apprach as the Registration journey and get the previous page url from the request headers
     previousPage = getPreviousPageUrl(req, config.UPDATE_AN_OVERSEAS_ENTITY_URL);
   }
 
--- a/src/controllers/update/update.sign.out.controller.ts
+++ b/src/controllers/update/update.sign.out.controller.ts
@@ -11,7 +11,7 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
   if (isRemoveJourney(req)) {
     journey = config.JourneyType.remove;
     // some sign-out back links don't work if just using the template name, some need other path structures and query params
-    // so use the same apprach as the Registration journey and get the previous page url from the request headers
+    // so use the same approach as the Registration journey and get the previous page url from the request headers
     previousPage = getPreviousPageUrl(req, config.UPDATE_AN_OVERSEAS_ENTITY_URL);
   }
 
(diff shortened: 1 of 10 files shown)


views/sign-out.html
```diff
@@ -21,7 +21,7 @@
 
 {% if saveAndResume and (journey == "register") %}
   {% set hintText = "We will save your application." %}
-{% elif saveAndResume and (journey == "update") %}
+{% elif saveAndResume and (journey == "update" or journey == "remove") %}
```
mwestacott, line 24: Is it maybe worth defining an `isUpdateOrRemoveJourney` variable at the top of this file to shorten this repeating logic check in the four times it's done? Or could you just do `journey != "register"` ? Of course, logic will further simplify if/when the non 'save and resume' code is ever ripped out...

views/includes/sign-out-user-banner.html
```diff
@@ -1,8 +1,17 @@
 {% if userEmail %}
+
+  {% set signOutPath = 'sign-out?page=' + templateName %}
+
+  {% if journey === 'remove' %}
+    {% set url = url if url else OE_CONFIGS.UPDATE_AN_OVERSEAS_ENTITY_URL %}
```
mwestacott, line 6: This just looks a bit confusing - setting 'update' as the url if we're on a remove journey. I assume this is handling reuse of the update pages in the remove journey, but worth a comment I think, explaining this for future readers of this code (or me, in a week's time when I've forgotten most things apart from my name).

src/controllers/update/update.sign.out.controller.ts
```diff
@@ -2,15 +2,24 @@ import { NextFunction, Request, Response } from "express";
...
+import { getPreviousPageUrl, isRemoveJourney } from "../../utils/url";
 
 export const get = (req: Request, res: Response, next: NextFunction) => {
+  let journey = config.JourneyType.update;
+  let previousPage = `${config.UPDATE_AN_OVERSEAS_ENTITY_URL}${req.query["page"]}`;
+
+  if (isRemoveJourney(req)) {
+    journey = config.JourneyType.remove;
+    previousPage = getPreviousPageUrl(req, config.UPDATE_AN_OVERSEAS_ENTITY_URL);
+  }
```
dlloyd1-ch, line 14: Might be worth a comment explaining why a remove journey needs to call the getPreviousPageUrl function whereas the update journey does not. Or maybe a comment on the function itself.

markpit: added a comment to say why we use that approach

dlloyd1-ch (approved):

PR #1351, opened 2024-03-27
UAR-1468 Fix isRemoveJourney logic to handle >1 journey url query params

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1468

### Change description
If there were > 1 journey=remove params in the url query, and no /remove/ in the url then isRemoveJourney() function would return false, which is incorrect.
Changing the function to handle multiple journey params (if there are multiple they get stored as comma separated values under 'journey' in the req.query object)
The function will now return true if there is at least 1 journey=remove in the url query params 

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/test/utils/url.spec.ts
+++ b/test/utils/url.spec.ts
@@ -6,8 +6,10 @@ import * as config from "../../src/config";
 import * as urlUtils from "../../src/utils/url";
 import { getApplicationData } from '../../src/utils/application.data';
 import { APPLICATION_DATA_MOCK } from "../__mocks__/session.mock";
+import { createAndLogErrorRequest } from "../../src/utils/logger";
 
 const mockGetApplicationData = getApplicationData as jest.Mock;
+const mockCreateAndLogErrorRequest = createAndLogErrorRequest as jest.Mock;
 
 describe("Url utils tests", () => {
   const req = request;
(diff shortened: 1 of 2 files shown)


test/utils/url.spec.ts
```diff
@@ -80,11 +80,15 @@ describe("Url utils tests", () => {
 
   describe("isRemoveJourney tests", () => {
 
-    test("returns true if app data not present in session and query param journey=remove", () => {
+    test.each([
+      ['journey=remove', 'remove'],
+      ['journey=remove&journey=remove', 'remove,remove'],
+      ['journey=remove&journey=update', 'remove,update']
```
mwestacott, line 86: This test case raises the question what _should_ the `isRemoveJourney()` function return in this scenario?

Probably not **true**, as if two different values are set we can't be sure what journey the user is on. I'd have thought it should throw an error.

markpit: It could do, i also thought in that case isRemoveJourney and isUpdateJourney functions could both return true. Perhaps the error is the better option.

mwestacott: I think an error is better. You can't be on both journeys at the same time.

mwestacott (approved):

PR #1365, opened 2024-04-24
UAR-1470-Amend 'Has the overseas entity disposed of all its property or land in the UK?' page

makhtar-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1470

### Change description

The title and text has been amended.
A new dropdown has been added.
Tests for the dropdown text.

### Work checklist

### DIFF
--- a/test/__mocks__/text.mock.ts
+++ b/test/__mocks__/text.mock.ts
@@ -253,6 +253,7 @@ export const UPDATE_MANAGE_TRUSTS_REVIEWED_HEADING = "What you have reviewed";
 
 // Remove journey
 export const REMOVE_SOLD_ALL_LAND_FILTER_PAGE_TITLE = "Has the overseas entity disposed of all its property or land in the UK?";
+export const REMOVE_NEW_SOLD_ALL_LAND_FILTER_PAGE_TITLE = "Has the overseas entity disposed of all relevant property or land in the UK?";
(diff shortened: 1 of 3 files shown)


test/__mocks__/text.mock.ts
```diff
@@ -253,6 +253,7 @@ export const UPDATE_MANAGE_TRUSTS_REVIEWED_HEADING = "What you have reviewed";
 
 // Remove journey
 export const REMOVE_SOLD_ALL_LAND_FILTER_PAGE_TITLE = "Has the overseas entity disposed of all its property or land in the UK?";
```
markpit, line 255: is REMOVE_SOLD_ALL_LAND_FILTER_PAGE_TITLE still used anywhere? if not, it can be removed and the "NEW" one can be renamed to remove the "NEW" from it.

markpit: looks like those 2 pages are still showing the old page title, perhaps they need updating too. I would check with Vernon if they should be updated as well.

makhtar-ch: They do, but yesterday in refinement was advised to leave the additional work and it'll be placed in the backlog.

makhtar-ch: Another ticket has been created to amend the const that is used on both update.review.statement and check.your.answers

markpit (approved):

PR #1375, opened 2024-04-30
UAR-1504 Add info for update statement date to update-filing-date page

makhtar-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1504

### Change description

Added an inset text component from GOV.UK components https://design-system.service.gov.uk/components/inset-text/
Used config.CHS_URL instead of hardcoding the URL.
Added tests to check for some text and a url.

Screenshot 2024-04-30 at 16 05 33

### Work checklist

### DIFF
--- a/src/controllers/update/update.filing.date.controller.ts
+++ b/src/controllers/update/update.filing.date.controller.ts
@@ -21,7 +21,7 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     return res.render(config.UPDATE_FILING_DATE_PAGE, {
       backLinkUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.UPDATE_FILING_DATE_PAGE,
-      chsUrl: process.env.CHS_URL,
+      chsUrl: config.CHS_URL,
       ...appData,
       [FilingDateKey]: filingDate
     });
--- a/src/controllers/update/update.filing.date.controller.ts
+++ b/src/controllers/update/update.filing.date.controller.ts
@@ -21,6 +21,7 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     return res.render(config.UPDATE_FILING_DATE_PAGE, {
       backLinkUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.UPDATE_FILING_DATE_PAGE,
+      chsUrl: process.env.CHS_URL,
       ...appData,
       [FilingDateKey]: filingDate
     });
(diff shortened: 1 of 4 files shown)


src/controllers/update/update.filing.date.controller.ts
```diff
@@ -21,6 +21,7 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
     return res.render(config.UPDATE_FILING_DATE_PAGE, {
       backLinkUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.UPDATE_FILING_DATE_PAGE,
+      chsUrl: process.env.CHS_URL,
```
markpit, line 24: the CHS_URL is already in the config file src/config/index.ts so you should be able to import it and use 
chsUrl: config.CHS_URL

test/controllers/update/update.filing.date.controller.spec.ts
```diff
@@ -93,6 +94,8 @@ describe("Update Filing Date controller", () => {
       expect(resp.text).toContain(BACK_LINK_FOR_UPDATE_FILING_DATE);
       expect(resp.text).toContain(saveAndContinueButtonText);
       expect(resp.text).not.toContain(PAGE_TITLE_ERROR);
+      expect(resp.text).not.toContain(UPDATE_DATE_OF_UPDATE_STATEMENT_TEXT);
```
dlloyd1-ch, line 97: Would have thought that we should be checking that the response text **does** contain the new test? Same with the other tests.

views/update/update-filing-date.html
```diff
@@ -20,7 +20,13 @@ <h1 class="govuk-heading-xl">{{ title }}</h1>
     <form method="post">
       <p class="govuk-body">All the information in this update statement must be correct on this date.
       Verification checks must be completed no more than 3 months before the statement date.</p>
-      {% set filingDateLabelText = "What is the date of the update statement?" %}
+
+      <div class="govuk-inset-text">
+        <p>The date you enter <strong> must be on or before the overseas entity's update statement date, </strong> or your filing may be rejected.</p>
```
dlloyd1-ch, line 25: Having a space before and after the '<strong>' markup would probably result in have two spaces between the words. What does it look like locally?

makhtar-ch: Initially I had it as just the single space, but after a quick look it seems that HTML multiple spaces are collapsed in to one single space. I have checked locally and it shows as a single space. I did this for readability but can change it to a single space?

makhtar-ch: For consistency, I have reverted it to single space.

markpit (approved):

PR #1376, opened 2024-05-01
UAR-1502 Make 'property name or number' field optional

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1502

### Change description

* data entry in this field is now optional for the Update and Remove journeys
* a value must still be entered if on the Registration journey

Change is for a spike - no unit-tests added

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/custom.validation.ts
+++ b/src/validation/custom.validation.ts
@@ -630,6 +630,29 @@ export const validateEmail = (email: string, maxLength: number) => {
   return true;
 };
 
+export const checkPropertyNameOrNumberIfRadioButtonSelected = (selected: boolean, propertyNameOrNumber: string, req) => {
+  if (selected) {
+    return checkPresenceOfPropertyNameOrNumber(propertyNameOrNumber, req);
+  }
+
+  return true;
+};
+
+export const checkPresenceOfPropertyNameOrNumber = (propertyNameOrNumber: string, req) => {
+  if (propertyNameOrNumber === undefined || propertyNameOrNumber === "") {
+    const appData: ApplicationData = getApplicationData(req.session);
+
+    // If this is a Registration journey, a property name or number must be entered, if not,
+    // a validation error must be shown to the user (the field is optional on the Update and Remove
+    // journeys, as value is moved into the 'Address Line 1' field)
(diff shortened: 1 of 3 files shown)


src/validation/custom.validation.ts
```diff
@@ -630,6 +630,29 @@ export const validateEmail = (email: string, maxLength: number) => {
...
+export const checkPropertyNameOrNumberIfRadioButtonSelected = (selected: boolean, propertyNameOrNumber: string, req) => {
+  if (selected) {
+    return checkPresenceOfPropertyNameOrNumber(propertyNameOrNumber, req);
+  }
+
+  return true;
+};
+
+export const checkPresenceOfPropertyNameOrNumber = (propertyNameOrNumber: string, req) => {
+  if (propertyNameOrNumber === undefined || propertyNameOrNumber === "") {
```
mwestacott, line 642: I copied the approach taken for the 'email' field here, but if we do ever implement this, we may wish to `trim` the value too before testing it.

src/validation/fields/address.validation.ts
```diff
@@ -3,14 +3,45 @@ import { defaultRequiredErrorMessages, defaultOptionalErrorMessages, ErrorMessag
...
+  checkMaxFieldIfRadioButtonSelected,
+  checkPresenceOfPropertyNameOrNumber
 } from "../custom.validation";
 import { ErrorMessages } from "../error.messages";
 import { VALID_CHARACTERS } from "../regex/regex.validation";
 
+export const overseas_entity_address_validations = (errors: ErrorMessagesOptional = defaultOptionalErrorMessages) => {
+  errors = { ...defaultOptionalErrorMessages, ...errors };
+  return [
+    body("principal_address_property_name_number")
```
markpit, line 19: seeing as this returns an a list of validators, some might be shared with the other lists, could we have a common list and use ... to import the common validations into the list mixed with the specific validations for that field/page, or wold that mean refactoring all the address validations?

mwestacott: Don't know, but a sensible suggestion. We should investigate if this becomes a 'real' changes.

PR #1379, opened 2024-05-03
ROECCT-115 Important Banner

kullas-ch:
### JIRA link
https://companieshouse.atlassian.net/browse/ROECCT-115

### Work checklist

### Merge instructions

```
```


views/includes/page-templates/important-banner.html
```diff
@@ -0,0 +1,10 @@
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
+{% set html %}
+
+You need to provide some information about the overseas entity from the pre-registration period <br></br>
+{% include "includes/page-templates/relevant-period.html" %}
```
ConnorOS-ch, line 5: could we enclose this text in a `<p>` element. You wont need to use `<br>` in that case also

views/includes/page-templates/relevant-period.html
```diff
@@ -0,0 +1,6 @@
+{% set start_date = "28 February 2022" %}
+{% set registration_date = dateMacros.formatDate(dateOfCreation["day"], dateOfCreation["month"], dateOfCreation["year"]) %}
+
+<p class="govuk-body-m">
```
ConnorOS-ch, line 4: I don't think `class="govuk-body-m"` is doing anything so safe to remove

amalviyach (approved):

views/includes/page-templates/important-banner.html
ConnorOS-ch, line 1: I don't know is includes/page-templates is the correct place to put this file and relevant-period.html

maybe just includes is better

ConnorOS-ch (changes requested):
A few stylistic comments. Looks great though

views/includes/page-templates/important-banner.html
mzacharias-ch: it is in correct path it is a page template only and not a whole page.

mzacharias-ch (approved):

views/includes/page-templates/relevant-period.html
mzacharias-ch: that class is doing margin on bottom 20px just like is in describe in prototype. Please check prototype and you will see they are using this class.

ConnorOS-ch (approved):

views/includes/page-templates/important-banner.html
ConnorOS-ch: Page-templates is used when a whole or most of a page content is reused between similar pages.

This is an individual component not a page template it isn't similar in use to the files already in page-templates

PR #1384, opened 2024-05-13
UAR-1505 update statement date next mud date checks

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1505

### Change description
- Pre-populate the update statement date with the next made up to date obtained from the confirmation statement details
- Add validation to ensure user can only enter a date before or on the the next made up to date.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/update.filing.date.controller.ts
+++ b/src/controllers/update/update.filing.date.controller.ts
@@ -63,7 +63,10 @@ const getFilingDate = async (req: Request, appData: ApplicationData): Promise<{}
   let filingDate = appData.update?.[FilingDateKey] ? mapDataObjectToFields(appData.update[FilingDateKey], FilingDateKeys, InputDateKeys) : undefined;
 
   // otherwise use the next made up to date from confirmation statement in company profile
-  if (!filingDate && appData.entity_number) {
+  if (!filingDate) {
+    if (!appData.entity_number) {
+      throw createAndLogErrorRequest(req, `update.filing.controller unable to find entity_number in application data for entity_name ${appData.entity_name}`);
+    }
     logger.debugRequest(req, `Getting confirmation statement next made up to date for entity number = ${appData.entity_number}`);
     const nextMudIsoString = await getConfirmationStatementNextMadeUpToDateAsIsoString(req, appData.entity_number);
     if (!nextMudIsoString) {
--- a/src/controllers/update/update.filing.date.controller.ts
+++ b/src/controllers/update/update.filing.date.controller.ts
@@ -1,7 +1,7 @@
 import { NextFunction, Request, Response } from "express";
 import { Session } from "@companieshouse/node-session-handler";
 
-import { logger } from "../../utils/logger";
+import { createAndLogErrorRequest, logger } from "../../utils/logger";
 import * as config from "../../config";
 import { isActiveFeature } from "../../utils/feature.flag";
 import { getApplicationData, setExtraData, mapFieldsToDataObject, mapDataObjectToFields } from "../../utils/application.data";
@@ -10,21 +10,24 @@ import { createOverseasEntity, updateOverseasEntity } from "../../service/overse
 import { OverseasEntityKey, Transactionkey, InputDateKeys } from '../../model/data.types.model';
 import { FilingDateKey, FilingDateKeys } from '../../model/date.model';
 import { ApplicationData } from "../../model/application.model";
+import { getConfirmationStatementNextMadeUpToDate } from "../../service/company.profile.service";
+import { convertIsoDateToInputDate } from "../../utils/date";
 
-export const get = (req: Request, res: Response, next: NextFunction) => {
+export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
     const appData = getApplicationData(req.session);
-    const filingDate = appData.update?.[FilingDateKey] ? mapDataObjectToFields(appData.update[FilingDateKey], FilingDateKeys, InputDateKeys) : {};
 
     return res.render(config.UPDATE_FILING_DATE_PAGE, {
       backLinkUrl: config.UPDATE_OVERSEAS_ENTITY_CONFIRM_URL,
       templateName: config.UPDATE_FILING_DATE_PAGE,
       ...appData,
-      [FilingDateKey]: filingDate
+      [FilingDateKey]: await getFilingDate(req, appData)
     });
   } catch (error) {
+    console.log("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
+    console.log(JSON.stringify(error));
     logger.errorRequest(req, error);
     next(error);
   }
@@ -55,4 +58,21 @@ export const post = async(req: Request, res: Response, next: NextFunction) => {
     logger.errorRequest(req, error);
     next(error);
   }
+};
+
+const getFilingDate = async (req: Request, appData: ApplicationData): Promise<{} | undefined> => {
+  // use date stored in appData if present
+  let filingDate = appData.update?.[FilingDateKey] ? mapDataObjectToFields(appData.update[FilingDateKey], FilingDateKeys, InputDateKeys) : undefined;
+
+  // otherwise use the next made up to date from confirmation statement in company profile
+  if (!filingDate && appData.entity_number) {
+    logger.debugRequest(req, `Getting confirmation statement next made up to date for entity number = ${appData.entity_number}`);
+    const nextMudIsoString = await getConfirmationStatementNextMadeUpToDate(req, appData.entity_number);
+    if (!nextMudIsoString) {
+      throw createAndLogErrorRequest(req, `No confirmation statement next made up to date found on company profile for entity number = ${appData.entity_number}; transaction id = ${appData.transaction_id}`);
+    }
+    const nextMudAsInputDate = convertIsoDateToInputDate(nextMudIsoString);
+    filingDate = mapDataObjectToFields(nextMudAsInputDate, FilingDateKeys, InputDateKeys);
+  }
+  return filingDate;
 };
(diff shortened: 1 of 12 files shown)


src/validation/error.messages.ts
```diff
@@ -176,6 +178,7 @@ export enum ErrorMessages {
   START_DATE_BEFORE_FILING_DATE = "Start date must be before or on the date of the update statement",
   CEASED_DATE_BEFORE_FILING_DATE = "Ceased date must be before or on the date of the update statement",
   RESIGNED_ON_BEFORE_FILING_DATE = "Resigned on date must be before or on the date of the update statement",
+  DATE_AFTER_MADE_UP_TO_DATE = "The date you provide must be on or before %s",
```
mwestacott, line 181: JIRA ticket needs updating to match this implementation - currently a bit unclear:

_Error message: The date you provide must be on or before ‘the entity's update statement date’ 
The entity's update statement date’ shall we replace this with the numeric date._

src/utils/date.ts
```diff
@@ -9,3 +9,11 @@ export const getTodaysDate = (): InputDate => {
     year: "" + now.year
   };
 };
+
+export const convertIsoDateToInputDate = (isoDateString: string): InputDate => {
+  const parsedDate = DateTime.fromISO(isoDateString);
+  const day = parsedDate.day.toString().padStart(2, "0"); // Ensure two-digit day
```
mwestacott, line 15: Maybe not worth it, but there's a `padWithZero()` function in `update.review.beneficial.owner.individual` which could be moved and reused here...

src/validation/error.messages.ts
```diff
@@ -35,6 +35,8 @@ export enum ErrorMessages {
   LEGAL_ENTITY_BO_NAME = "Enter its name",
   ENTITY_CORRESPONDENCE_ADDRESS = "Enter their correspondence address",
   START_DATE_MUST_BE_AFTER_DOB = "Start date must be on or after date of birth",
+  UNABLE_TO_RETRIEVE_ENTITY_NUMBER = "Unable to retrieve entity number",
```
mwestacott, line 38: Might be difficult to test/replicate but these new user error messages should get a mention on the JIRA ticket.

src/validation/custom.validation.ts
```diff
@@ -698,3 +700,26 @@ export const checkFieldIfRadioButtonSelectedAndFieldsEmpty = (isPrimaryField: bo
   }
 };
 
+export const checkDateIsBeforeOrOnNextMadeUpToDate = async (req, dayStr: string, monthStr: string, yearStr: string): Promise<boolean> => {
+  const appData: ApplicationData = getApplicationData(req.session);
+  if (!appData.entity_number) {
+    logger.errorRequest(req, "checkDateBeforeNextMadeUpToDate validation - Unable to find entity number in application data.");
+    throw new Error(ErrorMessages.UNABLE_TO_RETRIEVE_ENTITY_NUMBER);
```
mwestacott, line 707: A shame we can't get the 'service error page' to show here, as there's nothing the user can do to address this.

src/validation/custom.validation.ts
```diff
@@ -698,3 +700,26 @@ export const checkFieldIfRadioButtonSelectedAndFieldsEmpty = (isPrimaryField: bo
...
+  }
+
+  const nextMadeUpToDateIsoString = await getConfirmationStatementNextMadeUpToDateAsIsoString(req, appData.entity_number);
+  if (!nextMadeUpToDateIsoString) {
+    logger.errorRequest(req, `checkDateBeforeNextMadeUpToDate validation - Unable to find next made up to date for entity ${appData.entity_number}`);
+    throw new Error(ErrorMessages.UNABLE_TO_RETRIEVE_EXPECTED_DATE);
+  }
+
+  const nextMadeUpToDate = DateTime.fromISO(nextMadeUpToDateIsoString);
+  const userEnteredDate = DateTime.fromISO(`${yearStr}-${monthStr.padStart(2, "0")}-${dayStr.padStart(2, "0")}`);
```
mwestacott, line 717: There's that padding again. A common function would be better.

src/controllers/update/update.filing.date.controller.ts
```diff
@@ -56,3 +57,20 @@ export const post = async(req: Request, res: Response, next: NextFunction) => {
     next(error);
   }
 };
+
+const getFilingDate = async (req: Request, appData: ApplicationData): Promise<{} | undefined> => {
+  // use date stored in appData if present
```
mwestacott, line 62: I'm just curious as to when the date will be already present in `appData` and when not? Might be worth extending the comment to explain.

src/controllers/update/update.filing.date.controller.ts
```diff
@@ -56,3 +57,20 @@ export const post = async(req: Request, res: Response, next: NextFunction) => {
...
+  let filingDate = appData.update?.[FilingDateKey] ? mapDataObjectToFields(appData.update[FilingDateKey], FilingDateKeys, InputDateKeys) : undefined;
+
+  // otherwise use the next made up to date from confirmation statement in company profile
+  if (!filingDate && appData.entity_number) {
```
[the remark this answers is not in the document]
markpit: entity_number is of type 'string | undefined' but getConfirmationStatementNextMadeUpToDateAsIsoString requires a string, so the null check was to keep the compiler happy so it knows it will be a string passed into the function and not undefined. I could add a specific null check and throw a new error instead.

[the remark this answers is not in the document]
mwestacott: I think doing the specific check and throwing an error would be better. At the moment it looks like we're testing for `undefined` but not actually dealing with it when it is `undefined`.

src/utils/date.ts
mwejuli-ch: Perhaps not an issue for this PR but something to be considered for the future: a date handling package like [MomentJS](https://momentjs.com/). It's insanely awesome particularly as it's possible to compresses sizeable date handling routines into transparent one-liners (amongst other advantages).

Simplified [Timezone handling](https://momentjs.com/timezone/) is also a huge plus. 

Anyway, something to be looked at in a separate session when we can.

markpit: moment seems to be in maintenance mode now and considered a legacy project -https://www.npmjs.com/package/moment , so we use one of the recommended alternatives, luxon - https://momentjs.com/docs/#/-project-status/recommendations/
DateTime.fromISO is from the luxon package.

mwejuli-ch: Ah, got it! Thanks.

So rather than using string and padding callbacks, can we not just use Luxon's formatting options? Like this:

`DateTime.fromISO(isoDateString).toFormat('dd');`

markpit: I don't want to re-use the padWithZero, it has issues:
`export const padWithZero = (input: string, maxLength: number, fillString: string): string => {
  if (input && input.length > 1){
    return input;
  }
  return String(input).padStart(maxLength, fillString);
};`

a) it's called padWithZero, yet you have to supply the pad string which doesn't have to be a zero. All usages of this function pass the fillString as "0".
b) if input length > 1 then it doesn't pad, just returns the value. This doesn't add anything as .padStart will add leading fillString until it reaches the max length. So this could in theory fill a 1 length input string with 5 leading zeros, which doesn't make sense as input strings > 1 in length just get returned, and all usages of the function are using maxLength 2, fillString "0".
c) don't know why they are using the String(input) constructor when input is already a string.

So it doesn't seem to add anything to the built in string function padStart that is used inside it.

markpit: although it seems String(undefined).padStart(2, "0") returns undefined, 
whereas undefined.padStart(2, "0") throws an error

markpit: > Ah, got it! Thanks.
> 
> So rather than using string and padding callbacks, can we not just use Luxon's formatting options? Like this:
> 
> `DateTime.fromISO(isoDateString).toFormat('dd');`

that looks good, i'll try it out

markpit: updated this one to use the formatting

src/controllers/update/update.filing.date.controller.ts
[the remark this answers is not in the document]
markpit: added the extra check

src/validation/custom.validation.ts
markpit: I did have a quick look, i'll try and dig a bit further to see if there is a way of doing this

src/controllers/update/update.filing.date.controller.ts
markpit: updated with better comments

src/utils/date.ts
mwestacott: Nice.

src/controllers/update/update.filing.date.controller.ts
```diff
@@ -59,11 +59,15 @@ export const post = async(req: Request, res: Response, next: NextFunction) => {
...
+  // use date stored in appData if present, indicates that a value has already been entered on this page by user.
   let filingDate = appData.update?.[FilingDateKey] ? mapDataObjectToFields(appData.update[FilingDateKey], FilingDateKeys, InputDateKeys) : undefined;
 
-  // otherwise use the next made up to date from confirmation statement in company profile
-  if (!filingDate && appData.entity_number) {
+  // otherwise use the next made up to date from confirmation statement in company profile, this is the first visit to the page so they wouldn't
+  //  have already enetered a date, so pre-populate page with the next made up to date. User can then use that or change it to a different date.
+  if (!filingDate) {
+    if (!appData.entity_number) {
+      throw createAndLogErrorRequest(req, `update.filing.controller unable to find entity_number in application data for entity_name ${appData.entity_name}`);
```
mwestacott, line 69: I would just update the controller name you reference to be correct - `update.filing.date.controller`.

markpit: updated

src/validation/custom.validation.ts
markpit: i've re-jigged it so that the service error page will now show if unable to get the made up to date or entity number is missing

src/validation/custom.validation.ts
```diff
@@ -169,6 +170,26 @@ export const checkAllDateFields = (dayStr: string = "", monthStr: string = "", y
   }
 };
 
+export const checkDateForFilingDate = (dayStr: string = "", monthStr: string = "", yearStr: string = ""): boolean => {
```
mwestacott, line 173: I guess you followed the pattern of the other function you copied for the function name here, but would it be simpler (and correct) to just name it `checkFilingDate`?

markpit: updated

mwestacott (approved):

PR #1419, opened 2024-06-12
ROE-2562 Fix error messaging for trust after resuming a registration

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2562

### Change description

* 'entity_number' in session application model was 'undefined' during a registration but 'null' when resuming a registration
* Check extended so 'Still involved' error message is no longer displayed
* Unit tests added to check the changed implementation

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/validation/trust.details.validation.ts
+++ b/src/validation/trust.details.validation.ts
@@ -38,7 +38,7 @@ const setIsTrustToBeCeasedFlagOnBody = () => {
 export const checkTrustStillInvolved = (req: Request): boolean => {
   const appData: ApplicationData = getApplicationData(req.session);
 
-  const isUpdateOrRemove: boolean = appData.entity_number !== undefined;
+  const isUpdateOrRemove: boolean = !!appData.entity_number;
 
   return !hasNoBoAssignableToTrust(appData) && isUpdateOrRemove;
 };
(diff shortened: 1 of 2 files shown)


markpit (approved):

src/validation/trust.details.validation.ts
```diff
@@ -38,7 +38,7 @@ const setIsTrustToBeCeasedFlagOnBody = () => {
 export const checkTrustStillInvolved = (req: Request): boolean => {
   const appData: ApplicationData = getApplicationData(req.session);
 
-  const isUpdateOrRemove: boolean = appData.entity_number !== undefined;
+  const isUpdateOrRemove: boolean = !!appData.entity_number;
```
mattch1, line 41: Is this "!!" widely used? Could we explicitly check whether this is truthy instead?

mwestacott: It is already used in a few places within the OE web app, so not introducing something entirely new here.

If using a single ! (to check truthy value) I then need to NOT again on the next line, when checking the value, so after a quick discussion with @markpit we decided this was the best approach.

dlloyd1-ch (approved):

mattch1 (approved):

PR #1425, opened 2024-06-17
UAR-1451 Add removal of trust individuals for update and remove journeys

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1451

### Change description
Adding a new 'still involved' question to the trust individual page. If answer is 'no' then a cease date will be required to be entered.

These new question should only appear on the update and remove journeys, for both reviewing a trust or adding a new trust.

### Work checklist

### Merge instructions

```
```


markpit:
converting to draft as not to be merged until sdk and api work done and tested together

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
```diff
@@ -29,12 +29,18 @@ const getPageProperties = (trust, formData, trustee: TrustIndividual, errors?: F
       title: 'Tell us about the individual',
     },
     pageData: {
-      trustData: { trustName: trust?.trust_name },
+      trustData: {
+        trustName: trust?.trust_name,
+        creationDateDay: trust?.creation_date_day,
```
mwestacott, line 34: Wondering why 'creationDate' features in the names here and not 'ceasedDate'?

src/utils/url.ts
```diff
@@ -69,3 +69,8 @@ export function getPreviousPageUrl(req: Request, basePath: string) {
 
   return relativePreviousPageUrl;
 }
+
+export const isUpdateOrRemoveJourney = (req: Request): boolean => {
```
mwestacott, line 73: This will return `false` if user **is** on the Update or Remove journey but BEFORE a company has been loaded. There is a function `isRemoveJourney` higher up in this file that does a more thorough check...

So, I don't think this function and implementation should be in this file. Could it move to the controller/service that's calling it, at which point, presumably you know that you're far enough into the journey to be able to safely check just `entity_number`?

src/validation/trust.individual.beneficial.owner.validation.ts
```diff
@@ -7,9 +7,10 @@ import {
   usual_residential_service_address_validations
 } from "./fields/address.validation";
 import { second_nationality_validations } from "./fields/second-nationality.validation";
-import { dateBecameIPIndividualBeneficialOwner, dateOfBirthValidations } from "./fields/date.validation";
+import { dateBecameIPIndividualBeneficialOwner, dateOfBirthValidations, trustIndividualCeasedDateValidations } from "./fields/date.validation";
 import { DefaultErrorsSecondNationality } from "./models/second.nationality.error.model";
 import { ErrorMessagesOptional, ErrorMessagesRequired } from "./models/address.error.model";
+import { isUpdateOrRemoveJourney } from "../utils/url";
```
mwestacott, line 13: See above comment. Suggest moving this function into this file.

views/includes/page-templates/trust-individual-beneficial-owner.html
```diff
@@ -157,6 +166,10 @@ <h2 class="govuk-heading-m">What information we’ll show on the public register
         <input type="hidden" name="dateOfBirthYear" value="{{ formData.dateOfBirthYear }}"/>
       {% endif %}
 
+      <input type="hidden" name="trustCreationDateDay" value="{{ pageData.trustData.creationDateDay }}"/>
```
mwestacott, line 169: Again, the 'creation' fields here look wrong given this PR is all about cease dates.

src/model/trust.model.ts
```diff
@@ -79,6 +79,10 @@ export interface TrustIndividual {
   sa_address_postal_code?: string;
   sa_address_care_of?: string;
   sa_address_po_box?: string;
+  still_involved?: string,
```
mattch1, line 82: In the corresponding legal entity changes I have named these "stillInvolved" for the page model and "is_corporate_still_involved_in_trust" so as to:

1) clearly distinguish them in the web

2) reflect what is in the other micro services where the name is clearly related to the corresponding field

Which is the best course of action here?

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
markpit: one of the validation rules is to check that the cease date is after the creation date. As there might be more than one trust i figured it would be easier to get the creation date at this point and pass it into the html form where it can be used in validation of the ceased date.

src/model/trust.model.ts
markpit: i can change it to be of is_....still_involved_in_trust format

src/utils/url.ts
mwestacott: I would move it.

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
mwestacott: Is this generally the right approach - anything that we need to validate later on (when form is POSTed) has to be present in the form HTML?

I don't think we do that normally, rather, the custom validator gets hold of the app data and picks out the data it needs to validate against.

markpit: i'll look into using this approach.

src/model/trust.model.ts
mattch1: Having problem passing this value to the sdk so might use "still_involved" as well so that it is distinct at every stage.

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
lduranteau: a test comment due to problems with comments (permissions)

markpit: changed the validation to lookup the trust creation date

src/utils/url.ts
markpit: moved it into the validators

views/includes/page-templates/trust-individual-beneficial-owner.html
markpit: updated now to not use these

src/model/trust.model.ts
markpit: updated the name

src/utils/trust/individual.trustee.mapper.ts
```diff
@@ -33,6 +33,10 @@ export const mapIndividualTrusteeToSession = (
     is_service_address_same_as_usual_residential_address: (formData.is_service_address_same_as_usual_residential_address) ? Number(formData.is_service_address_same_as_usual_residential_address) : 0,
     sa_address_care_of: '',
     sa_address_po_box: '',
+    is_individual_still_involved_in_trust: formData.stillInvolved,
```
mattch1, line 36: Are we translating 0s and 1s to "Yes" and "No"?

markpit: I've followed the example used on the Trust still involved question - https://github.com/companieshouse/overseas-entities-web/blob/main/src/model/trust.model.ts#L36
The template will convert the 0's and 1's to yes and no.

markpit: actually i can see what you mean, like this - https://github.com/companieshouse/overseas-entities-web/blob/main/src/utils/trust/details.mapper.ts#L75
I'll update it.

src/model/trust.model.ts
```diff
@@ -79,6 +79,10 @@ export interface TrustIndividual {
   sa_address_postal_code?: string;
   sa_address_care_of?: string;
   sa_address_po_box?: string;
+  is_individual_still_involved_in_trust?: string,
```
mwestacott, line 82: @mattch1 has named this field as just `still_involved` in his PR. I think it would be good to be consistent and given this field (like all the others here) are in the `TrustIndividual` interface I think just `still_involved` is fine.

src/validation/error.messages.ts
```diff
@@ -180,6 +181,21 @@ export enum ErrorMessages {
   CEASED_DATE_BEFORE_FILING_DATE = "Ceased date must be before or on the date of the update statement",
   RESIGNED_ON_BEFORE_FILING_DATE = "Resigned on date must be before or on the date of the update statement",
   DATE_AFTER_MADE_UP_TO_DATE = "The date you provide must be on or before %s",
+  DAY_OF_CEASED_TRUST_INDIVIDUAL = "The date they ceased being involved in the trust must include a day",
+  DAY_LENGTH_OF_CEASED_TRUST_INDIVIDUAL = "The day of the date they ceased being involved in the trust must include 1 or 2 numbers",
+  MONTH_OF_CEASED_TRUST_INDIVIDUAL = "The date they ceased being involved in the trust must include a month",
+  MONTH_LENGTH_OF_CEASED_TRUST_INDIVIDUAL = "The month of the date they ceased being involved in the trust must include 1 or 2 numbers",
+  YEAR_OF_CEASED_TRUST_INDIVIDUAL = "The date they ceased being involved in the trust must include a year",
+  YEAR_LENGTH_OF_CEASED_TRUST_INDIVIDUAL = "The year of the date they ceased being involved in the trust must include 4 numbers",
+  ENTER_DATE_OF_TRUST_CEASED_INDIVIDUAL = "Enter the date they ceased being involved in the trust",
```
mwestacott, line 190: See this comment - https://github.com/companieshouse/overseas-entities-web/pull/1426/files#r1648895691.

I think `_CEASED_INDIVIDUAL` can be removed from the end of these error message constants and then error messages can be shared across the implementations.

src/model/trust.model.ts
markpit: ok, i'll change it back.

src/validation/error.messages.ts
markpit: will do, if any mention 'individual' then i'll leave them but can remove for others.

markpit: i'm changing them to end _CEASED_TRUSTEE
eg
DAY_LENGTH_OF_CEASED_TRUSTEE

markpit: updated

src/model/trust.model.ts
markpit: changed back to use still_involved

mattch1:
In my PR settled on "still_involved" changed everything to this.

src/utils/trust/individual.trustee.mapper.ts
```diff
@@ -9,6 +9,13 @@ export const mapIndividualTrusteeToSession = (
   formData: Page.IndividualTrusteesFormCommon,
   trustee?: Trust.TrustIndividual
 ): Trust.IndividualTrustee => {
+  let stillInvolved = (formData.stillInvolved === "1") ? "Yes" : "No";
+
+  // If a boolean value isn't receieved from the web form (it could be null or undefined, e.g. if question not displayed), need to set null
+  if (formData.stillInvolved === null || formData.stillInvolved === undefined) {
```
lduranteau, line 15: if (!formData.stillInvolved) stillInvolved = null;

src/utils/trust/individual.trustee.mapper.ts
```diff
@@ -9,6 +9,13 @@ export const mapIndividualTrusteeToSession = (
...
+    stillInvolved = null as unknown as string;
```
lduranteau, line 16: why force the type to string?

stillInvolved: string | null

src/utils/trust/individual.trustee.mapper.ts
```diff
@@ -85,6 +96,20 @@ export const mapIndividualTrusteeByIdFromSessionToPage = (
 export const mapIndividualTrusteeFromSessionToPage = (
   trustee: Trust.IndividualTrustee,
 ): Page.IndividualTrusteesFormCommon => {
+
+  let stillInvolvedInTrust: string;
+  switch (trustee.still_involved) {
```
lduranteau, line 101: it might be better to use an enum

src/validation/fields/date.validation.ts
```diff
@@ -413,3 +452,86 @@ export const dateBecameIPLegalEntityBeneficialOwner = conditionalDateValidations
...
+};
+
+/**
+ * Try and find the trust based on trust id, if not provided then assume it is the trust marked as 'in review'
+ * @param appData
+ * @param trustId
+ * @returns Trust | undefined
+ */
+const getTrust = (appData: ApplicationData, trustId?: string): Trust | undefined => {
+  let trust: Trust;
```
lduranteau, line 520: this block could be improved by using guard clauses

dlloyd1-ch:
Happy to approve once latest comments have been addressed.

src/utils/trust/individual.trustee.mapper.ts
mwestacott: Because VS code is complaining about the types:

`Type 'null' is not assignable to type 'string'`

But will give you a call to discuss.

src/utils/trust/individual.trustee.mapper.ts
mwestacott: I agree. These "Yes", "No, "0" and "1" strings are repeated throughout the code-base.

I'd suggest we leave as is for now but log this as tech-debt somewhere (a JIRA ticket, maybe). @dlloyd1-ch ?

src/validation/fields/date.validation.ts
mwestacott: We looked into the improvement but actually not quiet as straightforward as first thought. Plan to leave this as is.

src/utils/trust/individual.trustee.mapper.ts
dlloyd1-ch: The enum could just be local to this function for now. Then we'd have an example in the code that could be used elsewhere in future.

src/utils/trust/individual.trustee.mapper.ts
mwestacott: Changed this.

src/utils/trust/individual.trustee.mapper.ts
mwestacott: Changed this.

lduranteau (approved):

dlloyd1-ch (approved):
LGTM

lduranteau (approved):

PR #1461, opened 2024-07-09
ROECCT-207 refactor update.beneficial.owner.type.controller to handle…

ConnorOS-ch:
This PR moves validation handling for beneficial.owner.type.controller from the checkValidations function into the controller itself. This allows us to pass dynamic data to the page to be rendered without it interacting with the validation error handling.

This fixes a bug where the summary table at the top disappears when validation is triggered and will allow us to pass inn a relevant-period variable to the page to determine if relevant-period features should be active in a later PR

### JIRA link

https://companieshouse.atlassian.net/jira/software/c/projects/ROECCT/boards/562?selectedIssue=ROECCT-207

https://companieshouse.atlassian.net/browse/UAR-1363

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/beneficial.owner.type.controller.ts
+++ b/src/controllers/update/beneficial.owner.type.controller.ts
@@ -22,6 +22,43 @@ import { checkEntityRequiresTrusts, getTrustLandingUrl } from "../../utils/trust
 import { retrieveTrustData } from "../../utils/update/trust.model.fetch";
 import { saveAndContinue } from "../../utils/save.and.continue";
 import { Session } from "@companieshouse/node-session-handler";
+import { validationResult } from "express-validator/src/validation-result";
+import { FormattedValidationErrors, formatValidationError } from "../../middleware/validation.middleware";
+
+type BeneficialOwnerTypePageProperties = {
+  backLinkUrl: string;
+  templateName: string;
+  errors?: FormattedValidationErrors;
+  hasExistingBosMos: boolean;
+  hasNewlyAddedBosMos: boolean;
+};
+
+const getPageProperties = (
+  req: Request,
+  errors?: FormattedValidationErrors,
+): BeneficialOwnerTypePageProperties => {
+  const appData: ApplicationData = getApplicationData(req.session);
+
+  const allBosMos = [
+    ...(appData[BeneficialOwnerIndividualKey] ?? []),
+    ...(appData[BeneficialOwnerOtherKey] ?? []),
+    ...(appData[BeneficialOwnerGovKey] ?? []),
+    ...(appData[ManagingOfficerCorporateKey] ?? []),
+    ...(appData[ManagingOfficerKey] ?? [])
+  ];
+
+  const hasNewlyAddedBosMos = allBosMos.find(boMo => !boMo.ch_reference) !== undefined;
+  const hasExistingBosMos = allBosMos.find(boMo => boMo.ch_reference) !== undefined;
+
+  return {
+    backLinkUrl: config.UPDATE_BENEFICIAL_OWNER_BO_MO_REVIEW_URL,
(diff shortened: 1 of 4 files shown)


views/update/update-beneficial-owner-type.html
```diff
@@ -31,16 +31,16 @@
 {% set entity_name = entity_name %}
 {% set entity_number = entity_number %}
 
-{% if not errors %}
```
mwestacott, line 34: I'm just wondering why this if not errors check is no longer required?

src/controllers/update/beneficial.owner.type.controller.ts
```diff
@@ -22,6 +22,43 @@ import { checkEntityRequiresTrusts, getTrustLandingUrl } from "../../utils/trust
...
+  const hasExistingBosMos = allBosMos.find(boMo => boMo.ch_reference) !== undefined;
+
+  return {
+    backLinkUrl: config.UPDATE_BENEFICIAL_OWNER_BO_MO_REVIEW_URL,
+    templateName: config.UPDATE_BENEFICIAL_OWNER_TYPE_PAGE,
+    hasNewlyAddedBosMos,
+    hasExistingBosMos,
+    ...appData,
+    errors
+  };
```
markpit, line 60: I would do some manual checks to make sure the Remove journey still works ok. there is a section in the checkValidations that re-set the 'journey' page param - https://github.com/companieshouse/overseas-entities-web/blob/main/src/middleware/validation.middleware.ts#L62

it might be ok if we are passing appData in, but i would test it in docker to make sure it is ok, eg does it loose the 'Remove' page title if you get a validation error on that page on a Remove journey, and also clicking 'sign out' after a validation error on Remove journey on that page, then clicking the back link etc..

views/update/update-beneficial-owner-type.html
ConnorOS-ch: I can’t determine why this `if not errors` is here in the first place, including it makes it so that the list of BO’s you’ve added already doesn’t show on the page after a validation error is triggered, but I can figure out why we would want that to disappear or how this could break the code

It's hard to understand from the commit where it was added

mwestacott: Right, so you're only removing it in order to clean up the code? If you leave it, do tests all still pass and the app behave as expected?

ConnorOS-ch: Hi, yes essentially to clean up the code. But I'd like to understand the purpose better before we move ahead. So the behaviour of the app on the current main branch if you trigger a validation error on this page is in this screenshot:

image

I assumed that the summary table at the top should not disappear, so deleted `{% if not errors %}` in the template, but it's hard for me to determine from the original PR's why this was included in the first place. If i'm right Toro Loco where not directly involved with this page, do we have anyone we can ask about this?

ConnorOS-ch: To answer your question, unit tests still pass and I'd like to add an additional test to make sure this content "What you have reviewed" and "What you have added in this update" is still displayed even if errors are present on the page. Checking with our BA also about what the expected behaviour ought to be

src/controllers/update/beneficial.owner.type.controller.ts
ConnorOS-ch: Yes, I'm still seeing "Apply to remove an overseas entity from the register" as the page title after triggering an error with these changes

views/update/update-beneficial-owner-type.html
mwestacott: Yes, Update journey implemented by a Kainos team originally, so no history on this unfortunately. As discussed, as long as it's well tested, this change is hopefully fine and correct.

markpit:
could you add a link to the JIRA ticket in the description box

mwestacott (approved):

PR #1463, opened 2024-07-10
Uar 1448 check page trust statuses

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1448

### Change description

As on the trust/trustee review pages - added status fields to templates.

### Work checklist

### Merge instructions

```
```


test/controllers/update/check.your.answers.controller.spec.ts
```diff
@@ -466,7 +489,490 @@ describe("CHECK YOUR ANSWERS controller", () => {
...
+      expect(resp.status).toEqual(200);
+      expect(resp.text).toContain(TRUST_NAME);
+      expect(resp.text).toContain("wizzz trust");
+      expect(resp.text).toContain(TRUST_CREATION_DATE);
+      expect(resp.text).toContain("31");
+      expect(resp.text).toContain("December");
+      expect(resp.text).toContain("1999");
+      expect(resp.text).toContain(TRUSTS_ADDED);
+      expect(resp.text).not.toContain(TRUSTS_REVIEWED);
+      expect(resp.text).toContain(TRUST_INVOLVED);
```
markpit, line 555: would it be possible to check the value of TRUST_INVOLVED , not that constant value but it should show "No" on the page. There are probably lots of No's in the text but there might be a string in the resp.text containing the No and something to identify it as the trust involved value.

markpit: similarly for when that value is Yes, could we check that the Yes appears as the answer for that item.

mattch1: I did try to do this but it would difficult as there would be a lot of characters that don't fit well into expect string. This is covered implicitly by checking for the presence or absence of the ceased date.

markpit: yes, I had a look and you'd have to include html in the text which is not ideal. So leave it as it is.

markpit (approved):

PR #1464, opened 2024-07-10
ROECCT-204- bug - Using back buttons and selecting 'no' on filter page still shows pre-reg

krenjith-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROECCT-204

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/relevant.period.owned.land.filter.controller.ts
+++ b/src/controllers/update/relevant.period.owned.land.filter.controller.ts
@@ -5,6 +5,7 @@ import * as config from "../../config";
 import { ApplicationData } from "../../model";
 import { getApplicationData } from "../../utils/application.data";
 import { OwnedLandKey } from "../../model/update.type.model";
+import { ChangeBeneficiaryRelevantPeriodKey, ChangeBeneficiaryRelevantPeriodType, ChangeBoRelevantPeriodKey, ChangeBoRelevantPeriodType, TrusteeInvolvedRelevantPeriodKey, TrusteeInvolvedRelevantPeriodType } from "../../model/relevant.period.statment.model";
 
 export const get = (req: Request, res: Response, next: NextFunction) => {
   try {
@@ -26,10 +27,16 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
   try {
     logger.debugRequest(req, `POST ${config.RELEVANT_PERIOD_OWNED_LAND_FILTER_PAGE}`);
     const ownedLandKey = req.body[OwnedLandKey];
+    const appData: ApplicationData = getApplicationData(req.session);
 
     if (ownedLandKey === '1') {
       return res.redirect(config.RELEVANT_PERIOD_INTERRUPT_URL);
     } else {
+      if (appData.update) {
+        appData.update[ChangeBoRelevantPeriodKey] = ChangeBoRelevantPeriodType.NO;
+        appData.update[TrusteeInvolvedRelevantPeriodKey] = TrusteeInvolvedRelevantPeriodType.NO;
+        appData.update[ChangeBeneficiaryRelevantPeriodKey] = ChangeBeneficiaryRelevantPeriodType.NO;
+      }
       return res.redirect(config.UPDATE_FILING_DATE_URL);
     }
   } catch (error) {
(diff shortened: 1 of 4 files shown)


views/includes/page-templates/trust-involved.html
```diff
@@ -212,7 +212,7 @@ <h1 class="govuk-heading-m">
         } %}
         {% include "includes/inputs/radio-input.html" %}
 
-        {% if pageData.isRelevantPeriod %}
+        {% if pageData.isRelevantPeriod or (not update.change_bo_relevant_period === "undefined") and (not update.trustee_involved_relevant_period === "undefined") and (not update.change_beneficiary_relevant_period === "undefined") %}
```
ConnorOS-ch, line 215: pageData.isRelevantPeriod is already set in the controller based on the three other conditions, so both sides of your `or` are the same, I don't think we need to update any of the logic on these pages :)

I realise it is confusing the way I've done this page, I want to move to us always setting RelevantPeriod in the controller instead of checking `update.___` in the template, but at the moment we have both.

src/controllers/update/relevant.period.owned.land.filter.controller.ts
```diff
@@ -26,10 +27,16 @@ export const post = (req: Request, res: Response, next: NextFunction) => {
...
 
     if (ownedLandKey === '1') {
       return res.redirect(config.RELEVANT_PERIOD_INTERRUPT_URL);
     } else {
+      if (appData.update) {
+        appData.update[ChangeBoRelevantPeriodKey] = ChangeBoRelevantPeriodType.NO;
+        appData.update[TrusteeInvolvedRelevantPeriodKey] = TrusteeInvolvedRelevantPeriodType.NO;
+        appData.update[ChangeBeneficiaryRelevantPeriodKey] = ChangeBeneficiaryRelevantPeriodType.NO;
+      }
       return res.redirect(config.UPDATE_FILING_DATE_URL);
```
ConnorOS-ch, line 40: A we spoke about, you can set these to `undefined` and check it's pulling through to mongo.

bear in mind that the `await` call needs to be inside an `async` function so you will need to make this post controller `async`

views/update/update-beneficial-owner-type.html
```diff
@@ -93,11 +93,14 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ title }}</h1>
   <div class="govuk-grid-column-two-thirds">
     <form method="post">
       {% include "includes/csrf_token.html" %}
-      {% include "includes/beneficial-owner-type-add-radios.html" %}
```
ConnorOS-ch, line 96: I think this has undone some changes from main, shouldn't need any changes to this file

test/controllers/update/relevant.period.owned.land.filter.spec.ts
```diff
@@ -139,12 +145,20 @@ describe("owned land filter page tests", () => {
...
     });
     test(`renders the ${config.UPDATE_DO_YOU_WANT_TO_MAKE_OE_CHANGE_URL} page when no is selected`, async () => {
+      mockGetApplicationData.mockReturnValue({ ...APPLICATION_DATA_MOCK });
+      mockIsActiveFeature.mockReturnValueOnce(true);
       const resp = await request(app)
         .post(config.RELEVANT_PERIOD_OWNED_LAND_FILTER_URL)
         .send({ owned_land_relevant_period: "0" });
+      const mockAppData = {};
 
+      mockAppData[ChangeBoRelevantPeriodKey] = undefined;
```
ConnorOS-ch, line 155: What is this testing? do we need `mockAppData` here?

krenjith-ch: modified the code, thanks Connor

views/update/update-beneficial-owner-type.html
```diff
@@ -94,8 +94,7 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ title }}</h1>
     <form method="post">
       {% include "includes/csrf_token.html" %}
       {% include "includes/beneficial-owner-type-add-radios.html" %}
-
-      {% if (feature_flag_relevant_period === "true") and (update.change_bo_relevant_period === "CHANGE_BO_RELEVANT_PERIOD" or update.trustee_involved_relevant_period === "TRUSTEE_INVOLVED_RELEVANT_PERIOD" or update.change_beneficiary_relevant_period === "CHANGE_BENEFICIARY_RELEVANT_PERIOD") %}
+      {% if (feature_flag_relevant_period === "true") and (update.change_bo_relevant_period === "CHANGE_BO_RELEVANT_PERIOD") %}
```
ConnorOS-ch, line 97: If we include this change, you can only select the first checkbox for this case. We want any of the checkboxes to trigger relevant period content

ConnorOS-ch (approved):
Nice work

PR #1471, opened 2024-07-16
Feat/ROE-2559: Save `sold-land-filter` option to database

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2559

### Change description

- Saves `sold-land-filter` option to the API

### Work checklist


src/controllers/sold.land.filter.controller.ts
```diff
@@ -4,8 +4,13 @@ import { logger } from "../utils/logger";
 import * as config from "../config";
 import { ApplicationData } from "../model";
 import { deleteApplicationData, getApplicationData, setExtraData } from "../utils/application.data";
-import { HasSoldLandKey } from "../model/data.types.model";
+import { HasSoldLandKey, OverseasEntityKey, Transactionkey } from "../model/data.types.model";
 import { getSoldLandFilterBackLink } from "../utils/navigation";
+import { isActiveFeature } from "../utils/feature.flag";
+import { postTransaction } from "../service/transaction.service";
+import { createOverseasEntity, updateOverseasEntity } from "../service/overseas.entities.service";
+// import { getUrlWithTransactionIdAndSubmissionId } from "../utils/url";
```
lduranteau, line 12: I think you forgot to remove the import that you no longer use.

src/controllers/sold.land.filter.controller.ts
```diff
@@ -28,18 +33,34 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
...
+    const appData: ApplicationData = getApplicationData(session);
+    appData[HasSoldLandKey] = hasSoldLand;
+
+    let nextPageUrl: string = "";
 
-    if (hasSoldLand === '1') {
-      return res.redirect(config.CANNOT_USE_URL);
-    } else if (hasSoldLand === '0') {
-      return res.redirect(config.SECURE_REGISTER_FILTER_URL);
+    if (hasSoldLand === "1") {
```
lduranteau, line 48: It might be better to use a guard clause approach to avoid having nested if else.
This could improve readability.

src/service/overseas.entities.service.ts
```diff
@@ -11,18 +11,21 @@ export const createOverseasEntity = async (
...
   session: Session,
   transactionId: string,
-  isSaveAndResumeFeatureActive: boolean = false
+  isSaveAndResumeFeatureActive: boolean = false,
+  data?: ApplicationData
 ): Promise<string> => {
 
   logger.infoRequest(req, `Calling 'postOverseasEntity' for transaction id '${transactionId}'`);
 
+  const appData: ApplicationData = typeof data !== "undefined" ? data : getApplicationData(session);
```
lduranteau, line 20: If you don't need to check specifically for undefined you can maybe only check if data exists.
const appData: ApplicationData = !data ? getApplicationData(session) : data

This is just a suggestion in case you end up with null for some reason.

src/controllers/sold.land.filter.controller.ts
mwejuli-ch: Yeah, good spot. Will get rid of it. Thanks.

src/service/overseas.entities.service.ts
lduranteau: I agree, the nullish coalescing operator is more suitable in this case.

lduranteau (approved):

test/controllers/overseas.name.controller.spec.ts
```diff
@@ -195,24 +211,42 @@ describe("Overseas Name controller", () => {
...
-      mockGetApplicationData.mockReturnValueOnce({}); // Needed at the setExtraData
-      mockGetApplicationData.mockReturnValueOnce(mockData); // Needed inside the feature flag
+      const mockData = { ...APPLICATION_DATA_MOCK, [Transactionkey]: null, [OverseasEntityKey]: null };
+      mockGetApplicationData.mockReturnValueOnce(mockData);
       const resp = await request(app).post(OVERSEAS_NAME_URL).send({ [EntityNameKey]: OVERSEAS_NAME_MOCK });
 
       expect(resp.status).toEqual(302);
-      expect(mockData[Transactionkey]).toEqual(TRANSACTION_ID);
-      expect(mockData[OverseasEntityKey]).toEqual(OVERSEAS_ENTITY_ID);
+      expect(mockData[Transactionkey]).toEqual(null);
```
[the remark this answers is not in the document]
markpit: I can see what's happening, the setExtraData was calling getApplicationData but we had a second call to getApplicationData that meant we could test the mockAppData after it was modified. The new version removes the 2nd call to getApplicationData and now copies the fields , so we can't test the mockAppData anymore. We should be testing that appData contains the transactionKey and overseasEntityKey when it is passed into setExtraData

test/controllers/overseas.name.controller.spec.ts
```diff
@@ -195,24 +211,42 @@ describe("Overseas Name controller", () => {
...
+      mockIsActiveFeature.mockReturnValueOnce(true); // For FEATURE_FLAG_ENABLE_SAVE_AND_RESUME
+      mockIsActiveFeature.mockReturnValueOnce(true); // For FEATURE_FLAG_ENABLE_REDIS_REMOVAL
+
+      const mockData = { ...APPLICATION_DATA_MOCK, [Transactionkey]: null, [OverseasEntityKey]: null };
+      mockGetApplicationData.mockReturnValueOnce(mockData);
+      const resp = await request(app).post(OVERSEAS_NAME_URL).send({ [EntityNameKey]: OVERSEAS_NAME_MOCK });
+
+      expect(resp.status).toEqual(302);
+      expect(mockData[Transactionkey]).toEqual(null);
+      expect(mockData[OverseasEntityKey]).toEqual(null);
```
markpit, line 242: same here for the TRANSACTION_ID and OVERSEAS_ENTITY_ID check

test/controllers/sold.land.filter.controller.spec.ts
```diff
@@ -116,6 +139,20 @@ describe("SOLD LAND FILTER controller", () => {
...
+      mockGetApplicationData.mockReturnValueOnce({});
+      mockGetApplicationData.mockReturnValueOnce({});
+      mockIsActiveFeature.mockReturnValueOnce(true);
+      mockIsActiveFeature.mockReturnValueOnce(true);
+      const resp = await request(app).post(config.SOLD_LAND_FILTER_URL).send({ has_sold_land: '0' });
+
+      expect(resp.status).toEqual(302);
+      expect(mockUpdateOverseasEntity).toHaveBeenCalledTimes(0);
+      expect(mockTransactionService).toHaveBeenCalledTimes(1);
+      expect(mockCreateOverseasEntity).toHaveBeenCalledTimes(1);
```
markpit, line 153: need to check the redirect destination here as well

test/controllers/overseas.name.controller.spec.ts
mwejuli-ch: Done.

markpit (approved):

PR #1473, opened 2024-07-17
ROE-2571 remove enable save and resume feature flag for Registration

markpit:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-2571

### Change description
Removing the feature flag for save and resume on registration journey - FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022

* I've had to use hardcoded true/false values in some places where the function is shared between update and registration. Once Update S&R flag is removed then the true/false values can be removed. 
* package.json will need to be updated to pick up the modified api-sdk-node that removes the /start endpoint. I've tested with this new api-sdk-node in Docker.

### Work checklist

### Merge instructions

```
```


src/controllers/sign.out.controller.ts
```diff
@@ -14,7 +13,7 @@ export const get = (req: Request, res: Response, next: NextFunction) => {
 
     return res.render(config.SIGN_OUT_PAGE, {
       previousPage: previousPageUrl,
-      saveAndResume: isActiveFeature(config.FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022),
+      saveAndResume: true,
```
markpit, line 16: the SIGN_OUT_PAGE is also used by the update sign out controller so the template will still need the logic until the update S&R flag is removed, then this saveAndResume: true can also be removed.

config.UPDATE_SIGN_OUT_PAGE = "sign-out", the same as config.SIGN_OUT_PAGE, no idea why another config value was needed, but they both point to the same template.

src/controllers/update/remove.cannot.use.controller.ts
```diff
@@ -4,7 +4,7 @@ import { logger } from "../../utils/logger";
 
 export const get = (req: Request, res: Response, next: NextFunction) => {
   try {
-    logger.debugRequest(req, `GET ${config.SIGN_OUT_PAGE}`);
+    logger.debugRequest(req, `GET ${config.REMOVE_CANNOT_USE_PAGE}`);
```
markpit, line 7: this had wrong template name in logger

src/service/overseas.entities.service.ts
```diff
@@ -7,11 +7,13 @@ import { Transactionkey, OverseasEntityKey } from "../model/data.types.model";
...
 import { ApplicationData } from "../model/application.model";
 
+// TODO find all calls to this function and remove the isSaveAndResumeFeatureActive
+// parameter when registration and update save and resume flags are removed
 export const createOverseasEntity = async (
   req: Request,
   session: Session,
   transactionId: string,
-  isSaveAndResumeFeatureActive: boolean = false
+  isSaveAndResumeFeatureActive: boolean
```
markpit, line 16: removed the default of false here, and changed the calls that passed no value to pass false. This will be removed when update S&R flag removed.

test/controllers/check.your.answers.controller.spec.ts
```diff
@@ -1177,29 +1167,8 @@ describe("POST tests", () => {
...
-
-    expect(resp.status).toEqual(302);
-    expect(mockTransactionService).toHaveBeenCalledTimes(1);
-    expect(mockOverseasEntity).toHaveBeenCalledTimes(1);
-    expect(resp.text).toEqual(`${FOUND_REDIRECT_TO} ${PAYMENT_LINK_JOURNEY}`);
-  });
-
   test(`catch error when post data from ${CHECK_YOUR_ANSWERS_PAGE} page`, async () => {
-    mockOverseasEntity.mockImplementation(() => {
+    mockGetApplicationData.mockImplementation(() => {
```
markpit, line 1171: the mockOverseasEntity has now been removed from test so needed another mock to throw the error

test/utils/navigation.spec.ts
```diff
@@ -37,26 +37,23 @@ const mockRequestWithParams = {
 describe("NAVIGATION utils", () => {
 
   test(`getEntityBackLink returns ${config.DUE_DILIGENCE_URL} when ${WhoIsRegisteringType.AGENT} selected`, () => {
-    const entityBackLink = getEntityBackLink({ [WhoIsRegisteringKey]: WhoIsRegisteringType.AGENT });
+    const mockRequest = { query: {} } as Request;
```
markpit, line 40: there were warnings on these tests about missing request param so added them in.

mattch1 (approved):
Looks OK to me; see what other think.

views/includes/save-and-continue-button.html
```diff
@@ -5,9 +5,7 @@
     {% set textContinueButton = "Save and continue" %}
   {% endif %}
 {% else %}
-  {% if OE_CONFIGS.FEATURE_FLAG_ENABLE_SAVE_AND_RESUME_17102022 === "true" %}
-    {% set textContinueButton = "Save and continue" %}
-  {% endif %}
+  {% set textContinueButton = "Save and continue" %}
```
mwestacott, line 8: I think this logic could be simplified now (to set "Save and continue" to start with and have one conditional check that sets "Continue" if appropriate but not really a problem as this will all disappear when the next feature flag is removed.

mwestacott (approved):

src/utils/check.your.answers.ts
```diff
@@ -120,7 +120,7 @@ export const postDataForReview = async (req: Request, res: Response, next: NextF
       overseasEntityID = appData[OverseasEntityKey] as string;
     } else {
       transactionID = await postTransaction(req, session);
-      overseasEntityID = await createOverseasEntity(req, session, transactionID);
+      overseasEntityID = await createOverseasEntity(req, session, transactionID, false);
```
dlloyd1-ch, line 123: Just checking why we're setting `isSaveAndResumeFeatureActive` to `false`. Wont this feature be permanently active once the flag is removed?

dlloyd1-ch (approved):

PR #1475, opened 2024-07-17
ROECCT-221: Amend Continue Button Goals for Relevant Period Interrupt Card

jnicholson-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROECCT-221

### Change description

This ticket involves updating the goal tracking numbers for the "Continue" button on the /update-an-overseas-entity/relevant-period-interrupt-card page. Currently, this button raises the same goal as the "Continue" button on the /update-an-overseas-entity/update-interrupt-card page. The continue button on the new page needs to be amended to use distinct goal numbers for different environments: 33 for Docker, 55 for CIdev, and 52 for Staging.

This is one part of the ticket.

### Work checklist

### Merge instructions

```
```


src/config/index.ts
```diff
@@ -53,7 +54,7 @@ export const FEATURE_FLAG_ENABLE_UPDATE_STATEMENT_VALIDATION = getEnvironmentVal
 export const FEATURE_FLAG_ENABLE_UPDATE_MANAGE_TRUSTS = getEnvironmentValue('FEATURE_FLAG_ENABLE_UPDATE_MANAGE_TRUSTS_29082023', 'false');
 export const FEATURE_FLAG_ENABLE_REDIS_REMOVAL = getEnvironmentValue('FEATURE_FLAG_ENABLE_REDIS_REMOVAL_27092023', 'false');
 export const FEATURE_FLAG_ENABLE_CEASE_TRUSTS = getEnvironmentValue('FEATURE_FLAG_ENABLE_CEASE_TRUSTS_19022024', 'false');
-export const FEATURE_FLAG_ENABLE_RELEVANT_PERIOD = getEnvironmentValue('FEATURE_FLAG_ENABLE_RELEVANT_PERIOD_26042024', 'false');
+export const FEATURE_FLAG_ENABLE_RELEVANT_PERIOD = getEnvironmentValue('FEATURE_FLAG_ENABLE_RELEVANT_PERIOD_26042024', 'true');
```
ConnorOS-ch, line 57: it's okay to set these to true on your Local, although you shouldn't need to. It's not a good idea to commit these as true on the repo. These values are set through environment variables so that we can have different features turned on and off in different environments.

ConnorOS-ch: The values you've changed are the default values. i.e the value when no environment variable is found with that name. The service defaults to false for all these as a precaution I believe. If you need to change these on your local, a better way to do it is to set the values in `overseas-entities-web.docker-compose.yaml` in your docker-chs-development repo

ConnorOS-ch (changes requested):

src/config/index.ts
jnicholson-ch: > it's okay to set these to true on your Local, although you shouldn't need to. It's not a good idea to commit these as true on the repo. These values are set through environment variables so that we can have different features turned on and off in different environments.

I'll update the feature flags now.

ConnorOS-ch (approved):
LGTM

kullas-ch (approved):

PR #1477, opened 2024-07-18
Fix/uar 1371 incorporation country

lduranteau:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1371

### Change description
prevents incorporation country from being updated when updating overseas entity address

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/update/overseas.entity.update.details.controller.ts
+++ b/src/controllers/update/overseas.entity.update.details.controller.ts
@@ -6,7 +6,7 @@ import {
   mapDataObjectToFields,
   setExtraData
 } from "../../utils/application.data";
-import { EntityKey } from "../../model/entity.model";
+import { Entity, EntityKey } from "../../model/entity.model";
 import { ApplicationData, ApplicationDataType } from "../../model";
 import {
   AddressKeys,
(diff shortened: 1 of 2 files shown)


markpit (approved):

src/controllers/update/overseas.entity.update.details.controller.ts
```diff
@@ -68,6 +68,8 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
     const session = req.session as Session;
     const entityName = req.body[EntityNameKey];
 
+    setOriginalIncorporationCountry(session, data);
```
mwestacott, line 71: Worth a comment here to say why this is done. It's not done for any of the other fields. I think it's because it's a read-only field but there's a risk it's overwritten to due to auto-fill function in some browsers?

lduranteau: Do you have a suggestion for the comment?

mwestacott: // The original incorporation country is set again here in case it's been overwritten with an incorrect value (it's the only read-only field on the screen and may be replaced with another value by browser auto-fill functionality)

Assuming what I've written above is correct!

mwestacott (approved):

mwejuli-ch (approved):

PR #1511, opened 2024-07-31
UAR-1177 Delete manage trusts flag

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1177

### Change description

**Delete the ROE web managed trusts flag from several places -**
- src/controllers/update/beneficial.owner.type.controller.ts
- src/controllers/update/overseas.entity.change.controller.ts
- Numerous routing configs

**Updated a number of tests**
- Removed "flag off" tests for this flag.
- beneficial.owner.type.controller test - reworked after removing flag so reworked tests to cover all the new scenarios resulting from the removal and rework.
-  test/controllers/update/update.manage.trusts.review.the.trust.controller.spec - two lines were failing but now fixed:
1) expect(resp.text).not.toContain(ErrorMessages.TRUST_INVOLVED_BOS); Removed extra unneeded feature flag mock.
2) expect(resp.text).toContain(ErrorMessages.ENTER_DATE_OF_TRUST_CEASED); Relates tot he template

**Templates**
Updated the views/update/review-update-statement.html template

### Work checklist

### Merge instructions

```
```


src/controllers/update/beneficial.owner.type.controller.ts
```diff
@@ -103,35 +103,31 @@ export const postSubmit = async (req: Request, res: Response, next: NextFunction
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
     const appData: ApplicationData = getApplicationData(req.session);
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_UPDATE_TRUSTS) && checkEntityRequiresTrusts(appData)) {
```
markpit, line 106: I don't think you should move this section to the top as it affects the redirect-ing.
Previously, if hasTrustsToReview would redirect back to the start of the trusts pages before the checkEntityRequiresTrusts redirect.

Now it will perform the checkEntityRequiresTrusts redirect first, which will alter the behaviour of the page

markpit: i would just reset this method and just remove the if check for the feature flag we are removing

views/update/review-update-statement.html
```diff
@@ -315,23 +314,21 @@ <h2 class="govuk-heading-l">Managing officers</h2>
...
       <br>
       <h2 class="govuk-heading-l">Trusts</h2>
       {% for trust in update.review_trusts %}
         {% set trust_heading = trust.trust_name %}
         {% include "includes/check-your-answers/trusts.html" %}
       {% endfor %}
+    {% elif pageParams.hasAnyBosWithTrusteeNocs %}
+      <br>
+      <h2 class="govuk-heading-l">Trusts</h2>
+      <p class="govuk-body govuk-!-margin-bottom-9">Trust information cannot be displayed.</p>
```
markpit, line 328: I think this Trust information cannot be displayed option can be removed as it shouldn't be reachable now we have effectively set manageTrustsEnabled = true.
In previous version it was only reachable when:
    addTrustsEnabled = true
    manageTrustsEnabled = false
    pageParams.hasAnyBosWithTrusteeNocs = true

In this new version it has changed and is now reachable when :
    addTrustsEnabled = true
    hasExistingTrusts = false
    pageParams.hasAnyBosWithTrusteeNocs = true

src/controllers/update/overseas.entity.change.controller.ts
```diff
@@ -91,14 +90,12 @@ export const resetDataForNoChange = async (req: Request, appData: ApplicationDat
     appData.update.registrable_beneficial_owner = undefined;
     appData.update.bo_mo_data_fetched = false;
     await retrieveBoAndMoData(req, appData);
+
+    appData.update.trust_data_fetched = false;
+    appData.update.review_trusts = undefined;
+    await retrieveTrustData(req, appData);
```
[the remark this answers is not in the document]
mattch1: We do.

src/controllers/update/beneficial.owner.type.controller.ts
mattch1: Reset this.

views/update/review-update-statement.html
mattch1: Not sure this makes sense as the elif is not directly related to the managed trusts if but the nested if. Not sure this made sense in the original unchanged version tbh, so removed it.

src/controllers/update/overseas.entity.change.controller.ts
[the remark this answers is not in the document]
dlloyd1-ch: Ah ok, didn't see line 89 as it wasn't showing in the PR.

views/update/review-update-statement.html
markpit: Can have a call to discuss it, but the way to look at it is that we are now making manageTrustsEnabled permanently true, and when you look at the previous if /elif logic, when manageTrustsEnabled is true, that 'missing trusts' text can never be reached.

mattch1: This is clear now, the elif relates to managed trusts and was reachable only when the value was false, which will no longer be the case.

markpit (approved):

PR #1522, opened 2024-08-07
Feat/ROE-2375: Resume journey with `transactionId` and `submissionId` in URL

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2375

### Change description

Adds `transactionId` and `submissionId` in URL for the resume journey

### Work checklist

### DIFF
--- a/test/controllers/resume.submission.controller.spec.ts
+++ b/test/controllers/resume.submission.controller.spec.ts
@@ -8,6 +8,7 @@ jest.mock('../../src/utils/application.data');
 jest.mock("../../src/utils/feature.flag" );
 jest.mock("../../src/utils/logger");
 jest.mock("../../src/utils/trusts");
+jest.mock("../../src/utils/url");
 
 import { describe, expect, test, jest, beforeEach } from '@jest/globals';
 import { NextFunction, Request, Response } from "express";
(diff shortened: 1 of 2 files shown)


test/controllers/resume.submission.controller.spec.ts
```diff
@@ -185,6 +186,38 @@ describe("Resume submission controller", () => {
...
+      OVERSEAS_ENTITY_ID,
+      { headers: PAYMENT_HEADER },
+      baseURL
+    );
+    expect(mockInfoRequest).toHaveBeenCalledWith( expect.anything(), `Payments Session created on Resume link with, ${errorMsg}`);
+    expect(mockGetOverseasEntity).toHaveBeenCalledTimes(1);
+    expect(mockSetExtraData).toHaveBeenCalledTimes(1);
+    expect(mockCreateAndLogErrorRequest).not.toHaveBeenCalled();
+    expect(mockMapTrustApiReturnModelToWebModel).toHaveBeenCalledTimes(1);
+    expect(mockIsActiveFeature).toHaveBeenCalledTimes(2);
```
dlloyd1-ch, line 220: Can we add an assertion that checks that `getUrlWithTransactionIdAndSubmissionId()` has been called when the flag is set to `true`?

mwejuli-ch: done!

markpit:
LGTM, will await dan's comment response

dlloyd1-ch (approved):

mwestacott (commented):
I can't see a code change for the Pay Now link behaviour described in the story. Is one needed or not?

mwejuli-ch:
> I can't see a code change for the Pay Now link behaviour described in the story. Is one needed or not?

I do not think we need one coz we're only doing resume (from the `Your filings` tab). But I stand to be corrected..

mwestacott:
> > I can't see a code change for the Pay Now link behaviour described in the story. Is one needed or not?
> 
> I do not think we need one coz we're only doing resume (from the `Your filings` tab). But I stand to be corrected..

If you've tested the scenario and it's working you're probably fine. The transaction id appears on the confirmation page as the reference number and is read from the session but it's probably not needed within the URL. When we change `getApplicationData` to read from the API not session it will hopefully all magically just work.

mwestacott (approved):

lduranteau (approved):

PR #1564, opened 2024-09-02
Refactor/asm 350/asynchronous application data retrieval move async validations

lduranteau:
### Work checklist

### Merge instructions

```
```


src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
```diff
@@ -101,3 +106,11 @@ const getBackLink = (formerBosReviewed) => {
     return UPDATE_MANAGE_TRUSTS_REVIEW_FORMER_BO_URL;
   }
 };
+
+const checkErrors = async (req: Request): Promise<ValidationError[]> => {
```
mwestacott, line 110: It feels like this function if checking _for_ errors. It's really just validating the trust start date and cease date (I think) so how about calling it `validateFilingPeriodDates()`? The returned result is then 'checked'.

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
```diff
@@ -126,3 +132,10 @@ const getPagePropertiesRelevantPeriod = (relevant_period, trust, formData, trust
   pageProps.pageData.entity_name = entityName;
   return pageProps;
 };
+
+const checkErrors = async (appData: ApplicationData, req: Request): Promise<ValidationError[]> => {
```
mwestacott, line 136: Similar comment as above.

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
```diff
@@ -50,8 +51,12 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 
     // check for form validation errors
     const errorList = validationResult(req);
-    if (!errorList.isEmpty()) {
-      const pageProperties = getPageProperties(trust, formData, formatValidationError(errorList.array()));
+    const errors = await checkErrors(req);
```
mwestacott, line 54: This variable could be renamed to `filingPeriodDateErrors`.

src/validation/async/index.ts
```diff
@@ -0,0 +1,260 @@
...
+import * as config from '../../config';
+import { ApplicationData } from "model";
+import { RoleWithinTrustType } from '../../model/role.within.trust.type.model';
+import { ErrorMessages } from '../../validation/error.messages';
+import { checkCeasedDateOnOrAfterDateOfBirth, checkCeasedDateOnOrAfterTrustCreationDate, checkIndividualCeasedDateOnOrAfterInterestedPersonStartDate, checkLegalEntityCeasedDateOnOrAfterInterestedPersonStartDate, historicalBOEndDateContext, historicalBOStartDateContext } from '../../validation/fields/date.validation';
+import { hasNoBoAssignableToTrust } from '../../utils/trusts';
+import { dateContext } from '../../validation/fields/helper/date.validation.helper';
+import { checkDatePreviousToFilingDate } from '../../validation/custom.validation';
+
+export const checkTrustStillInvolved = (appData: ApplicationData, req): ValidationError[] => {
```
mwestacott, line 13: A comment (somewhere) explaining why these functions are sitting in a folder named `async` would be helpful. I know this is probably related to the async call to get application data but saying that `appData` is being passed in to these functions so why are they asynchronous?

src/validation/fields/date.validation.ts
```diff
@@ -92,45 +91,12 @@ const is_trust_still_active_validation = (error_message: string) => [
     )),
 ];
 
-const checkAgainstFilingDate = (date_field_id, error_message) =>
```
mwestacott, line 95: Could these functions not remain where they are and only the `checkDatePreviousToFilingDate` moves, as that's the one that makes the async call?

test/validation/custom.validation.spec.ts
```diff
@@ -231,7 +231,8 @@ describe('tests for checkDatePreviousToFilingDate ', () => {
       .toBeTruthy();
   });
 
-  test("should return error if startDate is after filingDate", () => {
+  // asm-350 - Jest worker encountered 4 child process exceptions, exceeding retry limit
+  test.skip("should return error if startDate is after filingDate", () => {
```
mwestacott, line 235: Should this be skipped as part of this 'async refactoring' change?

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
lduranteau: This name is the same in all controllers because the job is similar: to get validation errors that are not in errorList.
The name is as common as can be errorList = validationResult(req)

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
lduranteau: This name is the same in all controllers because the job is similar: to get validation errors that are not in errorList.
The name is as common as can be errorList = validationResult(req)

src/controllers/update/update.manage.trusts.tell.us.about.the.individual.controller.ts
lduranteau: Similar response as above 😁

src/validation/fields/date.validation.ts
lduranteau: No, because all these functions have been transformed because they use an async function and so must be async.

test/validation/custom.validation.spec.ts
lduranteau: No this is why I left a comment with the ticket number
The tests run but there is a problem because an async function is still running when the tests are stopped
There will be a deeper investigation

src/validation/async/index.ts
lduranteau: They need to make a call to getApplication, which is why they have been moved
But as we already make the getApplication call at controller level, I pass appData to avoid making two (or more) calls

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
mwestacott: I'd rather a specific name - it's not having to implement a common interface function or anything - but regardless, renaming to `getValidationErrors` would be better I think because that matches what you said in your comment. The actual checking of the errors happens on line 56 (I think) when you check the length of the array.

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
mwestacott: I like variable names to be as specific as possible, but up to you.

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
lduranteau: I agree with the name getValidationErrors

src/validation/async/index.ts
lduranteau: or I could put them in a separate file
because they are not async but require data to be fetched asynchronously

src/validation/async/index.ts
```diff
@@ -0,0 +1,260 @@
...
+      msg: ErrorMessages.TRUST_STILL_INVOLVED,
+      param: 'stillInvolved',
+      location: 'body',
+    });
+  }
+
+  return errors;
+};
+
+export const checkTrustLegalEntityBeneficialOwnerStillInvolved = (appData: ApplicationData, req): ValidationError[] => {
```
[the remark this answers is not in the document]
lduranteau: see the discussion of the comment on line 13

src/validation/async/index.ts
markpit: A separate file might be better, as I was expecting this file to contain only async functions.

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
lduranteau: name changed to getValidationErrors

src/validation/async/index.ts
lduranteau: stillInvolved validators have been moved to a separate file

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
mwestacott: 👍

mwestacott (commented):
LGTM now but there are conflicts to resolve.

markpit (approved):

dlloyd1-ch (commented):
Looks good. One minor comment.

src/controllers/update/update.manage.trusts.tell.us.about.the.former.bo.controller.ts
```diff
@@ -50,8 +51,12 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 
     // check for form validation errors
     const errorList = validationResult(req);
-    if (!errorList.isEmpty()) {
-      const pageProperties = getPageProperties(trust, formData, formatValidationError(errorList.array()));
+    const errors = await getValidationErrors(req);
```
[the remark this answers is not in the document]
lduranteau: I am not a big fan of putting comments everywhere
This is why @mwestacott's suggestion to change the name made sense
Do you think the name isn't explicit enough?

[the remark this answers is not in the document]
dlloyd1-ch: We have 
```
const errorList = validationResult(req);
const errors = await getValidationErrors(req);
```
Will it be obvious in 6 months time why we have two validation functions in each controller?

lduranteau:
> LGTM now but there are conflicts to resolve.

resolved

mwestacott (approved):

dlloyd1-ch (approved):

PR #1590, opened 2024-09-16
uar-1618 - trustee signout link no selection

lduranteau:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1618

### Change description
Fix trustee - not selecting an answer to 'Are you sure you wish to sign out?' results in error message (500)

### Work checklist

### Merge instructions

```
```


src/middleware/validation.middleware.ts
```diff
@@ -48,7 +48,7 @@ export const checkValidations = async (req: Request, res: Response, next: NextFu
         [ResignedOnDateKey]: prepareData(req.body, ResignedOnDateKeys),
       };
 
-      const routePath = req.route.path;
+      const routePath = req.route.path.includes(config.DYNAMIC_SUB_PATH) ? req.route.path.replace(config.DYNAMIC_SUB_PATH, "") : req.route.path;
```
markpit, line 51: there is also a req.route.path in the checkTrustValidations function, does that need to be updated as well?

lduranteau: This was a problem here because the trustee pages have part of the url dynamic, I don't think this is the case for the url using checkTrustValidations

markpit (approved):

mwestacott (commented):
Is it possible to write a unit test that replicates the bug which would then pass as a result of this fix?

lduranteau:
> Is it possible to write a unit test that replicates the bug which would then pass as a result of this fix?

I can't create a test replicating the bug because it is fixed in the code.
Whatever input I pass from the test, the code is supposed to handle it and avoid a bug crash.

I can add a test with an incorrect input and check that I have a 200.

mwestacott:
> > Is it possible to write a unit test that replicates the bug which would then pass as a result of this fix?
> 
> I can't create a test replicating the bug because it is fixed in the code. Whatever input I pass from the test, the code is supposed to handle it and avoid a bug crash.
> 
> I can add a test with an incorrect input and check that I have a 200.

So yes, test would be to check for the 200 and it would pass (and if you temporarily reverted your fix, presumably the test would then fail).

lduranteau:
> > > Is it possible to write a unit test that replicates the bug which would then pass as a result of this fix?
> > 
> > 
> > I can't create a test replicating the bug because it is fixed in the code. Whatever input I pass from the test, the code is supposed to handle it and avoid a bug crash.
> > I can add a test with an incorrect input and check that I have a 200.
> 
> So yes, test would be to check for the 200 when no input is selected and it would pass (and if you temporarily reverted your fix, presumably the test would then fail).

I tried various techniques and read a lot of documentation to write this test but it is impossible.
The logic is in req.route.path, so we need to be able to manipulate the Request object. But Supertest doesn't allow us to manipulate it before or during the request.
Because validation takes place before the controller, it is impossible to have a real unit test from the controller.

mwestacott (approved):

PR #1594, opened 2024-09-17
Fix/ROE-2492: Ensure sign out urls for trust screens have entity ID and submission ID in url

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2492

### Change description

 Ensure sign out urls for trust screens have entity ID and submission ID in url

### Work checklist


src/utils/url.ts
```diff
@@ -27,6 +27,22 @@ export const transactionIdAndSubmissionIdExistInRequest = (req: Request): boolea
 const getTransactionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_TRANSACTION_ID];
 const getSubmissionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_SUBMISSION_ID];
 
+export const isRegistrationJourney = (req: Request): boolean => {
+  const fullPath: string = `${req.baseUrl}${req.path}`;
+  if (fullPath.indexOf(config.LANDING_URL) === 0) {
```
markpit, line 32: instead of indexOf === 0, is it possible to use startsWith() instead?

src/middleware/navigation/sign.out.base.url.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as config from "../../config";
+import { logger } from "../../utils/logger";
+import {
+  isRegistrationJourney,
+  getUrlWithParamsToPath,
+  transactionIdAndSubmissionIdExistInRequest
+} from "../../utils/url";
+
+export const generateSignOutBaseUrl = (req: Request, res: Response, next: NextFunction) => {
+  logger.info("setting sign-our-base-url");
```
markpit, line 16: can you use logger.infoRequest(req, "setting sign-our-base-url"); instead of logger.info

src/utils/url.ts
```diff
@@ -27,6 +32,44 @@ export const transactionIdAndSubmissionIdExistInRequest = (req: Request): boolea
 const getTransactionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_TRANSACTION_ID];
 const getSubmissionIdFromRequestParams = (req: Request): string => req.params[config.ROUTE_PARAM_SUBMISSION_ID];
 
+/**
+ * This is required to extract the transactionId and submissionId prior to the request reaching its intended mount path
+ */
+export const getTransactionIdAndSubmissionIdFromOriginalUrl = (req: Request): TransactionIdAndSubmissionId | undefined => {
+  try {
+    logger.infoRequest(req, `extracting transactionId and submissionId from the req.originalUrl: ${req.originalUrl}`);
+    if (!req.originalUrl.includes('/transaction/') || !req.originalUrl.includes('/submission/')) {
```
markpit, line 41: for the paths like '/submission/'  and 'submission' , '/transaction/' etc.. can they be added to the src/config/index.ts

src/utils/url.ts
```diff
@@ -27,6 +32,44 @@ export const transactionIdAndSubmissionIdExistInRequest = (req: Request): boolea
...
+    const submissionIndex = elements.indexOf("submission") + 1;
+    if (!elements[transactionIndex] || !elements[submissionIndex]) {
+      return;
+    }
+    return {
+      transactionId: elements[transactionIndex],
+      submissionId: elements[submissionIndex],
+    };
+  } catch (e) {
+    logger.infoRequest(req, `error extracting transactionId and submissionId from req.originalUrl: ${req.originalUrl} with error object: ${e}`);
```
markpit, line 55: what happens to the execution if an error is caught, it doesn't seem to return or throw and I have a feeling this might cause the web journey to hang. Could we re-throw the error (or not catch it) so the errorHandler controller could catch it and display the 'service unavailable page'?

markpit: Also, should logger.infoRequest be logger.errorRequest if it is a non recoverable error

src/utils/url.ts
mwejuli-ch: Agreed. This has been updated now.

src/utils/url.ts
mwejuli-ch: In the unlikely event that this happens, we catch, log, and return `undefined` so that the calling method uses the fallback baseUrl.. no need to re-throw as we need to set the fallback baseUrl and proceed to `next`..

Yes, `logger.infoRequest` should be `logger.errorRequest`. This has been updated. Thanks for the good spot.

src/config/index.ts
```diff
@@ -493,3 +493,5 @@ export const PAYMENT_WITH_TRANSACTION_WITH_PARAMS_URL = REGISTER_AN_OVERSEAS_ENT
 // Trusts with id params
 export const TRUST_ENTRY_WITH_PARAMS_URL = REGISTER_AN_OVERSEAS_ENTITY_WITH_PARAMS_URL + TRUSTS_URL;
 export const TRUST_INFO_WITH_PARAMS_URL = REGISTER_AN_OVERSEAS_ENTITY_WITH_PARAMS_URL + TRUST_INFO_PAGE;
+export const TRANSACTION_ID_URL_KEY = "/transaction/";
```
mwestacott, line 496: Is this value (and one immediately below) really a 'key' or is it perhaps a sub-path within the URL?

test/utils/url.spec.ts
```diff
@@ -81,6 +84,131 @@ describe("Url utils tests", () => {
     });
   });
 
+  describe("getTransactionIdAndSubmissionIdFromOriginalUrl tests", () => {
+
+    const urlPathWithIds = "/transaction/123/submission/345";
+    const urlPathWithoutIds = "/path/without/ids";
+    const urlPathWithMissingTransactionId = "/trans/123/submission/345";
+    const urlPathWithMissingSubmissionId = "/transaction/123/sub/345";
```
mwestacott, line 92: How does the function cope with a URL like "/transaction/submission/no-ids" or perhaps first, what should the function return for this URL?

src/middleware/navigation/remove/remove.journey.middleware.ts
```diff
@@ -7,15 +7,28 @@ import { NextFunction, Request, Response } from "express";
 import {
   JourneyType
 } from "../../../config";
+
 import { logger } from "../../../utils/logger";
-import { isRemoveJourney } from "../../../utils/url";
+import { isRegistrationJourney, isRemoveJourney, isUpdateJourney } from "../../../utils/url";
 
+/**
+ * @todo: rename method to be less about remove journey and more about all journeys and refactor affected code
```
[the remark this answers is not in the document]
mwejuli-ch: You'll be pleased to know that the rename/refactor exercise has now been done as part of this PR.

src/middleware/navigation/remove/remove.journey.middleware.ts
```diff
@@ -7,15 +7,28 @@ import { NextFunction, Request, Response } from "express";
...
+  const isUpdate: boolean = await isUpdateJourney(req);
+  const isRegistration: boolean = isRegistrationJourney(req);
 
   if (isRemove) {
     logger.infoRequest(req, "Marking this request/response as a Remove Journey");
     res.locals.journey = JourneyType.remove;
+  } else if (isUpdate) {
+    logger.infoRequest(req, "Marking this request/response as a Update Journey");
+    res.locals.journey = JourneyType.update;
+  } else if (isRegistration) {
```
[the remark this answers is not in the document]
mwejuli-ch: Agreed. This has been refactored.

src/middleware/navigation/sign.out.base.url.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import * as config from "../../config";
+import { logger } from "../../utils/logger";
+import {
+  isRegistrationJourney,
+  getTransactionIdAndSubmissionIdFromOriginalUrl,
+  getUrlWithTransactionIdAndSubmissionId
+} from "../../utils/url";
+
+export const generateSignOutBaseUrl = (req: Request, res: Response, next: NextFunction) => {
```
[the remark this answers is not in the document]
mwejuli-ch: For consistency with method name, have changed the log message to say "Generating".

src/middleware/navigation/sign.out.base.url.ts
```diff
@@ -0,0 +1,31 @@
...
+  logger.info("setting sign-out-base-url");
+  let signOutBaseUrl: string = config.UPDATE_AN_OVERSEAS_ENTITY_URL;
+  if (isRegistrationJourney(req)) {
+    const transactionIdAndSubmissionId = getTransactionIdAndSubmissionIdFromOriginalUrl(req);
```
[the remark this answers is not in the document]
mwejuli-ch: No check for this flag is required here as we use the existence of the key/value pairs for `transaction` and `submission` to determine what our signOutBaseUrl should be.

test/utils/url.spec.ts
mwejuli-ch: Excellent spot @mwestacott. Have refactored this piece of code to gracefully ignore this type of malformed key/value pairing in the URL. There is a unit test to also check for this too.

src/middleware/navigation/sign.out.base.url.ts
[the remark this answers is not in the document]
mwestacott: But it looks like we're deciding what to do based on the presence of the ids. Normally these decisions are based on the feature flag check. Later, when we remove the feature flag (and search the code-base) we find all the occurrences and know what to change. That won't be the case here. I still think fundamentally that the decision should be based on the feature flag and if it's an Update/Remove journey with Redis removal enabled and no ids can be found then the code should error. At the moment, it won't error, it will set REGISTER_AN_OVERSEAS_ENTITY_URL as the URL.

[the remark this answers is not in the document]
mwestacott: Had a chat and now convinced. Marking as resolved.

mwestacott:
LGTM

markpit (approved):

lduranteau (approved):

PR #1619, opened 2024-10-02
Feat/ROE-2619: Fetch application data from DB for `sold-land-filter` page

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2619

### Change description

- Retrieves application data from DB for `sold-land-filter` page
- Updates `getApplicationData` method to fetch data from API
- Includes formatting changes for legibility from previously modified code

### Work checklist

### DIFF
--- a/src/controllers/sold.land.filter.controller.ts
+++ b/src/controllers/sold.land.filter.controller.ts
@@ -22,8 +22,7 @@ export const get = async (req: Request, res: Response, next: NextFunction): Prom
       deleteApplicationData(req.session);
     }
 
-    const session = req.session as Session;
-    const appData: ApplicationData = await getApplicationData(session, req);
+    const appData: ApplicationData = await getApplicationData(req);
 
     return res.render(config.SOLD_LAND_FILTER_PAGE, {
       backLinkUrl: getSoldLandFilterBackLink(),
@@ -45,7 +44,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 
     const session = req.session as Session;
     const hasSoldLand = (req.body[HasSoldLandKey]).toString();
-    const appData: ApplicationData = await getApplicationData(session, req);
+    const appData: ApplicationData = await getApplicationData(req);
     appData[HasSoldLandKey] = hasSoldLand;
 
     let nextPageUrl: string = "";
--- a/src/controllers/sold.land.filter.controller.ts
+++ b/src/controllers/sold.land.filter.controller.ts
@@ -13,20 +13,27 @@ import { Session } from "@companieshouse/node-session-handler";
 import { getUrlWithTransactionIdAndSubmissionId } from "../utils/url";
 
 export const get = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+
   try {
+
     logger.debugRequest(req, `GET ${config.SOLD_LAND_FILTER_PAGE}`);
 
     if (req.query[config.LANDING_PAGE_QUERY_PARAM] === '0') {
       deleteApplicationData(req.session);
     }
 
-    const appData: ApplicationData = await getApplicationData(req.session);
+    const session = req.session as Session;
+    const appData: ApplicationData = await getApplicationData(session, req);
 
+    console.log("<<<<<<<<<<<<<<<<<<<<<<<<GET>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>");
+    console.log(appData);
+    console.log("<<<<<<<<<<<<<<<<<<<<<<<<GET>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>");
     return res.render(config.SOLD_LAND_FILTER_PAGE, {
       backLinkUrl: getSoldLandFilterBackLink(),
       templateName: config.SOLD_LAND_FILTER_PAGE,
       [HasSoldLandKey]: appData?.[HasSoldLandKey]
     });
+
   } catch (error) {
     logger.errorRequest(req, error);
     next(error);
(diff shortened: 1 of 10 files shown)


src/controllers/sold.land.filter.controller.ts
```diff
@@ -13,46 +13,55 @@ import { Session } from "@companieshouse/node-session-handler";
...
+
     logger.debugRequest(req, `GET ${config.SOLD_LAND_FILTER_PAGE}`);
 
     if (req.query[config.LANDING_PAGE_QUERY_PARAM] === '0') {
       deleteApplicationData(req.session);
     }
 
-    const appData: ApplicationData = await getApplicationData(req.session);
+    const session = req.session as Session;
+    const appData: ApplicationData = await getApplicationData(session, req);
```
mwestacott, line 26: It looks a bit odd calling this method with `session` and `request` when the request contains the session. Is method overloading allowed in JS/Node and could we have a separate `getApplicationData()` function that just took a `request` and it extracted the `session`?

src/utils/application.data.ts
```diff
@@ -10,41 +10,75 @@ import {
...
+  ROUTE_PARAM_SUBMISSION_ID
 } from '../config';
+
 import { BeneficialOwnerCorporate } from '@companieshouse/api-sdk-node/dist/services/overseas-entities';
 import { Remove } from 'model/remove.type.model';
 import { isActiveFeature } from "./feature.flag";
 import { isNoChangeJourney } from "./update/no.change.journey";
+import { getOverseasEntity } from "../service/overseas.entities.service";
+
+export const getApplicationData = async (session: Session | undefined, req?: Request): Promise<ApplicationData> => {
```
mwestacott, line 49: See earlier comment regarding this function signature change. Would an overloaded function with different params be better perhaps?

mwestacott (commented):
There are quite a few additional empty lines being added in a number of files/places. Just wondering why? It adds some noise to the PR when reviewing.

src/utils/application.data.ts
```diff
@@ -10,41 +10,75 @@ import {
...
+    if (!isActiveFeature(FEATURE_FLAG_ENABLE_REDIS_REMOVAL) || !req) {
+      return session?.getExtraData(APPLICATION_DATA_KEY) || {} as ApplicationData;
+    }
 
-export const getApplicationData = async (session: Session | undefined): Promise<ApplicationData> => {
-  return await session?.getExtraData(APPLICATION_DATA_KEY) || {} as ApplicationData;
+    const transactionId: string = req.params[ROUTE_PARAM_TRANSACTION_ID] ?? "";
+    const submissionId: string = req.params[ROUTE_PARAM_SUBMISSION_ID] ?? "";
+
+    if (transactionId === "" || submissionId === "") {
```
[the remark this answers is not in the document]
lduranteau: I assume that this will evolve because at the moment it is just a question of registration journey

src/utils/application.data.ts
lduranteau: I think that the first parameter will be deleted once finished so that only req stays

mwejuli-ch: Yes, have updated this now to use a single parameter as per your previous comment.

src/utils/application.data.ts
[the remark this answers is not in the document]
mwejuli-ch: I believe we can only get data from Mongo if we have both the `transactionId` and the `submissionId`. Am not sure we can get it any other way? I could be wrong..

mwejuli-ch:
> There are quite a few additional empty lines being added in a number of files/places. Just wondering why? It adds some noise to the PR when reviewing.

These are only formatting changes for legibility as I had missed making them in previous commits. They probably make more sense when viewed in your editor. Happy to discuss.

src/utils/application.data.ts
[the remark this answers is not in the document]
mwestacott: No, I think that's right. In terms of getting OE application data the transaction needs to be created and the submission, so both ids are needed.

src/utils/application.data.ts
```diff
@@ -46,19 +46,29 @@ import { isActiveFeature } from "./feature.flag";
 import { isNoChangeJourney } from "./update/no.change.journey";
 import { getOverseasEntity } from "../service/overseas.entities.service";
 
-export const getApplicationData = async (session: Session | undefined, req?: Request): Promise<ApplicationData> => {
+export const getApplicationData = async (sessionOrRequest: Session | Request | undefined): Promise<ApplicationData> => {
+
+  const defaultAppData = {};
```
mwestacott, line 51: Perhaps call this variable `emptyAppData`.

lduranteau (approved):

test/middleware/authentication.middleware.spec.ts
```diff
@@ -174,6 +175,8 @@ describe('Authentication middleware', () => {
   test("should redirect to signin page for update", async () => {
     const signinRedirectPath = `/signin?return_to=${SECURE_UPDATE_FILTER_URL}`;
 
+    mockIsActiveFeature.mockReturnValueOnce(false);
+    mockIsActiveFeature.mockReturnValueOnce(false);
```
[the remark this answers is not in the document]
mwejuli-ch: Comments added.

src/utils/application.data.ts
mwejuli-ch: Updated.

mwestacott: Thanks

test/middleware/authentication.middleware.spec.ts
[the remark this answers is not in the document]
dlloyd1-ch: ok, I'm a bit confused. Why would we mock the redis flag to return `false` and then `true` in the same test?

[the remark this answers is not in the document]
mwejuli-ch: `false` for the first occurrence, and `true` for the next one . Jest applies them in the order in which they show up in the execution sequence.

test/middleware/authentication.middleware.spec.ts
```diff
@@ -174,8 +175,10 @@ describe('Authentication middleware', () => {
   test("should redirect to signin page for update", async () => {
     const signinRedirectPath = `/signin?return_to=${SECURE_UPDATE_FILTER_URL}`;
 
-    mockIsActiveFeature.mockReturnValueOnce(false);
-    mockIsActiveFeature.mockReturnValueOnce(true);
+    mockIsActiveFeature.mockReturnValueOnce(false); // SHOW_SERVICE_OFFLINE_PAGE
+    mockIsActiveFeature.mockReturnValueOnce(false); // FEATURE_FLAG_ENABLE_REDIS_REMOVAL
+    mockIsActiveFeature.mockReturnValueOnce(false); // FEATURE_FLAG_ENABLE_REDIS_REMOVAL
+    mockIsActiveFeature.mockReturnValueOnce(true); // FEATURE_FLAG_ENABLE_REDIS_REMOVAL
```
mwestacott, line 181: Like @dlloyd1-ch , also confused. Why would the value of the Redis feature flag ever change from being `false` to `true` within the scope of a single request-response cycle?

test/middleware/authentication.middleware.spec.ts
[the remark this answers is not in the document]
dlloyd1-ch: Why would it change for each occurrence? Surely it's either on or off.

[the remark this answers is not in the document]
mwejuli-ch: yes, you're right.. it's the same for every test scope. the last one was unnecessary which is why the test was passing. fixed now.

test/middleware/authentication.middleware.spec.ts
mwejuli-ch: it's the same for every test scope. the last one was unnecessary which is why the test was passing. fixed now.

mwestacott (approved):

dlloyd1-ch (approved):

PR #1623, opened 2024-10-07
Feat/ROE-2636: Retrieve application data from DB for `secure-register-filter page`

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2636

### Change description

- Retrieves application data from the API for `secure-register-filter page`

### Work checklist

### DIFF
--- a/src/utils/secure.filter.ts
+++ b/src/utils/secure.filter.ts
@@ -3,17 +3,27 @@ import { ApplicationData } from "../model";
 import { IsSecureRegisterKey, OverseasEntityKey, Transactionkey } from "../model/data.types.model";
 import { getApplicationData, setExtraData } from "./application.data";
 import { logger } from "./logger";
-import { getUrlWithTransactionIdAndSubmissionId, isRemoveJourney } from "../utils/url";
+import { getUrlWithTransactionIdAndSubmissionId, isRemoveJourney, isRegistrationJourney } from "../utils/url";
 import * as config from "../config";
 import { isActiveFeature } from "./feature.flag";
 import { updateOverseasEntity } from "../service/overseas.entities.service";
 import { Session } from "@companieshouse/node-session-handler";
 
 export const getFilterPage = async (req: Request, res: Response, next: NextFunction, templateName: string, backLinkUrl: string): Promise<void> => {
+
   try {
+
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
-    const appData: ApplicationData = await getApplicationData(req.session);
-    const isRemove: boolean = await isRemoveJourney(req);
+
+    const isRegistration = isRegistrationJourney(req);
+    const isRemove = await isRemoveJourney(req);
+    let appData: ApplicationData;
+
+    if (isRegistration) {
+      appData = await getApplicationData(req);
+    } else {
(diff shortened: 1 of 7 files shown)


src/utils/secure.filter.ts
```diff
@@ -3,17 +3,27 @@ import { ApplicationData } from "../model";
...
 
 export const getFilterPage = async (req: Request, res: Response, next: NextFunction, templateName: string, backLinkUrl: string): Promise<void> => {
+
   try {
+
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
-    const appData: ApplicationData = await getApplicationData(req.session);
-    const isRemove: boolean = await isRemoveJourney(req);
+
+    const isRegistration = isRegistrationJourney(req);
```
lduranteau, line 18: if you often need to check the different journeys in the same place
maybe having a function returning the status of each one would be simpler

{
   isRegistration: true,
   isUpdate: false,
   isRemove: false
}

mwejuli-ch: Ok, I see your point. The only issue is that this is temporary code and will be deleted when the REDIS ticket for the Update journey is done. Only the existing `isRemove` check will remain (as before).

lduranteau: it was a proposal
if we don't have the case elsewhere in the code, we certainly don't need it

mwejuli-ch: Because of a code smell from Sonar, I have refactored the `getApplicationData` piece that is common to both methods into a separate, non-exportable local method called `fetchApplicationData`. But this is also temporary until the Update journey ticket is done.

markpit (approved):

lduranteau (approved):

src/utils/secure.filter.ts
```diff
@@ -3,17 +3,29 @@ import { ApplicationData } from "../model";
...
 import { getApplicationData, setExtraData } from "./application.data";
 import { logger } from "./logger";
-import { getUrlWithTransactionIdAndSubmissionId, isRemoveJourney } from "../utils/url";
+import { getUrlWithTransactionIdAndSubmissionId, isRemoveJourney, isRegistrationJourney } from "../utils/url";
 import * as config from "../config";
 import { isActiveFeature } from "./feature.flag";
 import { updateOverseasEntity } from "../service/overseas.entities.service";
 import { Session } from "@companieshouse/node-session-handler";
 
+const fetchApplicationData = (req: Request, isRegistration: boolean): Promise<ApplicationData> => {
```
mwestacott, line 12: Can't be sure from looking at the code in GitHub, but could you potentially just do the line:

`const isRegistration = isRegistrationJourney(req);
`
within this function and avoid the need for the second `isRegistration` parameter?

mwejuli-ch: Oh yes, thought about that, but...

1) We need that value elsewhere in the POST handler so makes sense running it once and passing it as a parameter, rather than running it twice.
2) It's temporary code that will be deleted once the Update portion of `secure-register-filter` is done so has only got a pretty short TTL.

mwestacott (approved):

PR #1627, opened 2024-10-08
Feat/ROE-2654: Retrieve app data from the DB for the `overseas-entity-name` page

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2654

### Change description

- Retrieves application data from the API for the `overseas-entity-name` page/screen

### Work checklist

### DIFF
--- a/src/controllers/overseas.name.controller.ts
+++ b/src/controllers/overseas.name.controller.ts
@@ -1,31 +1,44 @@
 import { NextFunction, Request, Response } from "express";
 import { Session } from "@companieshouse/node-session-handler";
-
 import * as config from "../config";
 import { ApplicationData } from "../model";
 import { logger } from "../utils/logger";
 import { isActiveFeature } from "../utils/feature.flag";
-import { getApplicationData, setExtraData } from "../utils/application.data";
 import { postTransaction } from "../service/transaction.service";
+import { getApplicationData, setExtraData } from "../utils/application.data";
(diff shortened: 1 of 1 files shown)


src/controllers/overseas.name.controller.ts
```diff
@@ -1,52 +1,69 @@
...
+} from "../utils/url";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
+
   try {
+
     logger.debugRequest(req, `GET ${config.OVERSEAS_NAME_PAGE}`);
-    const appData: ApplicationData = await getApplicationData(req.session);
 
+    const appData: ApplicationData = await getApplicationData(req);
```
mwestacott, line 29: A fair few formatting changes and other miscellaneous clean-ups. Is this the only 'real' change to the functionality?

Also, does it - or rather should it - have a corresponding unit test change? I'd have though we mock out the call to `getApplicationData`, so wondering how existing tests are passing if param has changed.

markpit (approved):

src/controllers/overseas.name.controller.ts
mwejuli-ch: - The 'real' change is that the call to `getApplicationData` is  passing in the full request and not the session
- No need to add a unit test coz it's already covered by the mock

lduranteau (approved):

PR #1631, opened 2024-10-14
Feature/uar 1635 save resume deselect nocs

lduranteau:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1635

### Change description
Save and Resume: Deselect invalid NOCs when returning to web journey after new NOCs are implemented.
Add error message if nature of control is an empty list

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/controllers/beneficial.owner.type.controller.ts
+++ b/src/controllers/beneficial.owner.type.controller.ts
@@ -12,35 +12,19 @@ import {
 } from "../model/beneficial.owner.type.model";
 import { isActiveFeature } from '../utils/feature.flag';
 import { getUrlWithParamsToPath } from "../utils/url";
+import { beneficialOwnersTypeEmptyNOCList } from "../validation/async";
+import { FormattedValidationErrors, formatValidationError, } from "../middleware/validation.middleware";
+import { ValidationError } from "express-validator";
+import { BeneficialOwnerIndividualKey } from "../model/beneficial.owner.individual.model";
+import { BeneficialOwnerOtherKey } from "../model/beneficial.owner.other.model";
+import { BeneficialOwnerGovKey } from "../model/beneficial.owner.gov.model";
+import { NonLegalFirmNoc } from "../model/data.types.model";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   try {
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
 
-    const appData: ApplicationData = await getApplicationData(req.session);
-    const requiresTrusts: boolean = checkEntityRequiresTrusts(appData);
-
-    logger.infoRequest(req, `${config.BENEFICIAL_OWNER_TYPE_PAGE} requiresTrusts=${requiresTrusts}`);
-
-    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL)) {
-      return res.render(config.BENEFICIAL_OWNER_TYPE_PAGE, {
-        // Even though the value of this feature flag is available in the template via the OE_CONFIGS variable, passing it in
-        // like this enables unit tests to assert different outcomes, based on whether it is set or not
-        FEATURE_FLAG_ENABLE_REDIS_REMOVAL: true,
-        activeSubmissionBasePath: getUrlWithParamsToPath(config.ACTIVE_SUBMISSION_BASE_PATH, req),
-        backLinkUrl: getUrlWithParamsToPath(config.BENEFICIAL_OWNER_STATEMENTS_WITH_PARAMS_URL, req),
-        templateName: config.BENEFICIAL_OWNER_TYPE_PAGE,
-        requiresTrusts,
-        ...appData,
-      });
-    }
-
-    return res.render(config.BENEFICIAL_OWNER_TYPE_PAGE, {
-      backLinkUrl: config.BENEFICIAL_OWNER_STATEMENTS_URL,
-      templateName: config.BENEFICIAL_OWNER_TYPE_PAGE,
-      requiresTrusts,
-      ...appData,
-    });
+    return await getPageRender(req, res);
   } catch (error) {
     logger.errorRequest(req, error);
     next(error);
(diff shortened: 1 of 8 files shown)


src/validation/async/index.ts
```diff
@@ -152,3 +155,55 @@ export const filingPeriodTrustStartDateValidations = async (req: Request) => is_
...
+
+  try {
+    const boiList = checkIfBeneficialOwnerHasNOC(appData?.[BeneficialOwnerIndividualKey] ?? []);
+    const booList = checkIfBeneficialOwnerHasNOC(appData?.[BeneficialOwnerOtherKey] ?? []);
+    const bogList = checkIfBeneficialOwnerHasNOC(appData?.[BeneficialOwnerGovKey] ?? []);
+
+    const boList: string[] = [...boiList, ...booList, ...bogList];
+
+    if (boList.length) {
+      throw new Error(`
```
lduranteau, line 180: The message has not yet been defined

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -57,14 +41,30 @@ export const postSubmit = async (req: Request, res: Response) => {
...
   }
+
   if (requiresTrusts) {
     nextPageUrl = isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUSTS_WEB)
       ? getTrustLandingUrl(appData, req)
       : config.TRUST_INFO_URL;
   }
+
+  if (isActiveFeature(config.FEATURE_FLAG_ENABLE_PROPERTY_OR_LAND_OWNER_NOC)) {
+    console.debug("Removing old NOCs");
```
markpit, line 62: we should use logger instead of console, there should be a logger.debugRequest that will also log the request id to help identify it in the logs.

src/controllers/update/beneficial.owner.type.controller.ts
```diff
@@ -116,6 +113,13 @@ export const postSubmit = async (req: Request, res: Response, next: NextFunction
     moveReviewableTrustsIntoReview(appData);
     resetReviewStatusOnAllTrustsToBeReviewed(appData);
 
+    if (isActiveFeature(config.FEATURE_FLAG_ENABLE_PROPERTY_OR_LAND_OWNER_NOC)) {
+      console.debug("Removing old NOCs");
```
markpit, line 117: console.debug -> logger.debugRequest

src/validation/async/index.ts
markpit: can you move the message to src/validation/error.messages.ts, it's where we keep them all

src/controllers/beneficial.owner.type.controller.ts
lduranteau: changed to logger.debugRequest

src/controllers/update/beneficial.owner.type.controller.ts
lduranteau: changed to logger.debugRequest

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -96,3 +96,39 @@ const getNextPage = (req: Request): string => {
...
 };
+
+// Get validation errors that depend on an asynchronous request
+const getValidationErrors = async (appData: ApplicationData, req: Request): Promise<ValidationError[]> => {
+  const beneficialOwnersTypeEmptyNOCListErrors = await beneficialOwnersTypeEmptyNOCList(req, appData);
+
+  return [...beneficialOwnersTypeEmptyNOCListErrors];
+};
+
+const getPageRender = async (req: Request, res: Response, errors?: FormattedValidationErrors) => {
```
mwestacott, line 107: Suggest calling this method just `renderPage`.

src/controllers/update/beneficial.owner.type.controller.ts
```diff
@@ -152,3 +156,28 @@ const getNextPage = (beneficialOwnerTypeChoices: BeneficialOwnerTypeChoice | Man
...
 };
+
+// Get validation errors that depend on an asynchronous request
+const getValidationErrors = async (appData: ApplicationData, req: Request): Promise<ValidationError[]> => {
+  const beneficialOwnersTypeEmptyNOCListErrors = await beneficialOwnersTypeEmptyNOCList(req, appData);
+
+  return [...beneficialOwnersTypeEmptyNOCListErrors];
+};
+
+const getPageRender = async (req: Request, res: Response, errors?: FormattedValidationErrors) => {
```
mwestacott, line 167: Again, just `renderPage`? Although... this function either redirects or renders depending on some checks, so maybe `redirectOrRenderPage`?

src/validation/async/index.ts
```diff
@@ -152,3 +156,57 @@ export const filingPeriodTrustStartDateValidations = async (req: Request) => is_
...
+  const allowed: boolean = isAllowedUrls(allowedUrls, req);
+
+  const errors: ValidationError[] = [];
+
+  if (!allowed) {
+    return errors;
+  }
+
+  try {
+    const isActive = isActiveFeature(config.FEATURE_FLAG_ENABLE_PROPERTY_OR_LAND_OWNER_NOC);
```
mwestacott, line 174: You could move this line into the `checkIfBeneficialOwnerHasNOC` function and avoid having to pass it in as a parameter - it looks like no other code in this function requires it.

src/validation/error.messages.ts
```diff
@@ -37,6 +37,7 @@ export enum ErrorMessages {
   START_DATE_MUST_BE_AFTER_DOB = "Start date must be on or after date of birth",
   UNABLE_TO_RETRIEVE_ENTITY_NUMBER = "Unable to retrieve entity number",
   UNABLE_TO_RETRIEVE_EXPECTED_DATE = "Unable to retrieve the entity's expected date",
+  MISSING_NATURE_OF_CONTROL = "Natures of control options may have changed since you the last time you saved your update statement. Select Change for each beneficial owner to check their natures of control are correct:",
```
mwestacott, line 40: This message doesn't match the text in AC5, which will need updating. Is it agreed that we will now show a list of BOs?

src/validation/async/index.ts
```diff
@@ -152,3 +156,57 @@ export const filingPeriodTrustStartDateValidations = async (req: Request) => is_
...
+      location: 'body',
+    });
+
+    return errors;
+  }
+};
+
+const checkIfBeneficialOwnerHasNOC = (beneficialOwner, isActive: boolean): string[] => {
+  return beneficialOwner?.filter(bo =>
+    !bo?.beneficial_owner_nature_of_control_types?.length &&
```
mwestacott, line 201: Formatting looks a bit odd on this line - only 4 spaces instead of 8, or is that GitHub?

src/validation/error.messages.ts
lduranteau: The message is in the comment

src/validation/async/index.ts
lduranteau: There is no formatting issue

src/validation/async/index.ts
lduranteau: It was there but I moved it here because otherwise I have to mock the isActiveFeature call at each iteration

src/validation/error.messages.ts
mwestacott: Message in the comment doesn't have the colon (:) and I think we need to make clear in the AC what the error message will look like when we have multiple BOs of different types that have missing NOCs - i.e. what's displayed, first name, last name, company name.

src/validation/async/index.ts
mwestacott: Line 201 has 4 spaces and lines 206 to 209 6. Is that intended/correct?

src/validation/error.messages.ts
lduranteau: AC5 updated

mwestacott (approved):
Looks good.

Suggest adding a comment to the JIRA story and tagging testers letting them know that the Registration behaviour has also changed (fixes that bug you found). They'll need to regression test that journey now too.

src/controllers/beneficial.owner.type.controller.ts
```diff
@@ -12,35 +12,19 @@ import {
 } from "../model/beneficial.owner.type.model";
 import { isActiveFeature } from '../utils/feature.flag';
 import { getUrlWithParamsToPath } from "../utils/url";
+import { beneficialOwnersTypeEmptyNOCList } from "../validation/async";
```
mwejuli-ch, line 15: If we're using camel-case, shouldn't this be `beneficialOwnersTypeEmptyNocList` rather than `beneficialOwnersTypeEmptyNOCList`? This is a non-blocking comment so should not hold up merging 🙂

lduranteau: NOC is an acronym, not a real word

mwejuli-ch: Yes, conventionally, acronyms are also camel-cased.. but it's not a biggie anyway. Will resolve comment

mwejuli-ch (approved):

PR #1686, opened 2024-12-11
feat/roe-2675: add change links with entity IDs in url

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2675

### Change description

- Adds change links for the `check-your-answers` page with the the entity IDs (transaction & submission) substituted in.
- Adds missing address fields and type fields in trust forms and on the `check-your-answers` page
- Includes some formatting tweaks

### Work checklist


src/app.ts
```diff
@@ -47,7 +53,9 @@ nunjucksEnv.addGlobal("OE_CONFIGS", config);
 nunjucksEnv.addGlobal("ERROR_MESSAGES", ErrorMessages);
 nunjucksEnv.addGlobal("COUNTRY_FILTER", countryFilter );
 nunjucksEnv.addGlobal("CREATE_CHANGE_LINK", createChangeLinkConfig);
+nunjucksEnv.addGlobal("CREATE_CHANGE_LINK_WITH_IDs", createChangeLinkWithIds);
 nunjucksEnv.addGlobal("SUMMARY_LIST_LINK", createSummaryListLink);
+nunjucksEnv.addGlobal("IS_REDIS_REMOVAL_ENABLED", isActiveFeature(config.FEATURE_FLAG_ENABLE_TRUSTS_WEB));
```
markpit, line 58: the config object is already passed to the template and the templates can already use the redis feature flag - https://github.com/companieshouse/overseas-entities-web/blob/main/views/beneficial-owner-gov.html#L16
Adding another call to isActiveFeature here will mean a lot of tests will need updating.

src/model/trust.model.ts
```diff
@@ -89,6 +89,8 @@ export interface TrustIndividual {
   ceased_date_month?: string;
   ceased_date_year?: string;
   relevant_period?: boolean;
+  usual_residential_address?: Address
```
markpit, line 92: not sure why we need to add the address to the trusts model? it already has ura and sa address fields

src/app.ts
mwejuli-ch: Just tested this locally with `FEATURE_FLAG_ENABLE_REDIS_REMOVAL` in the view and it's not working. Happy to refactor to use it if it works. Perhaps am using it wrong?

src/model/trust.model.ts
mwejuli-ch: With the flag set to `true`, "ura" and "sa" fields where all resolving to `undefined` or `null`. Adding "usual_residential_address" and "service_address" field objects that map to the DB values resolved this.  The refactor now works with either value for the flag.

I could have missed something so happy to go over this with you whenever you have a few moments to spare.

src/app.ts
markpit: try OE_CONFIGS.FEATURE_FLAG_ENABLE_REDIS_REMOVAL

src/model/trust.model.ts
mwejuli-ch: In that case, will give you a shout when you have a moment and show you the issue. Perhaps you might advise on a more suitable approach to resolve it. I could've missed something.

src/service/payment.service.ts
```diff
@@ -3,7 +3,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { Session } from "@companieshouse/node-session-handler";
 import ApiClient from "@companieshouse/api-sdk-node/dist/client";
 import { isActiveFeature } from "../utils/feature.flag";
-import { ApplicationData } from "model";
+import { ApplicationData } from "../model";
```
mwestacott, line 6: Does this need to change? Was something broken, with the introduction in this PR of the Redis refactorings?

views/includes/check-your-answers/beneficial-owner-gov.html
```diff
@@ -80,7 +80,11 @@
 {% endset %}
 
 {% if pageParams.isRegistration %}
-  {% set changeLinkUrl = OE_CONFIGS.BENEFICIAL_OWNER_GOV_URL + "/" + bog.id %}
+  {% if not IS_REDIS_REMOVAL_ENABLED %}
```
mwestacott, line 83: I suggest simplifying the logic - remove the `not` and switch around what the blocks do.

src/utils/change.link.ts
```diff
@@ -17,3 +19,15 @@ export const createSummaryListLink = (
...
+
+/**
+ * Return a change link with the transactionId and submissionId included
+ *   - In case of an exception, return an HTML-safe hash(#)
+ */
+export const createChangeLinkWithIds = (link: string, transactionId: string, submissionId: string) => {
+  try {
+    return getUrlWithTransactionIdAndSubmissionId(link, transactionId, submissionId);
+  } catch (e) {
+    return "#";
```
mwestacott, line 31: I'm just interested in what kind of scenario an exception would be thrown (and caught here)? Also, what kind of URL would ultimately be used by the app/template? # followed by something else?

views/includes/check-your-answers/beneficial-owner-gov.html
```diff
@@ -80,7 +80,11 @@
...
+    {% set changeLinkUrl = OE_CONFIGS.BENEFICIAL_OWNER_GOV_URL + "/" + bog.id %}
+  {% else %}
+    {% set changeLinkUrl = CREATE_CHANGE_LINK_WITH_IDs(OE_CONFIGS.BENEFICIAL_OWNER_GOV_WITH_PARAMS_URL, OE_TRANSACTION_ID, OE_SUBMISSION_ID) + "/" + bog.id %}
```
mwestacott, line 86: So looks like URL here could be `#/<<a-bog-id>>`. Is that actually valid as a change link?

views/includes/check-your-answers/beneficial-owner-individual.html
```diff
@@ -99,7 +99,11 @@
 {% endset %}
 
 {% if pageParams.isRegistration %}
-  {% set changeLinkUrl = OE_CONFIGS.BENEFICIAL_OWNER_INDIVIDUAL_URL + "/" + boi.id %}
+  {% if not IS_REDIS_REMOVAL_ENABLED %}
```
mwestacott, line 102: Again, `not` should be removed and logic reversed.

src/app.ts
```diff
@@ -47,7 +53,9 @@ nunjucksEnv.addGlobal("OE_CONFIGS", config);
 nunjucksEnv.addGlobal("ERROR_MESSAGES", ErrorMessages);
 nunjucksEnv.addGlobal("COUNTRY_FILTER", countryFilter );
 nunjucksEnv.addGlobal("CREATE_CHANGE_LINK", createChangeLinkConfig);
+nunjucksEnv.addGlobal("CREATE_CHANGE_LINK_WITH_IDs", createChangeLinkWithIds);
```
mwestacott, line 56: All constant letters should be capitalised.

views/includes/page-templates/trust-historical-bo.html
```diff
@@ -51,15 +52,15 @@ <h1 class="govuk-heading-xl govuk-!-margin-0">{{ pageParams.title }}</h1>
           {
             value: pageData.trusteeType.LEGAL_ENTITY,
             text: "Legal Entity",
-            checked: formData.type == pageData.trusteeType.LEGAL_ENTITY,
+            checked: (formData.type | lower) === (pageData.trusteeType.LEGAL_ENTITY | lower),
```
mwestacott, line 55: Wondering why this change is needed for Redis?

src/app.ts
```diff
@@ -61,6 +69,13 @@ app.use(express.json());
 app.use(express.urlencoded({ extended: false }));
 app.use(cookieParser());
 
+app.use((req, res, next) => {
```
mwestacott, line 72: Should this new code be wrapped in a feature flag?

views/includes/check-your-answers/beneficial-owner-gov.html
mwejuli-ch: The `not` ensures that the larger chunk of the `if` statement is at the bottom (in the `else` clause) and not the top, which makes for cleaner, better formatted code (and avoids having top-heavy conditionals).

src/utils/change.link.ts
mwejuli-ch: The thinking behind this is that, in the very unlikely event that a url substitution fails, then we do not have to redirect the user to an error page and break the journey; the user can still view their `check-your-answers` page and see what they entered.

The `#` is an HTLM-safe option that does not break the mark-up. Obviously something would be wrong at this point that needs fixing (and would be reported by the user or tester), but the user would still be able to see a summary of their changes. Happy to discuss.

views/includes/check-your-answers/beneficial-owner-gov.html
mwejuli-ch: No, it's not a valid link. This would only avoid the journey from breaking so the user can still see a summary of the changes they made without terminating the journey by loading the error screen.

This is a bit of defensive coding as this scenario is very unlikely to happen. But if it does, we can still show the user a summary of their changes.

views/includes/check-your-answers/beneficial-owner-individual.html
mwejuli-ch: To paraphrase a previous comment: The `not` ensures that the larger chunk of the `if` statement is at the bottom (in the `else` clause) and not the top, which makes for cleaner, better formatted code (and avoids having top-heavy conditionals).

mwejuli-ch: To paraphrase a previous comment: The `not` ensures that the larger chunk of the `if` statement is at the bottom (in the `else` clause) and not the top, which makes for cleaner, better formatted code (and avoids having top-heavy conditionals).

src/app.ts
mwejuli-ch: Ok, will changes this.

Was trying to make it obvious to the user that this should read as "IDs" and not "I.D.S." 😄

views/includes/page-templates/trust-historical-bo.html
mwejuli-ch: This was needed because the mixed-case type value was not being deciphered as being equal to the all-caps type value. But with the mapping done as discussed in the huddle, we might not need this. Thanks for pointing it out though.

src/app.ts
mwejuli-ch: No, it does not need to be because that bit of code is feature-flag agnostic. Having the feature-flag present doesn't break any flow, but it's not required.

Happy to discuss this (and any of the issues raised above) if you believe I might have overlooked something important.

Cheers.

src/app.ts
mwejuli-ch: Have tested `OE_CONFIGS.FEATURE_FLAG_ENABLE_REDIS_REMOVAL` locally and it works. Will refactor and use that.

src/service/payment.service.ts
mwestacott: 👍 I'd just add a comment to testers on the JIRA ticket to ensure Payment journey is regression tested.

views/includes/page-templates/trust-historical-bo.html
mwestacott: 👍 yes, hopefully not required when we use that common mapping code.

src/app.ts
mwestacott: We've done something similar (I think) in the `journey.detection.middleware` but there we use `res.locals` to get the value available in all the Nunjucks templates.

Is `addGlobal` better to use here?

src/utils/change.link.ts
mwestacott: We call the `getUrlWithTransactionIdAndSubmissionId()` function in quite a few places and it is invoked pretty much on every page when Redis flag is switched off. We haven't taken a defensive coding approach elsewhere so not convinced it's needed here either.

views/includes/check-your-answers/beneficial-owner-gov.html
mwestacott: Yep, I've responded to the previous comment related to this.

views/includes/check-your-answers/beneficial-owner-gov.html
mwestacott: When you say 'larger' you mean that in the `else` block the line is a bit longer in length?

Not sure that justifies the addition of `not`, but up to you.

src/app.ts
mwejuli-ch: Yes, I believe `addGlobal` is a better approach here so as to keep the data export consistent with the chunk of export code a few lines prior.

src/service/payment.service.ts
mwejuli-ch: Yes, absolutely. Will mention this on the JIRA ticket so that the payment journey is also regression tested when the flag is OFF. We have a ticket (ROE-2732) about a payment issue when the flag is ON that I'll be working on immediately after this PR is approved.

src/utils/change.link.ts
mwejuli-ch: Ok, no problem. I'll remove it. As long as you're happy that in case of a url substitution error/exception, it's better to terminate the journey and show the user an error screen, rather than have a view of their data so far entered.

views/includes/check-your-answers/beneficial-owner-gov.html
mwejuli-ch: Thanks. We'll use the related comment as a proxy for this one.

views/includes/page-templates/trust-historical-bo.html
mwejuli-ch: Agreed. We'll not need this after the mapping is applied, so will refactor it back to what it was originally.

src/service/payment.service.ts
mwestacott: Thanks @mwejuli-ch .

src/utils/change.link.ts
mwestacott: I'm happy. It's an unexpected scenario - something serious has gone wrong which might even mean that the data displayed on the 'check your answers' page isn't even correct. And like I said, the `getUrlWithTransactionIdAndSubmissionId()` function would have already been called in order to get them to this screen.

mwejuli-ch: Cheers Martin. Will go ahead and remove it.

src/app.ts
mwejuli-ch: Now working on the unit tests and it looks like `OE_CONFIGS.FEATURE_FLAG_ENABLE_REDIS_REMOVAL` causes a lot of failures for the `check-your-answers` test suites. Reverting to my earlier changes fixes the issue.

So, we have:

- `FEATURE_FLAG_ENABLE_REDIS_REMOVAL` fails for journey but works for tests
- `OE_CONFIGS.FEATURE_FLAG_ENABLE_REDIS_REMOVAL` works for the journey but fails for the tests
- `IS_REDIS_REMOVAL_ENABLED` works for both the journey and the tests

Will give you a shout so I can show you in a screen-share session when you have a moment. Otherwise, if you don't mind, will revert to what I'd built before, i.e. use `IS_REDIS_REMOVAL_ENABLED` as this works for both.

Cheers.

markpit: We had that issue previously and solved it with this approach: https://github.com/companieshouse/overseas-entities-web/blob/main/src/controllers/beneficial.owner.type.controller.ts#L30

mwejuli-ch: ok, got it

my only issue is that passing it in the controller just means if we have a similar issue on another page, then we'll have to pass in that value in that other controller again. this leads to to fragmentation everywhere. doing it the way i've done it in `app.ts` means we only do it once and we reuse that value in any view where we need it. we avoid the controller-to-controller repetition.

src/controllers/shared/common.resume.submission.controller.ts
```diff
@@ -27,17 +22,22 @@ import { BeneficialOwnerOther, BeneficialOwnerOtherKey } from "../../model/benef
...
+} from "../../model/data.types.model";
 
 export const getResumePage = async (req: Request, res: Response, next: NextFunction, resumePage: string) => {
   try {
     logger.debugRequest(req, `GET a saved OE submission`);
 
     const { transactionId, overseaEntityId } = req.params;
     const infoMsg = `Transaction ID: ${transactionId}, OverseasEntity ID: ${overseaEntityId}`;
-    const isRegistration: boolean = req.path.startsWith(config.LANDING_URL);
+    const isRegistration: boolean = isRegistrationJourney(req);
```
mwestacott, line 40: Again, changes look fine, but given that the resume controller is also now being updated here for the 'check your answers' story I would add a note to testers that the resume journey will require a regression test (registration, update and remove journeys).

mwejuli-ch: Every story that is tested is also regression tested with the flag on/off till a submission to chips is successful. This is done with a new submission, a resumed submission, and the update/remove journeys. So this bit will be tested. Thanks.

mwestacott: Excellent. Thanks.

src/config/index.ts
```diff
@@ -497,3 +497,25 @@ export const TRUST_ENTRY_WITH_PARAMS_URL = REGISTER_AN_OVERSEAS_ENTITY_WITH_PARA
 export const TRUST_INFO_WITH_PARAMS_URL = REGISTER_AN_OVERSEAS_ENTITY_WITH_PARAMS_URL + TRUST_INFO_PAGE;
 export const TRANSACTION_ID_URL_KEY = "/transaction/";
 export const SUBMISSION_ID_URL_KEY = "/submission/";
+
+// Check-your-answers page links
+export const ENTITY_CHANGE_NAME_WITH_PARAMS = OVERSEAS_NAME_WITH_PARAMS_URL + ENTITY_NAME;
+export const PRESENTER_CHANGE_FULL_NAME_WITH_PARAMS = PRESENTER_WITH_PARAMS_URL + "#full_name";
+export const PRESENTER_CHANGE_EMAIL_WITH_PARAMS = PRESENTER_WITH_PARAMS_URL + "#email";
```
mwestacott, line 504: You should use the `EMAIL` constant already defined here.

Ideally, we'd have a new `FULL_NAME` constant too for "#full_name".

src/utils/change.link.ts
```diff
@@ -17,3 +19,10 @@ export const createSummaryListLink = (
 export const createChangeLinkConfig = (href: string, text: string, dataEventId: string) => {
   return createSummaryListLink('Change', href, text, dataEventId);
 };
+
+/**
+ * Return a change link with the transactionId and submissionId included
+ */
+export const createChangeLinkWithIds = (link: string, transactionId: string, submissionId: string) => {
+  return getUrlWithTransactionIdAndSubmissionId(link, transactionId, submissionId);
```
mwestacott, line 27: Now you've removed that error handling, this function only delegates to another. I would just get rid of it and call `getUrlWithTransactionIdAndSubmissionId()` directly.

mwejuli-ch: Here, we're exporting a method from `src` to `views` in a manner consistent with how other methods have been exported (see other exports in `app.ts`). To keep that consistency intact, I believe it's best we leave it like this. Happy to discuss.

src/config/index.ts
mwejuli-ch: Good spot Martin. Have done so and pushed. Cheers.

src/utils/change.link.ts
mwestacott: Ah yes, I see. Makes sense.

mwestacott (commented):
LGTM

markpit (approved):

mwestacott (approved):

markpit (approved):

PR #1809, opened 2026-05-11
feat/cc-3017: get and save app data using DB for the resume update journey

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/CC-3017

### Change description

- Gets and saves data to and from the API for the resume update journey
- Updates and adds unit test as required
- Adds some formatting changes

### Work checklist

### DIFF
--- a/src/controllers/shared/common.resume.submission.controller.ts
+++ b/src/controllers/shared/common.resume.submission.controller.ts
@@ -10,17 +10,17 @@ import { isActiveFeature } from "../../utils/feature.flag";
 import { startPaymentsSession } from "../../service/payment.service";
 import { getTransaction } from "../../service/transaction.service";
 import { mapTrustApiReturnModelToWebModel } from "../../utils/trusts";
-import { isRegistrationJourney } from "../../utils/url";
+import { isRegistrationJourney, isRemoveJourney } from "../../utils/url";
 import { DueDiligenceKey } from "../../model/due.diligence.model";
 import { OverseasEntityDueDiligenceKey } from "../../model/overseas.entity.due.diligence.model";
 
 import { getOverseasEntity, updateOverseasEntity } from "../../service/overseas.entities.service";
 import { WhoIsRegisteringKey, WhoIsRegisteringType } from "../../model/who.is.making.filing.model";
 import { BeneficialOwnerGov, BeneficialOwnerGovKey } from "../../model/beneficial.owner.gov.model";
-import { BeneficialOwnerIndividual, BeneficialOwnerIndividualKey } from "../../model/beneficial.owner.individual.model";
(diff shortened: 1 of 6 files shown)


src/controllers/shared/common.resume.submission.controller.ts
```diff
@@ -94,7 +95,7 @@ const setWebApplicationData = async (
   appData: ApplicationData,
   transactionId: string,
   overseasEntityId: string,
-  isRegistration: boolean
+  isRegistrationOrUpdate: boolean
```
knite51: Given that when the function is invoked the `isRemove` logic is checked, will it not be appropriate to rename the parameter as such?
 ` await setWebApplicationData(req, session, appData, transactionId, overseasEntityId, !isRemove);`
Either that or the isRemove might not be the best value to pass. Also having to negate the value and the logic twice is quite confusing - perhaps something in this line

```
await setWebApplicationData(req, session, appData, transactionId, overseasEntityId, isRemove)
const setWebApplicationData = async (
  req: Request,
  session: Session,
  appData: ApplicationData,
  transactionId: string,
  overseasEntityId: string,
  isRemove: boolean
)
 if (isRemove || !isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL)) {}
 
  if (!isRemove && isActiveFeature(config.FEATURE_FLAG_ENABLE_REDIS_REMOVAL)) {


```

Not sure if that logic is what you aiming at though especially if this was a remove journey

mwejuli-ch: after some thought, i believe the logic is written correctly. 

additionally, the parameter name reflects the fact that it's required only for the registration or update journey, hence `isRegistrationOrUpdate`

also, it will be updated whilst we're working on the remove journey so we will have another opportunity to look at it then. but as it stands, i see no problem with the logic, and it also happens to be working as expected from a journey perspective.

hope this helps

knite51: I am not convinced, it is reading double negative and very confusing. If it is required only at registration and update why not pass in those parameters? Might as well make it cleaner once and for all, but if you insist.

mwejuli-ch: the if statement reads:

_"if the redis flag is not set, or the journey is not registration or update..."_

am not sure that the statement above is a confusing statement at all

would have been easier to refactor the double negative if we had an `else` statement.. however, without an `else` statement, that's the logic required to execute that block of code.

mwejuli-ch: refactored with `isRemove` flag.. please have another look

knite51 (approved):

PR #1815, opened 2026-06-01
feat/cc-3015, feat/cc-3043: get/save application data from DB for Update and Remove journeys

mwejuli-ch:
### Important note:

This PR is an amalgam of several constituent PRs that have been merged into this branch (as a central `Redis removal` branch) and now, given that the journeys are dev-complete, and thereby ready for testing, we're creating a PR to merge all of those constituent PRs into the `main` branch.

### JIRA links

- Epic 1: https://companieshouse.atlassian.net/browse/CC-3015
- Epic 2: https://companieshouse.atlassian.net/browse/CC-3043

### Change description

- Gets and saves data to and from the API for the end-to-end Update journey
- Gets and saves data to and from the API for the end-to-end Remove journey
- Updates and adds unit test as required
- Adds formatting changes

### Work checklist


src/controllers/update/update.review.managing.officer.corporate.controller.ts
```diff
@@ -1,85 +1,141 @@
...
+      return res.redirect(boRedirectUrl);
     }
-  } catch (error){
+  } catch (error) {
     logger.errorRequest(req, error);
     next(error);
   }
 };
+
+export const isMoReviewable = (appData: ApplicationData, moIndex: any, requestId: string | undefined): boolean => {
```
markpit, line 131: I was wondering about the flag being used here, as far as I know, the service currently allows users to review managing officers and beneficial owners and is not affected by the flag. Can you confirm if this new logic means that if the FEATURE_FLAG_ENABLE_REDIS_REMOVAL is switched off that the review MO and BO part of the service will still be available for the user? thanks.

src/service/transaction.service.ts
```diff
@@ -40,6 +40,33 @@ export const postTransaction = async (req: Request, session: Session, data?: App
   return response.resource.id;
 };
 
+export const updateTransaction = async (req: Request, session: Session, data?: ApplicationData) => {
```
markpit: updateTransaction doesn't look like it is used anywhere, only by a test, can this be removed?

src/controllers/update/update.review.managing.officer.corporate.controller.ts
mwejuli-ch: yes, this logic will still be available to users when the flag is OFF coz the underlying review method `checkAndReviewManagingOfficers` is called in the  `update/beneficial.owner.type.controller.ts` controller.

however, when the flag is ON, the initial call in `update/beneficial.owner.type.controller.ts` does not work as expected at that stage of the request/response cycle hence the need to push that logic forward to the pertinent controller.

so, the call to `isMoReviewable` is only relevant to a flag ON scenario, and not the flag OFF (and will not affect the review logic with the flag OFF).

you can confirm this in a test journey with the flag set to either value 

hope this helps,

thanks.

src/service/transaction.service.ts
mwejuli-ch: Good spot Mark. Cheers.

Will remove this method - and the corresponding test - in the next push.

eedwards0 (approved):
The individual changes in this PR have been heavily scrutinised, raised as PRs in their own right, and tested in Docker before being merged into this feature branch.

PR #1828, opened 2026-07-21
feat/cc-3443:  add query param to links of reviewed BOs and MOs

mwejuli-ch:
### JIRA link

https://companieshouse.atlassian.net/browse/CC-3443

### Change description

 - Adds query param to links of reviewed BOs and MOs to indicate that they have been reviewed

### Work checklist


src/controllers/update/update.review.beneficial.owner.gov.ts
```diff
@@ -47,6 +47,7 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
     logger.debugRequest(req, `${req.method} ${req.route.path}`);
     const appData = await getApplicationData(req);
     const index = req.query.index;
+    const reviewed = req.query.r;
```
eedwards0, line 50: Could we use a more descriptive name, such as reviewed, instead of a single letter?

eedwards0 (approved):
Looks good. 
Optional suggestion: consider renaming r to reviewed for readability, but not a blocker.

src/controllers/update/update.review.beneficial.owner.gov.ts
mwejuli-ch: the idea is to obfuscate it from the end-user as it's in the query string of the url, in the address bar (and also to keep the query signature short)..

happy to have it renamed if the consensus is to have it explicitly named in the url.

also, when followed through in  the code, it's documented there as "reviewed".

..but happy either way

